PR #87, opened 2015-05-03
Fixed turkish translation preventing a successfull build

pvagner:
I think turkish translation might be still a work in progress however in the mean time I've tweaked it a bit so the build won't fail.


bear101:
Hi, I got another Turkish translation so I'm submitting that instead.

PR #184, opened 2016-04-14
Turkish updated for Client Name

OzancanKaratas:
You had added a new string, I've translated it.


bear101:
Thanks for the rapid update :)

PR #186, opened 2016-04-21
Turkish updated

OzancanKaratas:
I have updated Turkish translations for iTeamTalk and TeamTalk for Android. Please send iTeamTalk into App Store for iPhone 4.


bear101:
Thanks!

PR #192, opened 2016-04-30
iTeamtalk: Updated spanish translation

manuelcortez:


bear101:
Thanks a lot! I've already submit the 5.1.3 release to Apple, so the updated translation will not be included until the next release.

PR #195, opened 2016-05-03
Turkish updated

OzancanKaratas:
I have translated to added new strings in Android and fixed some translations in iOS.


bear101:
Teşekkürler

OzancanKaratas:
Yaa Bjorn, is "Toggle to enable transmission" doing work into iOS client? I don't find a suitable translation to it.

bear101:
It's an accessible hint for blind users on the transmit button. Toggle means to switch state, e.g. on/off.

bear101:
You will only hear it with VoiceOver enabled.

bear101:
Hm. Maybe it should be "toggle to enable/disable transmission".

OzancanKaratas:
Aaaa yes, please change it. Also I have translated the string "Double tap and hold to transmit. Tripple tap to toggle.", but this text has been read as English by VoiceOver. This string has been configured as wrong. Please fully review Localizable.strings file because translations is verry important and shouldn't be wrong.

bear101:
Yes, I've found the bug. I need to update the translation files with the 
latest strings.

I'll try and see if I can make a script to convert the UCS-2 format to 
UTF-8 so it's possible to see changes on GitHub. The UCS-2 format used 
by Apple really stinks.

--
Kind regards

Bjoern D. Rasmussen
www.bearware.dk

bear101:
I've updated the Turkish language files and also converted the files to UTF-8. Now I just need to figure out how to make Git treat the files as text and not as binaries.

PR #199, opened 2016-05-06
Translation fixes for Turkish

OzancanKaratas:
I've updated language files but you should review this files anyway. :)


bear101:
Thanks but please use UTF-8 when you save the file. That way it's possible to see what has changed.

OzancanKaratas:
Aaaa apologise. But you had changed just Localizable.strings to other commit. Doesn't the file Main.strings use UTF-8?

bear101:
I have updated Localizable.strings to UTF-8 after your pull-request was commit. 
Main.strings has always been UTF-8.

PR #203, opened 2016-05-27
Updated Turkish translation

OzancanKaratas:
I've updated Android translations, translated new iOS strings and saved as UTF-8 with non-BOM to Localizable.strings. Check anyway :)


bear101:
Great thanks!

PR #204, opened 2016-05-27
Updated german strings

schulle4u:
Sorry for being late with updates, sometimes it is a little difficult to keep track of changes among all other commits. :)


bear101:
Danke!

PR #214, opened 2016-07-07
i18n: Updated new strings for Turkish

OzancanKaratas:
Hi Bjorn,
New Android strings wer translated. Also some iOS strings wer fixed.
Add this to new releases.
Thanks. :)


bear101:
Thanks!

PR #226, opened 2016-10-19
Updated Turkish

OzancanKaratas:


bear101:
thanks @OzancanKaratas! Btw, I've just come home from a one week holiday in Turkey :)

OzancanKaratas:
Aaaaa, have you gone which city?

bear101:
I went to Northen Cypres. It was nice to get away from the horrible Danish weather :)

bear101:
But I've been to Turkey several times.. Istanbul, Izmir and Antalya.

OzancanKaratas:
Hmmm. Kayseri is waiting for you :)
In the meantime, 49th string should be changed in the iTeamTalk. Because impossible to understand it. Also Version Information title and all its content isn't found into localization files. I think you need to run the script :)

PR #229, opened 2016-11-07
Updated Turkish and more

OzancanKaratas:


bear101:
Can you create a pull-request where the English.ts file isn't deleted. English.ts is required by translators (they may not be able to generate it themselves).

OzancanKaratas:
I created a revert commit, you can merge it now. If you will add new strings, don't merge it which I can translate new strings on this pull request. For example, run iTeamTalk localization script. App Information header was not translated yet. We don't open a new pull request for 1 string.

PR #242, opened 2016-12-21
Some changes

OzancanKaratas:
iTeamTalk uses en.lproj folder, base.lproj folder is unnecessary. But I didn't deleted folder, just deleted Localizable.strings file.


bear101:
Localizable.strings(Base) is required by Xcode. If you know how to make Xcode not depend on this file then you're welcome to fix it.

PR #255, opened 2017-01-14
Update german Qt translation

### LINKED ISSUE #251: [Qt client] Keyboard shortcuts broken in some locales
In some locales such as german, the modifier keys (CTRL, Shift etc) have different names. This causes broken keyboard shortcuts in the Qt client. In the MacOS version keyboard shortcuts probably are generally broken when used with locales, because of totally different modifier names (CMD, option...). 

Therefore we should think about removing shortcuts completely from translation files to avoid this problem or just leave them as untranslated. Or is there another possibility to fix this? I'll provide a temporary fix for the german locale this weekend. 

Steffen

schulle4u:
* Removed translation of keyboard shortcuts to make them work again (fixes #251)
* other small and cosmetic fixes

Can someone test this on MacOS please? Windows and linux clients are working.

### DIFF
--- a/qtTeamTalk/languages/German.ts
+++ b/qtTeamTalk/languages/German.ts
@@ -1277,7 +1277,7 @@
     </message>
     <message>
         <source>Modifying subscriptions is not allowed</source>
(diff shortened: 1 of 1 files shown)


bear101:
Danke Steffen!

PR #294, opened 2017-04-05
Completed Qt TeamTalk translation

OzancanKaratas:


bear101:
Phew, it's a big one :)

OzancanKaratas:
Yes, because I made it from the beginning :)<br>I want to ask a question:<br>Do you try to compile iTeamTalk deleting settings.bundle folder?<br>Because some apps is not using it. So I think it is unnecessary. Also a bad appearance in the iOS's Settings.

PR #302, opened 2017-05-06
New updates

OzancanKaratas:
Hi @bear101
I've updated readme file but if you are not want to supported by an unofficial platform, please revert this commit after merging.


OzancanKaratas:
@bear101 Please edit the following lines on TeamTalk Android:
"Voice Activation" = "Voice activation"
"No interruptions" = "No Interruptions"
These lines is wrong. Do not merge this PR, because a new strings.xml file will be sent after editing base file.

PR #306, opened 2017-06-10
Updated Russian

alex-crystal:


bear101:
@alex-crystal Can you see if the new runtime permissions on Android 6 are working? Unfortunately I don't have an Android 6 or 7 device and the emulators are too slow to be usable.

PR #371, opened 2018-09-25
Turkish update for iOS and Android

OzancanKaratas:


bear101:
Thanks @OzancanKaratas

OzancanKaratas:
@bear101 No problem. However, you should update content of Main.strings. The file Main.storyboard have a few translatable strings.

PR #372, opened 2018-09-26
Update iTeamTalk v5.3.3 Simplified Chinese translation

dingpengyu:
This is the latest iTeamTalk v5.3.3 Simplified Chinese translation, which has been checked to request to sync to the official GitHub. Thank


bear101:
Thanks @dingpengyu

PR #382, opened 2018-11-03
Update chinese translation for iTeamTalk and TeamTalk Classic.

hejiadong0608:


bear101:
Thanks for the update!

PR #383, opened 2018-11-04
Add chinese simplified translator for android, update chinese simplified for iOS and classic.

hejiadong0608:
Add chinese simplified translator for android, update chinese simplified for iOS and classic.


bear101:
Thanks a lot! I'll include it in the next release.

PR #397, opened 2019-01-05
Fix Chinese translation for android and iOS.

hejiadong0608:


bear101:
Thanks @hejiadong0608 !

bear101:
I'll see if I can find a way to TTS the received text messages. There should probably be a checklist of which announcements the user wants to enable.

Issue #407, opened 2019-03-11
The following string is missing from the TeamTalkClassic client


dingpengyu:
Hi bjoerndrasmussen, there are some strings in the TeamTalkClassic client that are not available, the following is a list:
User Account：
User can change nickname checkbox selected (cannot be translated)
User can record voice in all channels check box unchecked (cannot be translated)
Close button Alt+l (cannot be translated)
Channel information:
OK button (cannot be translated)
Cancel button (cannot be translated)
Server statistics:
OK button (cannot be translated)
TeamTalk5Classic connection:
An error occurred while perform a requested command: Invalid user account
(cannot be translated)
Failed to connect to host "" TCP port UDP port . Check that the server is running on the specified address and that a firewall isn't preventing clients from connecting.
(cannot be translated)
thank

dingpengyu:
cc
@bear101

bear101:
I fixed a few of the missing in this commit: 2d986d2b17d7caa8653692748df4d5d5959f02bf

dingpengyu:
Hi Thanks for the work of bear, the following is my statistical feed string:
Server property:
Create anonymous user account? (no translation)
User Account:
button(Untitled)
button (Untitled)
Kick out + kick out and ban:
An error occurred while perform a requested command:
Command not authorized
thank

dingpengyu:
Hi @bear101 
I tested TeamTalk 5 Classic 5.5.0.4978
The following translation string is not in lng_classic/English.lng:
No Audio Preprocessor
Start position 滑块 0
Play 按钮 Alt+p
TeamTalk Audio Preprocessor
Setup 按钮 Alt+s
Start position 滑块 0
Play 按钮 Alt+p
Speex DSP Audio Preprocessor
Setup 按钮 Alt+s
Start position 滑块 0
Play 按钮 Alt+p
Audio preprocessor
thanks

bear101:
I think has been fixed by @Oreonan

PR #409, opened 2019-03-18
Update Chinese translation.

hejiadong0608:


bear101:
Thanks for the translation. I still need several of the strings mentioned in #407. I hope to get around to it eventually.

hejiadong0608:
Hello.
I don't know the ID of these strings. How can I get these IDs?
Thank you!
@bear101

bear101:
If you open the new English.lng file you'll get the new IDs. Otherwise look in this commit: 2d986d2b17d7caa8653692748df4d5d5959f02bf

hejiadong0608:
These new IDs are already included in this pull request. But there are many more strings memtioned in #407 
that are not included in new English.lng.
@bear101

PR #410, opened 2019-03-18
Update Polish translation.

michalciolek:


bear101:
Thanks for the updated translation!

PR #416, opened 2019-05-27
Update Chinese translation for classic and android.

hejiadong0608:


bear101:
Thanks a lot! I've just pushed out a new Android beta client. I'll include your translation before the final release.

poretsky:
And, please, don't miss Russian translations update I've just pushed to
master.

PR #424, opened 2019-06-05
Update Chinese translation for android

hejiadong0608:
Thank you very much for #396 .


dingpengyu:
cc
@bear101

bear101:
Many thanks! I'll make sure it makes it to the TeamTalk v5.4 release

PR #425, opened 2019-06-05
Croatian and czech languages update

zstanecic:
Here are new translated czech and croatian files.
Czech files provided by Roberto Lombino.


bear101:
Thanks for the updates!

PR #428, opened 2019-06-09
Update Chinese translation for iOS client.

hejiadong0608:


bear101:
Thanks! That was quick :)

PR #429, opened 2019-06-09
czech and croatian translation updates

zstanecic:
Updated croatian and czecz translations.
For czech again thanks to Roberto Lombino.


bear101:
Thanks! Quick as well!

PR #433, opened 2019-06-26
little fix updated croatian, small fixes. 

zstanecic:
serbian translation added


bear101:
Thanks for the new translation. I'll include it in the next release.

PR #480, opened 2020-02-27
Android client: update Chinese translation

hejiadong0608:


bear101:
Thanks. Very quick :)

PR #481, opened 2020-03-09
Added new Brazilian Portuguese translations in TeamTalk Classic

DalisonCPU:


bear101:
Thanks for the update!

PR #482, opened 2020-03-13
Turkish translations

OzancanKaratas:


bear101:
Thanks for the corrections @OzancanKaratas

PR #488, opened 2020-04-13
Update french translation of classic client and add french translation to Android client

Oreonan:


Oreonan:
I will improve that and reopen à New pull request later

PR #501, opened 2020-04-21
Update Chinese translation for classic client

hejiadong0608:


bear101:
Thanks! Many merge commits :)

PR #516, opened 2020-05-19
Updated spanish translation for Classic client

jmdaweb:
Updated spanish translation for classic client using the provided translation tool on bearware.dk website.


bear101:
Wow! A lot of new items. Thanks!

PR #520, opened 2020-06-01
Update TeamTalkClassic simplified Chinese translation

dingpengyu:
No description


dingpengyu:
hi
@bear101 
add <item id="1470">Sort channels</item> String
in
 TeamTalkClassic_v5.6.0.4995_Portable
Cannot display properly
test environment
Microsoft Windows [version 10.0.19041.264]
thanks

PR #522, opened 2020-06-09
First attempt to add more translatable elements to the classic client

Oreonan:


Client/TeamTalkClassic/TeamTalkDlg.cpp
```diff
@@ -429,7 +429,7 @@ void CTeamTalkDlg::StartMediaStream()
 
     if (!TT_StartStreamingMediaFileToChannelEx(ttInst, szFileName, &mfp, &vidcodec))
     {
-        MessageBox(_T("Failed to stream media file."),
+        MessageBox(LoadText(IDS_STREAMFAILEDBOX, _T("Failed to stream media file."),
```
bear101, line 432: Missing ending parentesis. Should be:
MessageBox(LoadText(IDS_STREAMFAILEDBOX, _T("Failed to stream media file.")),

bear101: But other than that it looks good! I can go through your code when you think it's ready and compile a version for you.

PR #527, opened 2020-06-13
Continue translation of classic"

Oreonan:
Add:
*ClientPage (GUI)
*ChatRichEditControl (GUI)
*BanTypeDlg (GUI)
*ChannelDlg (GUI)


bear101:
Please download latest v5.6.0.4998 again to try translation: http://bearware.dk/beta/TeamTalkClassic_v5.6.0.4998_Portable.zip

bear101:
And good job! As you can see it is not a trivial task :) But it has already been a big improvement.

Let me know also if some controls need to be resized.

PR #530, opened 2020-06-15
Dummy channels as pure info items

poretsky:
It appears to be more or less common practice to publish
announcements, schedules, and other info via channel items. Such
channels usually have maximum users limit set to 0. In this sense they
are dummy, i.e. not joinable. Thus there is no reson to keep
corresponding button. Since text fields of the item bear useful
information, it seems rational to make them multiline in order to
avoid cropping. And, at last, these info items should be displayed at
the top of list. All this is implemented here.

Maybe, there is another way, more natural and elegant?

Of course, I know that server can pass general info to clients via
messages. But it seems to be less convenient, since requires shift to
another tab where general info becomes lost in other messages.

### DIFF
--- a/Client/TeamTalkAndroid/src/main/java/dk/bearware/gui/MainActivity.java
+++ b/Client/TeamTalkAndroid/src/main/java/dk/bearware/gui/MainActivity.java
@@ -278,7 +278,7 @@ implements TeamTalkConnectionListener,
     @Override
     public boolean onPrepareOptionsMenu(Menu menu) {
         boolean isEditable = curchannel != null;
-        boolean isJoinable = (ttclient != null) && (curchannel != null) && (ttclient.getMyChannelID() != curchannel.nChannelID);
+        boolean isJoinable = (ttclient != null) && (curchannel != null) && (ttclient.getMyChannelID() != curchannel.nChannelID) && (curchannel.nMaxUsers > 0);
         boolean isMyChannel = (ttclient != null) && (curchannel != null) && (ttclient.getMyChannelID() == curchannel.nChannelID);
         menu.findItem(R.id.action_edit).setEnabled(isEditable).setVisible(isEditable);
         menu.findItem(R.id.action_join).setEnabled(isJoinable).setVisible(isJoinable);
@@ -866,8 +866,9 @@ implements TeamTalkConnectionListener,
         private static final int PARENT_CHANNEL_VIEW_TYPE = 0,
             CHANNEL_VIEW_TYPE = 1,
             USER_VIEW_TYPE = 2,
+            INFO_VIEW_TYPE = 3,
 
-            VIEW_TYPE_COUNT = 3;
+            VIEW_TYPE_COUNT = 4;
 
         private LayoutInflater inflater;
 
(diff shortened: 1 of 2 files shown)


poretsky:
I notice that this pull request has not passed automatic check. But,
of course, I've tested it before submission and everything went
well. What happens? Should something be corrected?

bear101:
There's unit-test which is constantly failing. I've been trying to
figure out why it's failing but so far without luck.

--
Kind regards

Bjørn Damstedt Rasmussen
www.bearware.dk

bear101:
Sorry, I haven't had time to look at it yet. I'm currently investigating some latency issues on Android.

bear101:
@poretsky On a different matter. Should we change sorting so users are before channels? This is default on Windows and iOS.

Oreonan:
@bear101 I think do this it's a very good idea, to uniformize client's behavior.

poretsky:
I think, it would be useful as an option.

beqabeqa473:
hi all.

also, a bit related to this thread, maybe add some content description
to users with empty nicknames.
i am saying that because talkback doesn't focus on such empty cells,
and it is nearly impossible to do some administrative tasks with such
type of items.

On 6/22/20, Igor B. Poretsky <notifications@github.com> wrote:
> I think, it would be useful as an option.
>
>
> --
> https://github.com/BearWare/TeamTalk5/pull/530#issuecomment-647201388


-- 
with best regards beqa

poretsky:
Are such users permitted at all? I think, they shouldn't be.

beqabeqa473:
i also think so, but on some servers they are still using empty nicknames.

On 6/25/20, Igor B. Poretsky <notifications@github.com> wrote:
> Are such users permitted at all? I think, they shouldn't be.
>
>
> --
> You are receiving this because you commented.
> https://github.com/BearWare/TeamTalk5/pull/530#issuecomment-649329761


-- 
with best regards beqa

bear101:
I've modified the Android client so it now displays users before channels.

poretsky:
Possible solution for empty nick names is suggested in #552.

poretsky:
About sorting items in a channel. I think, it would be better
implemented as an option. anyway, in real circumstances users appear and
disappear in a channel much more frequently than subchannels. So, having
users at the top might cause some problems in subchannels
navigation. Moreover, it seems natural to have so-called sticky channels
(with no users allowed) always at the top.

bear101:
Hm, yes, sticky channels in the top would be good but it's a bit difficult to implement.

poretsky:
Yes, I thought about that, but an option for sorting order seems to be a
good solution.

poretsky:
I'd suggest to introduce a checkbox leaving the legacy behaviour as a
default, since the most users are just used to it.

@bear101 I can implement this option, but I don't know whether you
already do it yourself. So, I don't want to duplicate the work.

And after that, I think, we'll have pretty enough sensible changes to
make new Android release.

bear101:
How about sorting like this: Sticky channel, users, parent channel, subchannels?

Otherwise create the legacy behavior. I haven't started looking into this.

poretsky:
I think, it would be good variant as well. Anyway, it looks quite
natural, thus, no need to introduce an option. There is necessary to
keep sticky channel in a separate list, but seems not too complicated in
general.

poretsky:
Implementation suggested in #562.

PR #538, opened 2020-06-19
New try to add label to main window's fields

Oreonan:


Oreonan:
I've make an error while created my branch :(, I'll do a new PR just for good commit in some minutes.

PR #548, opened 2020-06-25
Android client: switching tabs by buttons

poretsky:


bear101:
good change. Much simpler!

beqabeqa473:
hello.

i tried to test this feature and on my device i observed the following.

when switching tabs, content is not updated. when i am logging on the server, i have selected channels tab, when i've tried to switch to messages tab, content pane is empty.

can you reproduce this problem?

poretsky:
Fixed in f349d5ec.

PR #552, opened 2020-06-25
Android client: Display empty nicks as anonymous

poretsky:


Oreonan:
@poretsky can you also labellize edit fields please? Specificly on chat's tab and on MP dialog? I've tried this but I'm very bad on Java :(
@bear101 please think to update translation's files if it's necessary ;)

bear101:
I've changed the blank nickname so it's "NoName - #USERID". Then they will be unique.

bear101:
@Oreonan latest Windows clients can be found here: http://bearware.dk/beta/TeamTalk_v5.6.0.5000_Setup.exe

Oreonan:
@bear101 Yes :) I know :), I wanted to talk about updating the translation for Android if new strings are added. Like "NoName" for example :)

Oreonan:
@bear101 Oh but this string it's already on translation's files, so I've say nothing :), sorry :)

bear101:
@Oreonan I've updated Qt client and Classic client to display "NoName -#USERID" for blank nicknames.

Oreonan:
@bear101 Great :), thanks

PR #553, opened 2020-06-26
Work on Android client

Oreonan:


bear101:
@Oreonan apostrophe has to be escaped by backslash in Android's XML

PR #557, opened 2020-06-27
Updating status message when manually switch to away mode

Oreonan:


Client/qtTeamTalk/changestatusdlg.cpp
```diff
@@ -59,12 +59,16 @@ void ChangeStatusDlg::slotAccepted()
     m_user.nStatusMode &= STATUSMODE_FLAGS;
     if(ui.availBtn->isChecked())
         m_user.nStatusMode |= STATUSMODE_AVAILABLE;
+    ttSettings->setValue(SETTINGS_GENERAL_STATUSMESSAGE, ui.msgEdit->text());
```
bear101, line 62: Lines 62-64 are duplicate of lines 70-72.

Client/qtTeamTalk/changestatusdlg.cpp
```diff
@@ -61,10 +61,12 @@ void ChangeStatusDlg::slotAccepted()
         m_user.nStatusMode |= STATUSMODE_AVAILABLE;
     else if(ui.awayBtn->isChecked())
         m_user.nStatusMode |= STATUSMODE_AWAY;
+                    TT_DoChangeStatus(ttInst, m_statusmode, _W(tr("Away")));
```
bear101, line 64: The status message should be what the user has specified in this case. E.g. away status set to "I'll be back in 10 minutes".

Client/TeamTalkClassic/gui/ChangeStatusDlg.cpp
```diff
@@ -78,6 +78,7 @@ void CChangeStatusDlg::OnBnClickedRadioAway()
 {
     m_nStatusMode &= ~STATUSMODE_MASK;
     m_nStatusMode |= STATUSMODE_AWAY;
+                        TT_DoChangeStatus(ttInst, STATUSMODE_AWAY, LoadText(IDS_AWAY, _T("Away")));
```
bear101, line 81: Change status is handled in TeamTalkDlg.cpp when the DoModal() function returns.

Client/qtTeamTalk/changestatusdlg.cpp
Oreonan: I've tried to implemente this in 99a1867, I really don't sûre of result :)

Oreonan:
Unfortunatly, d70b455 don't remove "Away" when you reswitching to another status mode :(

bear101:
I think it would be a better way to set Away/Question/Streaming mode on the "receiver side". This way the text is also translated into your local language.

In order words the szStatusMsg should be set here for all modes:
https://github.com/BearWare/TeamTalk5/blob/master/Client/TeamTalkClassic/gui/SessionTreeCtrl.cpp#L997

Oreonan:
Better implementation to come little more later

PR #558, opened 2020-06-27
Android client: bluetooth headset microphone usage facility

poretsky:
It can be enabled in sound settings menu now.


bear101:
Great! I'll try it out with my headset.

bear101:
I had to reconnect in order for the Bluetooth microphone to be enabled but it works! Super!

poretsky:
Fixed in 3e487645. Reconnection shouldn't be necessary now.

PR #563, opened 2020-06-29
Adding about infos to translation

Oreonan:


Client/TeamTalkAndroid/src/main/AndroidManifest.xml
```diff
@@ -30,6 +30,9 @@
         android:label="@string/app_name"
         android:theme="@style/AppTheme"
         android:usesCleartextTraffic="true">
+            <intent-filter>
+<category android:name="android.intent.category.LAUNCHER" />
```
bear101, line 34: This is apparently not the right way to do it because it cannot launch.

PR #566, opened 2020-06-29
QT client: Generate languages files when using make

Nardol:
Call lrelease <filename>.pro for teamtalk5.pro and teamtalk5pro.pro to have updated languages when compiling the QT client.
Using this method we can have last translations from the repository.

### DIFF
--- a/Client/qtTeamTalk/teamtalk5.pro
+++ b/Client/qtTeamTalk/teamtalk5.pro
@@ -32,3 +32,8 @@ CONFIG(release, debug|release) {
         LIBS += -sectcreate __TEXT __info_plist $${TARGET}.app/Contents/Info.plist
     }
 }
+
+@
+QMAKE_POST_LINK = lrelease teamtalk5.pro
+@
+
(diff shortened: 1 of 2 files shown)


bear101:
I typically run lupdate and lrelease manually because you end up with so many modified files in your git checkout.

Nardol:
As *.qm files are ignored by .gitignore in language directory, how these 
files can cause trouble?

That's why I though about doing so and not update. We have last 
translations but we don't alter the repository with file which could be 
accidentaly pushed.

bear101:
.ts files should only be generated at the request of the user. Otherwise the checkout will have modified files and "git add -u" will include unwanted files.

Nardol:
Sorry to be so insistent but lrelease generate .qm files, not ts files.
To quote lrelease command line help:
> lrelease is part of Qt's Linguist tool chain. It can be used as a stand-alone tool to convert XML-based translations files in the TS format into the 'compiled' QM format used by QTranslator objects.

So it doesn't modify TS files, only QM.

QM files are part of files ignored by Git (listed in .gitignore).
As I don't make a call to lupdate, TS file are not modified durring ```make```
I tested and it did not add any modified files on git side.

bear101:
Can't you make the .qm file generation as a separate make-target? Translation files are not part of the normal compilation process.

I think you can add something like this to the teamtalk5.pro file:
`localize.commands += lrelease teamtalk5.pro;
QMAKE_EXTRA_TARGETS += localize`

Then you run `qmake teamtalk5.pro; make localize`

Nardol:
Done in the same branch.

PR #570, opened 2020-06-30
Display specific message when streaming media file to channel

Oreonan:


Oreonan:
Better implementation to come little more later

PR #571, opened 2020-06-30
Writing change log in HTML format

Oreonan:


bear101:
I prefer Mark-Down to HTML. It's much easier to edit.

PR #572, opened 2020-07-01
Readonly history on QT

Oreonan:


bear101:
If I remember correctly read-only disables the cursor which is quite convenient.

Oreonan:
I have not seen this behavior. According to my tests the cursor is 
active. Tests performed in the private message history area.

bear101:
On macOS there's no cursor if text field is readonly. This means you cannot use keyboard for text selection.

PR #575, opened 2020-07-03
Brazilian Portuguese translations

DalisonCPU:
PT language for ITeamTalk has also been added


bear101:
Wow! Thanks for iOS translation @DalisonCPU

PR #591, opened 2020-07-11
Modify shortcuts

Oreonan:
-To increase user's voice volume : is now ctrl+right instead of ctrl+alt+right
-To decrease user's voice volume: is now ctrl+left instead of ctrl+alt+left
-Reasons of this changes are:
 - In some keyboards, ctrl+alt+arrow keys modify display orientation
 - This uniformizing behavior with QT client


bear101:
Hm, yes, good idea. But we need to update all *.lng files with this change.

Oreonan:
Better changes avoiding all the problems are in the works. The QT part is ready, I'm working on the Classic. I will open a new PR very soon. CC @ bear101

PR #582, opened 2020-07-06
Changing behavior of displaying status

Oreonan:

PR #596, opened 2020-07-12
Update french translation for QT

Oreonan:


bear101:
I've pushed out a new release: https://bearware.dk/?p=1073

Oreonan:
Thanks :)

You write: "The features that will go into TeamTalk v5.6 are now 
complete and it is now ready for testing.", but what about #582 and #591 ?

bear101:
I still need to investigate these. They can still go into the final release since they are still minor changes.

PR #606, opened 2020-07-23
Classic - Small changes on messages display behavior

Oreonan:


bear101:
With this change all messages in MessageDlg.cpp ends up on the same line. Also the space after nickname becomes formatted with an underscore. Unfortunately I didn't find an easy way of fixing it, so it will require a bit more investigate.

Oreonan:
@bear101 Normally I've fixed problem of messages in same line on f16442f, now we have a line return after all messages, like in chatrichedit.cpp.

PR #609, opened 2020-07-24
Define an accessible name to files list and add alt+f to directly go to files tab

Oreonan:


OzancanKaratas:
The channel list of classic client has been named as “Voice Activation Level”. Please fix it.

Oreonan:
Have you tested with latest beta 5.6.0.5004? And if yes, what is your 
language, here, with french translation, I've no problem.

OzancanKaratas:
Yes, I have the latest test version. Language is English.

Oreonan:
I've checked all files affected, and I've tested on my computer, label 
of channel list is "channel's list" in english.

I not understand your issue, maybe you can test with portable version 
and an empty xml?

OzancanKaratas:
I tested it by deleting the Lng_classic folder so that the language is processed by the source.

Oreonan:
OK, I've reproduced.

I'll try to fix it little more later, but it's a very strange bug.

Oreonan:
Normally fixed in #610

Client/qtTeamTalk/mainwindow.cpp
```diff
@@ -2348,9 +2349,13 @@ void MainWindow::updateChannelFiles(int channelid)
     ui.channelLabel->setText(tr("Files in channel: %1").arg(_Q(chanpath)));
 
     if(m_filesmodel->rowCount() == 0)
-        ui.tabWidget->setTabText(TAB_FILES, tr("Files"));
+        ui.tabWidget->setTabText(TAB_FILES, tr("&Files"));
+        ui.deleteButton->setVisible(false);
```
bear101, line 2353: Hm, does this compile? It doesn't look right

Oreonan:
Absolutly not :D, fixed in 682174a2

PR #610, opened 2020-07-25
Add some missing lines in rc

Oreonan:


Client/TeamTalkClassic/TeamTalk.rc
```diff
@@ -2209,6 +2209,8 @@ BEGIN
     ID_CHANNELS_STREAMMEDIAFILETOCHANNEL "Stream media file to channel"
     IDS_CHANHISTORYLAB "History"
     IDS_CHANMESSAGELAB "Message"
+    IDS_TREECTRLLAB            "Channels list"
```
bear101, line 2212: Shouldn't IDS_TREECTRLLAB use SetAccessibleName() like IDS_FILESLISTLAB does?

bear101: Ah, it does. Sorry...

PR #611, opened 2020-07-25
Modify behavior on file added

Oreonan:
Attempt to fix # 535, for which the real description should be: "name of user who added incorrect file on servers without user accounts"
Unfortunately I can't test in this time, so result it maybe catastrofic

### DIFF
--- a/Client/TeamTalkClassic/TeamTalkDlg.cpp
+++ b/Client/TeamTalkClassic/TeamTalkDlg.cpp
@@ -1854,7 +1854,7 @@ void CTeamTalkDlg::OnFileAdd(const TTMessage& msg)
        m_commands[m_nCurrentCmdID] != CMD_COMPLETE_JOIN)
     {
         PlaySoundEvent(SOUNDEVENT_FILES_UPDATED);
-        TT_GetUserByUsername(ttInst, remotefile.szUsername, &user);
+//        TT_GetUserByUsername(ttInst, remotefile.szUsername, &user);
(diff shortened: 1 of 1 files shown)


Oreonan:
Not working because UserID it's not returned when a file is added. Closing.

Client/TeamTalkClassic/TeamTalkDlg.cpp
```diff
@@ -1854,7 +1854,7 @@ void CTeamTalkDlg::OnFileAdd(const TTMessage& msg)
        m_commands[m_nCurrentCmdID] != CMD_COMPLETE_JOIN)
     {
         PlaySoundEvent(SOUNDEVENT_FILES_UPDATED);
-        TT_GetUserByUsername(ttInst, remotefile.szUsername, &user);
+//        TT_GetUserByUsername(ttInst, remotefile.szUsername, &user);
         CString szMsg;
         szMsg.Format(LoadText(IDS_FILEADDED), GetDisplayName(user), remotefile.szFileName);
```
bear101, line 1859: what will 'user' be initialized as after this change?

Oreonan:
Hmm, to be honest, I've not tested this change, I've only test similar 
changes on QT and I've always build errors.

Oreonan:
Tested now on QT with @Nardol

First time, user is initialized by strange unicodes characters, second 
time, it's initialized by NoName #ID.

So, it's not expected result :)

Issue #535, opened 2020-06-18
Probleme send file with teamtalk classic


math65:
When we send a file to the server, the pseudonym is not ours but a random one and always the same.

bear101:
You mean the owner of the file? The owner of the file will be your username.

--
Kind regards

Bjørn Damstedt Rasmussen
www.bearware.dk

bear101:
No response. Closing...

Oreonan:
I've reproduced this issue, shis only present on a server who have no user. I've found cause of this issue and I'm trying to fix it.

PR #612, opened 2020-07-26
Modify behavior for files events

Oreonan:
Unfortunately on QT we are same problem like on classic on server without user account.
Server don't return user ID when an user added or removed a file, but only username... CC @bear101


Oreonan:
Fixed #535 by f8843d9

PR #613, opened 2020-07-27
Add more history events on QT

Oreonan:
This make same behavior on QT like on Classic and Android clients for example

### DIFF
--- a/Client/qtTeamTalk/mainwindow.cpp
+++ b/Client/qtTeamTalk/mainwindow.cpp
@@ -1011,31 +1011,10 @@ void MainWindow::processTTMessage(const TTMessage& msg)
         processTextMessage(msg.textmessage);
         break;
     case CLIENTEVENT_CMD_FILE_NEW :
-    {
-        Q_ASSERT(msg.ttType == __REMOTEFILE);
-        const RemoteFile& file = msg.remotefile;
-        User user; 
-        //only update files list if we're not currently logging in or 
-        //joining a channel
-        cmdreply_t::iterator ite = m_commands.find(m_current_cmdid);
-        if(m_filesmodel->getChannelID() == file.nChannelID &&
-           (ite == m_commands.end() || (*ite != CMD_COMPLETE_LOGIN && 
-                                        *ite != CMD_COMPLETE_JOINCHANNEL)) )
-        {
-            updateChannelFiles(file.nChannelID);
-            playSoundEvent(SOUNDEVENT_FILESUPD);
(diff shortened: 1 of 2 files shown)


Client/qtTeamTalk/mainwindow.cpp
```diff
@@ -753,6 +753,7 @@ bool MainWindow::parseArgs(const QStringList& args)
 void MainWindow::processTTMessage(const TTMessage& msg)
 {
     bool update_ui = false;
+    QString rootchanname = _W(tr("root"));
```
bear101, line 756: I don't think the _W macro is needed here

Client/qtTeamTalk/mainwindow.cpp
```diff
@@ -1100,6 +1131,13 @@ void MainWindow::processTTMessage(const TTMessage& msg)
                m_talking.size() == 0)
                 m_sysicon->setIcon(QIcon(APPTRAYICON_CON));
         }
+
+        if(user.uUserState & USERSTATE_MEDIAFILE_AUDIO && user.nUserID != TT_GetMyUserID(ttInst))
+        {
+            User nameuser;
```
bear101, line 1137: You don't need to extract user as "nameuser". Just use existing "user".

Btw, if you have a C-struct in C++ then initialize it like this: "User user = {};" Then all data in the struct is initialized to zero (this is important for strings since they are zero terminated).

Client/qtTeamTalk/mainwindow.cpp
```diff
@@ -940,26 +941,57 @@ void MainWindow::processTTMessage(const TTMessage& msg)
...
         Q_ASSERT(msg.ttType == __USER);
         if(msg.user.nUserID == TT_GetMyUserID(ttInst))
             processMyselfJoined(msg.user.nChannelID);
         emit(userJoined(msg.user.nChannelID, msg.user));
+        if(m_commands[m_current_cmdid] != CMD_COMPLETE_LOGIN) {
+            if(msg.user.nUserID != TT_GetMyUserID(ttInst)) {
+                Channel chan = {};
+                ui.channelsWidget->getChannel(msg.user.nChannelID, chan);
+                QString userjoinchan = tr("%1 joined channel").arg(getDisplayName(msg.user));
+                if(chan.nParentID == 0 && msg.user.nChannelID != TT_GetMyChannelID(ttInst))
```
bear101, line 968: Why the "nChannelID != TT_GetMyChannelID(ttInst)"?

Oreonan:
Because another message already exists if someone join same channel.

PR #628, opened 2020-08-10
Display admin and channel op in channels tree

Oreonan:


Client/qtTeamTalk/channelstree.cpp
```diff
@@ -969,6 +969,12 @@ void ChannelsTree::slotUpdateTreeWidgetItem(QTreeWidgetItem* item)
         if (user.nStatusMode & STATUSMODE_STREAM_MEDIAFILE)
             itemtext += tr(" (Streaming media file)");
 
+        if(user.uUserType & USERTYPE_ADMIN)
+            itemtext += tr(" [Administrator]");
```
bear101, line 973: Why brackets instead of parenthesis?

Oreonan:
Because I've wanted to separate informations on status and on rights. 
But if you prefere I can use parenthesis for all?

Client/qtTeamTalk/channelstree.cpp
```diff
@@ -969,6 +969,12 @@ void ChannelsTree::slotUpdateTreeWidgetItem(QTreeWidgetItem* item)
         if (user.nStatusMode & STATUSMODE_STREAM_MEDIAFILE)
             itemtext += tr(" (Streaming media file)");
 
+        if(user.uUserType & USERTYPE_ADMIN)
+            itemtext += tr(" (Administrator)");
```
bear101, line 973: Shouldn't this also be comma separated?

bear101: Ah, just read your comment. Ok

PR #632, opened 2020-08-13
Changing behavior for move

Oreonan:
- Return if an user is selected or no on CTRL+G and in the menu
- Allow to remove an user from list of selected users


Client/TeamTalkClassic/TeamTalkDlg.cpp
```diff
@@ -5717,14 +5717,38 @@ void CTeamTalkDlg::OnServerSaveconfiguration()
...
 void CTeamTalkDlg::OnAdvancedStoreformove()
 {
     int nMoveUserID = m_wndTree.GetSelectedUser();
-    if (nMoveUserID)
-        m_moveusers.insert(nMoveUserID);
-
+    if(nMoveUserID) {
+        if(m_moveusers.find(nMoveUserID) != m_moveusers.end()) {
+            m_moveusers.erase(nMoveUserID);
+            if(m_xmlSettings.GetEventTTSEvents() & TTS_USER_JOINED) {
```
bear101, line 5734: The move action doesn't really have to do with the TTS event TTS_USER_JOINED. There should probably be a new TTS event for this action.

I saw in another PR you made that you use existing TTS events but there should probably be separate. Alternatively we should have a general TTS event for stating menu action.

Client/TeamTalkClassic/TeamTalkDlg.cpp
```diff
@@ -5717,14 +5717,38 @@ void CTeamTalkDlg::OnServerSaveconfiguration()
...
+                User user;
+                TT_GetUser(ttInst, nMoveUserID, &user);
+                CString szMsg;
+                szMsg.Format(LoadText(IDS_UNSELECTFORMOVE, _T("%s unselected for move")), GetDisplayName(user));
```
bear101, line 5738: Use "deselected" instead of "unselected"

Oreonan:
Hmmm, I'm not really sure to know how to add an option :), but I can try :)

Client/TeamTalkClassic/TeamTalkDlg.cpp
bear101: Should I make a new TTS group called "Menu actions"? Then it can be used for move-action and your other PR with mute on/off.

Oreonan:
Oh yes, if you can do this it can be really apreciated :)

I'll update my PR with this new option after youa ddition, and I'll see 
how to do this for the futur :)

PR #642, opened 2020-08-22
613 without conflicts :)

Oreonan:


Client/qtTeamTalk/mainwindow.cpp
```diff
@@ -922,7 +922,7 @@ void MainWindow::processTTMessage(const TTMessage& msg)
             //update AGC, denoise, etc. if changed
             updateAudioConfig();
 
-            update_ui = true;
```
bear101, line 925: Why not just have one "slotUpdateUI()" at the bottom instead of calling it in every case-label? Also here the slotUpdateUI(); is called prior to invoking the event so actually the menu-items will not have been updated.

Oreonan:
Like this: b301c1ac ?

bear101:
Yes, I think it's a good idea you leave out the boolean check since it was missing in some cases. I'll make a small correction where we don't update the UI if there's an active command.

PR #644, opened 2020-08-25
QT - same behavior when activate VA via me menu and via hotkey

Oreonan:


Client/qtTeamTalk/mainwindow.cpp
```diff
@@ -1871,9 +1871,17 @@ void MainWindow::hotkeyToggle(HotKeyID id, bool active)
 #endif
         break;
     case HOTKEY_VOICEACTIVATION :
-        if(active)
+        if(active) {
             TT_EnableVoiceActivation(ttInst,
```
bear101, line 1875: Minor thing (I'll change it); In Qt the default code style is curly brace on next line.

PR #646, opened 2020-08-30
Classic - Translate credits

Oreonan:


Client/TeamTalkClassic/resource.h
```diff
@@ -458,6 +458,12 @@
 #define IDS_DEFAULTNICKNAME             556
 #define IDS_TREECTRLLAB                 557
 #define IDS_FILESLISTLAB                558
+#define IDS_ABOUT_CREDITS               900
```
bear101, line 461: Remember to put in STRINGTABLE as well in TeamTalk.rc

bear101:
It's actually not mandatory to put a STRINGTABLE but it gets quite difficult to know which IDs are currently in use. I've added a couple of missing ones in this commit: 7942e6785987933f1f6840f1711a3f25832874af

Oreonan:
Yes, I'm agree :), another problem is to know what items are non use in 
all lng files, I'll try to clean lng files and complete rc little more 
later.

bear101:
Yes, it's especially difficult in the Korean and Chinese :)

--
Kind regards

Bjørn Damstedt Rasmussen
www.bearware.dk

PR #647, opened 2020-08-30
update chinese translation

hejiadong0608:


Client/TeamTalkAndroid/src/main/res/values-zh/strings.xml
```diff
@@ -325,6 +325,9 @@
     <string name="permission_read_phone_state">来电监测将不可用</string>
     <string name="permission_bluetooth">蓝牙耳机麦克风将不可用</string>
 
+    <string translatable="false" name="facebook_app_id">611833262539381</string>
```
bear101, line 328: These two should not be included. They are not translatable. Otherwise it might cause problems if facebook-ID changes later.

hejiadong0608: These strings have been deleted.

PR #653, opened 2020-09-06
TTS events for messages send

Oreonan:


bear101:
Since it's a new feature, let's wait until v5.6.1 is completed before merging. I also have two branches that are waiting

PR #658, opened 2020-09-07
Android - Translate developer string on about dialog

Oreonan:


poretsky:
This is incorrect.

Oreonan:
What is the problem please ?

poretsky:
android:title="Beqa Gozalishvili, @string/pref_title_dev"

Resource is not referenced in such constructs, but string is rendered
literally.

I'd suggest to check pull requests before submitting.

Oreonan:
OK, I'm sorry. And are you an idea to do this orrectly ?

Yes, I'm totally agree with your suggest, but unfortunately I can't test 
Android client, I'll try to understand why little more later.

bear101:
I've reverted commit ca4308aac. It was me who merged the pull request without checking properly. I only checked the translation editor.
I guess strings in the xml-resources only work if the entire string is a string-resource.

poretsky:
It could be done by constructing actual string in java code, but on the
other hand, I think, the word "developer" can be safely removed because
all items in the category have this identical postfix. And, since the
category itself actually lists developers (not all contributors), it
should be titled as "Developers". Strictly speaking, developers and
translators are both contributors.

Oreonan:
I've just replaced contributors by developers, I've also updated all 
translations affected, can you check if it's ok for russian ? Thanks.

poretsky:
You mean 5fa23bf3? But you didn't touch Russian string and it is
absolutely correct. Thanks.

PR #674, opened 2020-10-25
Continue to add some emojis in channels tree of QT

Oreonan:


Client/qtTeamTalk/channelstree.cpp
```diff
@@ -954,7 +954,9 @@ void ChannelsTree::slotUpdateTreeWidgetItem(QTreeWidgetItem* item)
 
         QString itemtext;
         QString name = getDisplayName(user);
-        itemtext = name;
+        if(emoji && item->data(COLUMN_ITEM, Qt::UserRole).toInt() & MESSAGED_TYPE)
+            itemtext += "✉ ";
```
bear101, line 958: Shouldn't the "letter" emoji come after the username?

Oreonan:
I'm not really sure, I think it can be better to have info of pending 
message first of username. But if you prefere, I can move it :)

PR #687, opened 2020-12-20
Change some sounds

Oreonan:
Change some sounds, I think these will be better audible and differentiables

New sound found here: https://lasonotheque.org/search?duree=-1&page=3


Oreonan:
I'll make some aditional tests with some users. I'll reopen more later finally... :)

PR #688, opened 2020-12-21
Don't play sound for our channel messages

Oreonan:
Uniformize behavior bitween channel and private messages


bear101:
I guess this should also be the case for Qt client?

Oreonan:
@bear101 Yes, you are right, ef82ab7

ChrisDuffley:
I much prefer the channel message sound being played  as there may be several people who have their TTS events turned off or their speech may be muted on its own for screen reader users. I strongly recommend leaving it as is.

Christopher Duffley
Cell: +1 (603) 851-6971

Oreonan:
In fact, the idea is to standardize the behavior between channel 
messages and private messages, and therefore to leave the sound only for 
messages from other people than ourselves.

ChrisDuffley:
O, that makes sense, thanks to the subject.

Christopher Duffley
Cell: +1 (603) 851-6971

PR #691, opened 2020-12-23
Updated polish and croatian translations

zstanecic:
Updated polish and croatian translation files


bear101:
Thanks a lot for the updated translation! I hope to send out an alpha-release of TeamTalk v5.7 in early January.

PR #693, opened 2020-12-29
Updated polish and croatian translations

zstanecic:
Updated polish and croatian translations. 
Note: to derive the context of the strings and improve the translation, i will need the beta, or know a way how to compile the beta locally.


Oreonan:
I think an alpha it's on the way and it will publish in earlyer january

PR #704, opened 2021-01-05
New sound events

Oreonan:
- On classic:
* User logged in
* User logged out
- On QT:
* User logged in
* User logged out
* Voice activation enabled
* Voice activation disabled
- New logged_on.wav and logged_off.wav


Client/TeamTalkClassic/TeamTalkBase.h
```diff
@@ -306,7 +306,9 @@ enum SoundEvent
     SOUNDEVENT_NONE = 0,
     SOUNDEVENT_USER_JOIN                    = 0x00000001,
     SOUNDEVENT_USER_LEFT                    = 0x00000002,
+    SOUNDEVENT_USER_LOGGED_IN               = 0x00000003,
```
bear101, line 309: This is a bitmask. SOUNDEVENT_USER_LOGGED_IN would be the same as (SOUNDEVENT_USER_LEFT | SOUNDEVENT_USER_JOIN). I'll fix it.

PR #718, opened 2021-01-11
Fix MP order on classic

Oreonan:


bear101:
Hm, some thing in the formatting is not right. Before the username/nickname was in bold, blue and underscore. Now it only appears on the two line. Can you fix it or should I look into it?

Oreonan:
Ouch, it's difficult to me to fix it :)

Maybe you must have to define cf before insering content.

I can make a commit if you want, but I don't guaranty the result :(

bear101:
I'll see if I can simply it. Maybe just have own messages in one color and other users in default

Oreonan:
OK :), let me know if you want I do anything :)

bear101:
Please look at the sent sound message instead in Classic client

Oreonan:
Euh, hmmmm, OK, if you want :), I can do this, but I'm currently talking 
with some users who prefere have separate event for MP and public 
message send.

I must have to add an event for each type of outgoing message? What is 
problem with checkbox? Maybe a GUI issue? :)

I only had it checked once and then committed without having my changes 
checked, maybe it's still broken?

bear101:
The checkbox didn't really fit in the layout because it was a grid layout. All other controls were using a different layout.
We can also have two "sent message" sounds but isn't it obvious since you're either in a text-message-dialog or in main window.

Oreonan:
Other solution is to move checkbox in another tab, but it's probably 
litle bit out of context :)

I let you choose :)

bear101:
let's make two sound events for "sent message". One for private and one for public.

bear101:
by default they can point to the same .wav file in the ini-config file

Oreonan:
My idea is to don't have default sound for this new events. Like my 
checkbox :), there was not checked by default. I'll make a commit this 
evening or tommorow.

bear101:
I've tried to improve the formatting so it's ok now. But what about scrolling? It doesn't scroll automatically

Oreonan:
Oh, I don't know. To say right for me it's not a problem, but if it's 
necessary maybe we must have to retrieve a part of code from 
chatrichedit.cpp to make scrolling automatic.

bear101:
Hm, yes, why don't we just use CChatRichEditCtrl::AddMessage() in CMessageDlg::AppendMessage() ?

Oreonan:
I've don't know if that it was possible. But if yes, it probably the 
better solution :)

bear101:
Ok, I'll do that instead. I had forgotten about it.

ChrisDuffley:
That's how it has been for received messages, both channel and private. It honestly wouldn't matter for sent messages because, as said earlier, it's pretty obvious from what window you're in what type of message you're sending.

Christopher Duffley
Cell: +1 (603) 851-6971

bear101:
@Oreonan please try now and merge if you're satisfied

beqabeqa473:
i am not sure i am writing to correct topic, i didn't look at the code.

here you were talking about scrolling of messages as i understand.

i wanted to add several words in terms of usability from point of view of screenreader user.

if someone sent a text in public channel chat, and you are trying to read it with arrow keys, and at the middle of history someone joins to a channel or simply tt will force history to be updated, cursor will jump to the end.

it is very hard to return to the previous position after that.

ChrisDuffley:
That's a very valid point.

Christopher Duffley
Cell: +1 (603) 851-6971

Oreonan:
Yes, it's very difficult to return to previous position when cursor is 
move to the end. I just confirm. But this is another point which require 
another commit or PR :)

Oreonan:
@bear101 there is a problem, we don't have message timestamp.

Except that all it's ok for me.

PR #721, opened 2021-01-12
Option to play sounds for outgoing messages: Fixed #720

### LINKED ISSUE #720: Optional sound when sent a channel or an user message
Hi!

In TeamTalk 5.7 beta classic version there are new sounds that are very good, and as a new feature, TTS says when a private or channel message sent.

Can you add options to enable sound on channel and user message sent? Channel message sent sound was very good in earlier clients to detect lags in TCP connection, and easier to hear the message sent instead of TTS says it.

Oreonan:
I just hope GUI is not broken.


bear101:
I changed the Qt client so sent message is a separate sound event. I think this is simpler to implement since you can use existing functionality.

Oreonan:
So I must have to do the same on classic?

bear101:
I would suggest the same way

bear101:
> So I must have to do the same on classic?

Yes, please :)

Oreonan:
This have only one inconvenient, sound for private and channel messages 
send is the same. The checkbox respected diference bitween two events :)

Maybe we need to add a different event for each message sent?

ChrisDuffley:
Don't we already have a checkbox for each event separately, unless this is only for TTS in Classic? We'll want both private and channel messages separate I believe.

Christopher Duffley
Cell: +1 (603) 851-6971

Oreonan:
We have currently a sound event for channel message, and for private 
message, and a TTS event for each type of message also.

ChrisDuffley:
Correct. I'd keep it that way, having 2 different sound events that can be toggled in Prefs would probably be most preferred.

Christopher Duffley
Cell: +1 (603) 851-6971

PR #742, opened 2021-01-25
Jitter buffer

Lauriumora:
This PR contains three jitter-related new features.
-A fixed and adapative jitter buffer, controller by a new API call
-A new TT_SetUserMediaStorageDirEx API function that allows an extra delay on the closing of voice recordings so they can capture all audio in a stream even with heavy jitter
-Support for a %starttick% variable in the recording file format to facilitate post-record mixing of the recording into a jitter-free mixed channel recording


Library/TeamTalk_DLL/TeamTalk.h
```diff
@@ -6915,6 +6915,40 @@ extern "C" {
...
+     * @param nUserID The user ID of the user to apply the configuration to.
+     * @param nStreamType The type of stream to change, currently only
+     * #STREAMTYPE_VOICE is supported. Other types are a no-op.
+     * @param nFixedDelayMSec The fixed delay in milliseconds. Default = 0.
+     * @param bMute Turns adaptive jitter buffering ON/OFF. Default is OFF.
+     * @param nMaxAdaptiveDelayMSec A hard maximum on the adaptive delay.*/
+     TEAMTALKDLL_API TTBOOL TT_SetUserJitterControl(IN TTInstance* lpTTInstance,
+                                                    IN INT32 nUserID,
+                                                    IN StreamType nStreamType,
+                                                    IN INT32 nFixedDelayMSec,
```
bear101, line 6948: Should we use a struct here instead for the last three parameters? Then it's easier to extend later with more features.

Lauriumora: Changed this in an additional commit. 
I have no idea why the build on Github fails. The test that fails runs perfectly on this jitter_buffer branch as well as the master branch on Ubuntu18.04 and Windows10

Lauriumora:
Those commits were solid changes, but I still don't see why the test failed on it (and succeeds locally)

Lauriumora:
Ah, sorry I see your comment now. It was the allow_record. Nice catch

PR #757, opened 2021-02-10
Option to display server name on window's title when in a channel

Oreonan:
It's probably not a typical case of use, but I know some users who have connected on multiple servers on the same time, so it can be useful, maybe.


Client/TeamTalkClassic/TeamTalk.rc
```diff
@@ -288,6 +288,8 @@ BEGIN
     CONTROL         "Always on top",IDC_CHECK_ALWAYSONTOP,"Button",BS_AUTOCHECKBOX | BS_MULTILINE | WS_TABSTOP,15,93,117,16
     CONTROL         "Double-click for joining channels",IDC_CHECK_DBCLICKJOIN,
                     "Button",BS_AUTOCHECKBOX | BS_MULTILINE | WS_TABSTOP,15,114,117,16
+    CONTROL         "In channels, show the name of the server in the window's title",IDC_CHECK_SERVNAMEDISP,
```
bear101, line 291: Can we change the label to "Show only server name as window title" and default is unchecked.

Then default value for "ClientXML::GetDisplayServerName()" should be FALSE.

Otherwise the title is a big confusing

Client/TeamTalkClassic/settings/ClientXML.h
```diff
@@ -197,6 +197,8 @@ namespace teamtalk {
 
         bool SetJoinDoubleClick(bool bEnable);
         bool GetJoinDoubleClick();
+        bool SetDisplayServerName(bool bEnable);
+        bool GetDisplayServerName();
```
bear101, line 201: Change name to "GetDisplayServerNameOnly(bool bEnable = false)"

Oreonan:
Euh, I'm not agree, initial idea is not to only display server name, but 
server name and channel name.

Currently, if we are in a channel "test" on a server "blabla", window's 
title is: "test - teamtalk V5.7" for example.

With this change, window's title will can be "blabla/test - teamtalk v5.7"

bear101:
@Oreonan I've rephrased the wording to "Show both server and channel name in window title". If you agree then it's fine to merge it in.

PR #758, opened 2021-02-11
New sounds options for classic

Oreonan:
- Work did:
* Always allow to refresh devices
* Ability to choose sounds playing mode
* For TT audio mode, ability to choose sounds volume
- Todo:
* Check and fix GUI
* If possible, only display slider to choose sounds volume if TT audio mode is selected


Client/TeamTalkClassic/TeamTalkDlg.cpp
```diff
@@ -6415,10 +6421,14 @@ void CTeamTalkDlg::PlaySoundEvent(SoundEvent event)
         szFilename = STR_UTF8(m_xmlSettings.GetEventTransmitQueueStop());
         break;
     }
-
     if (szFilename.GetLength())
     {
-        m_pPlaySndThread->AddSoundEvent(szFilename, PLAYBACKMODE_TEAMTALK);
+        if (PlaybackMode == 1)
```
bear101, line 6426: Instead of comparing hardcoded integer values use the enum PLAYBACKMODE_*

Oreonan:
Euh, I don't know how to use this in this case... :(

Client/TeamTalkClassic/TeamTalk.rc
```diff
@@ -270,6 +270,12 @@ BEGIN
     CONTROL         "",IDC_SLIDER_MEDIASTREAM_VOL,"msctls_trackbar32",TBS_BOTH | TBS_NOTICKS | WS_TABSTOP | 0x400,117,176,131,15
     PUSHBUTTON      "Refresh devices",IDC_BUTTON_REFRESHSND,132,193,60,12
     PUSHBUTTON      "&Default",IDC_BUTTON_DEFAULT,197,193,51,12
+    GROUPBOX        "Client sounds events settings",IDC_STATIC_GRPCLIENTSOUNDSYSTEM,200,7,246,207
+    CONTROL         "Play all sounds simultaneously",IDC_RADIO_PBMODETT,"Button",BS_AUTORADIOBUTTON | WS_TABSTOP,41,22,95,10
```
bear101, line 275: There's no room for all these radio buttons. We should use a combobox with a droplist instead. You can find an example of this in Display-tab where you choose how to sort channels.

I can also look into it but it will take a few days before I can look at it.

bear101:
@Oreonan  I'll force push and fix conflicts

bear101:
@Oreonan there was no space left in the "Sound System"-tab so I moved this new "playback mode feature" to "Sound Events"-tab.

I also had to rephrase the text for the combobox because it would otherwise be too big. You should probably update the French translation as well of strings: IDS_FULLY, IDS_STOPPREVIOUS, IDS_SIMULTANEOUSLY.

Note that when you use combobox in MFC then the default option is to sort items. This causes problems when we have translations. Often the right way is to use item-data to map items.

bear101:
Can you give it a try and see if you're satisfied?

Oreonan:
It's OK for me, I'll update translations when it'll merge in master.

PR #762, opened 2021-02-15
Turkish update

OzancanKaratas:
@Oreonan


Client/qtTeamTalk/languages/Turkish.ts
```diff
@@ -1,6 +1,4 @@
-<?xml version="1.0" encoding="utf-8"?>
-<!DOCTYPE TS>
-<TS version="2.1" language="tr">
+<?xml version="1.0" ?><!DOCTYPE TS><TS language="tr" version="2.1">
```
Oreonan, line 1: Why this change? QT translation tool will be replace this one line by 3 lines at next update I think.

OzancanKaratas: No problem. It will work correctly. You can merge.

PR #765, opened 2021-02-16
Jitter buffer api expansion

Lauriumora:
I've expanded the API relating to the Jitter buffer. Please check if the python addition is OK.
The API changes are done to support storing and re-applying the jitter buffer values after a reconnect of a user


Library/TeamTalk.NET/TeamTalk.cs
```diff
@@ -2113,6 +2113,10 @@ public struct User
          * content.
          * @see TeamTalkBase.SetUserAudioStreamBufferSize() */
         public int nBufferMSecMediaFile;
+        /** @brief the currently active adaptive jitter delay for
+        received voice streams for this user.
+        * @see TT_SetUserJitterControl */
+        public int nActiveAdaptiveDelayMSec;
```
bear101, line 2119: How come you only want this value in User-struct and not the whole JitterConfig-struct?

Lauriumora: Because I didn't want to overcrowd the User-struct and because the other JitterConfig elements are fixed application-provided values (so the app should already know them). Only this element changes dynamically.

PR #766, opened 2021-02-16
tt5 polish and croatian translation

zstanecic:
updated latest strings for classic


Oreonan:
More strings will still be added, so it's best to wait a bit before merging this

PR #769, opened 2021-02-18
More aggressive adaptive delay

Lauriumora:
PR with two commits relating to jitter buffering
1 - more aggressive adaptive jitter buffering
2 - new clientevent for the first received voice packets to allow apps to follow/visualize delay introduced by the jitter buffer


Library/TeamTalk.NET/TeamTalk.cs
```diff
@@ -3496,7 +3496,27 @@ public enum ClientEvent : uint
...
+        * of the voice playout that is notified via
+        * #CLIENTEVENT_USER_STATECHANGE due to jitter buffering.
+        *
+        * The time between #CLIENTEVENT_USER_FIRSTVOICESTREAMPACKET
+        * and #CLIENTEVENT_USER_STATECHANGE is the fixed jitter delay
+        * configuration plus the currently active adaptive jitter
+        * buffering in the nActiveAdaptiveDelayMSec member of the
+        * User struct
+        *
+        * @param nSource 0
```
bear101, line 3514: Can 'nSource' be the stream-ID? Because then we can map the new voice stream in CLIENTEVENT_USER_AUDIOBLOCK.

bear101: I can do it if you just say ok.

Lauriumora: Yes, you're right. Will you add it?

bear101: yup

bear101: I'll modify .NET and Java API tomorrow

Lauriumora:
I'll have a look at the failing unit test ( REQUIRE( delay >= fixeddelay ) ). 
I don't get the valgrind results however. I don't really see anything TT-related in those results

bear101:
I think the timeout is because GitHub uses sound-device-id 1978 (virtual sound device). It's much faster than ALSA.

Try to reproduce the issue by passing sound-device-id [1978](https://github.com/BearWare/TeamTalk5/blob/da86a881460f6e9b037aa4143c9beb577f43bc11/Library/TeamTalk_DLL/TeamTalk.h#L511) here:
https://github.com/BearWare/TeamTalk5/blob/da86a881460f6e9b037aa4143c9beb577f43bc11/Library/TeamTalkLib/test/CatchDefault.cpp#L1655

Lauriumora:
No, it has to be something else because the test now fails here locally as well. 
Also, the sound device performance is not a factor in this test.

Lauriumora:
Hmm, if I run the test 20 times it fails about 5 times and with all different timings. 
The test measures the time between two client events and I think the test timing is just not reliable:
1) The CLIENTEVENT_USER_FIRSTVOICESTREAMPACKET event is enqueued after the packet is processed.
2) The test itself doesn't measure the *exact* moment of receiving CLIENTEVENT_USER_FIRSTVOICESTREAMPACKET.

My guess is that the queuing of the client events is the main culprit and that makes these types of stringent timing tests impossible. We should either remove the test or relax the timing.
What do you think?

Lauriumora:
The test fails with both under- and overtiming. It's just a test we can't do because we're essentially measuring the thread scheduling performance. I'm removing the timing checks

bear101:
I've also had many issues with timing. It's hard to get right on a
virtual environment like GitHub's backend.

Kind regards

Bjoern Damstedt Rasmussen
www.bearware.dk

bear101:
This test also fails once in a while due to timing of threads:
https://github.com/BearWare/TeamTalk5/blob/bade9b1b136c184bcc2b31fb2b2203ba2cbe727e/Library/TeamTalkJNI/test/dk/bearware/TeamTalkTestCase.java#L1161

It should really be avoid to do any kind of timing. Only test on events.

Lauriumora:
Ok, all tests should pass after that commit. Will you merge into master when it's done?

PR #784, opened 2021-02-26
Sound events for "mute all", Fixed #782

### LINKED ISSUE #782: Sound events for "mute all" in Classic client
Second, we found there are no hint tones for open and recover mute all during our use. Could you please add hint tones for these two, just like the ones for open and close voice activated? Many people have no awareness when they keep mute all on during their use, especially visually impaired, for they can`t see, they use this software totally on hint tones.

@Oreonan maybe one you can easily do and have some sound files available.

Oreonan:


bear101:
Thanks @Oreonan. Fixes #782

PR #799, opened 2021-03-08
TTAndroid - Accessibility hint for away status

Oreonan:


Oreonan:
Hmmmm, I made a mistake, now if an user is away, we only have "away" announce in his nickname, we don't have his nickname announce anymore. @bear101 Do you think moving the away status check before other conditions can correct the problem?

Client/TeamTalkAndroid/src/main/java/dk/bearware/gui/MainActivity.java
```diff
@@ -1138,9 +1138,8 @@ else if(item instanceof User) {
                 if(user.nUserID == ttservice.getTTInstance().getMyUserID()) {
                     talking = ttservice.isVoiceTransmitting();
                 }
-                
+                String name1 = Utils.getDisplayName(getBaseContext(), user);
```
bear101, line 1141: Why do we extract "name1" when we have "name" already?

Client/TeamTalkAndroid/src/main/java/dk/bearware/gui/MainActivity.java
```diff
@@ -1158,6 +1157,9 @@ else if(item instanceof User) {
                         icon_resource = away? R.drawable.man_orange : R.drawable.man_blue;
                     }
                 }
+                if(away) {
+                    nickname.setContentDescription(getString(R.string.user_state_away, name1));
```
bear101, line 1161: Shouldn't it be "status" instead of "nickname"?

Oreonan:
Ah, I didn't know it worked too. I will do that

PR #806, opened 2021-03-13
Set the WebRTC AEC delay to the capture/render delay provided by Port…

Lauriumora:
…Audio


bear101:
Good observation with the clock skew and very happy that PortAudio reports the Time-info.

Lets hope we don't need to port the WebRTC echo canceller to Android and iOS because then we have to remember to take this clock skew into account.

PR #811, opened 2021-03-25
Korean translation for iOS  and Classic client

ajackpot:
Hello?
I created a Korean translation file for iOS TeamTalk and updated the Korean translation file for Windows Classic client to match version 5.7.
In the case of the TeamTalk Korean translation file for iOS, the line ending string is CR+LF, so conversion is required. 
Thank you very much.


bear101:
Great! Thanks a lot. I'll include the translations in the next release.

PR #813, opened 2021-03-30
Webrtc improve delay accuracy

Lauriumora:


Lauriumora:
I had a separate PR for this first commit, but then based the branch for the next commit on that by accident. 
Decided to leave it like this and closed the other PR
So each commit in this PR is really a separate issues in this PR.

Library/TeamTalkLib/avstream/WebRTCPreprocess.cpp
```diff
@@ -97,6 +93,17 @@ int WebRTCPreprocess(webrtc::AudioProcessing& apm, const media::AudioFrame& infr
...
+            // As per audio_processing.h, set_stream_delay_ms should be called after ProcessReverseStream
+            int ret = apm.set_stream_delay_ms(delayms);
+
+            MYTRACE_COND(ret != webrtc::AudioProcessing::kNoError,
+                ACE_TEXT("WebRTC failed to set stream delay for echo cancellation. Result: %d, delay: %d\n"), ret, infrm.duplex_callback_delay);
+            if (ret != webrtc::AudioProcessing::kNoError)
+                return -1;
+
+            //WebRTC is fed in 10ms frames, so increment the delay 10ms for every WebRTC frame
+            delayms += 10;
```
bear101, line 106: I don't quite understand this increment of 10. Isn't the set_stream_delay_ms() function an offset? Shouldn't all 10msec frames have the same offset?

Library/TeamTalkLib/teamtalk/client/AudioThread.cpp
```diff
@@ -499,7 +499,17 @@ void AudioThread::ProcessAudioFrame(media::AudioFrame& audblock)
 
 #if defined(ENABLE_WEBRTC)
     bool vad = false;
-    PreprocessWebRTC(audblock, vad);
+    if (m_gainlevel > 0)
```
bear101, line 502: Good idea. There's not much processing to do on nothing.

Library/TeamTalkLib/avstream/WebRTCPreprocess.cpp
Lauriumora: Yes, you're right. This is simply a mistake on my part. I'll remove the +10.
I'm chasing a gradually failing AEC and tried this.

PR #818, opened 2021-04-05
Improvements for status events on channel joined and left

Oreonan:


Client/qtTeamTalk/mainwindow.cpp
```diff
@@ -2472,8 +2490,29 @@ void MainWindow::processMyselfJoined(int channelid)
 
 void MainWindow::processMyselfLeft(int channelid)
 {
-    Q_UNUSED(channelid);
-    m_mychannel = {};
+    TT_GetChannel(ttInst, channelid, &m_mychannel);
```
bear101, line 2493: This call may fail if user is last in channel and channel is removed.

Also "m_mychannel" has to be reset otherwise the client will appear to still be in a channel (m_mychannel will appear valid).

bear101:
I made a change where I "reuse" m_mychannel prior to resetting it.

Oreonan:
Yes, it's good for me :), thanks.

PR #822, opened 2021-04-12
Fix wav playout/streaming on Wndows 7

Lauriumora:


bear101:
I'm not able to test on Windows 7 but I take it you have tested.

Users have reported a similar issue when TT requests to transcode video formats (using MFT). It's probably the same issue. I'll try to investigate this later.

Lauriumora:
What are these failed builds? Is there something I should have done? The committed code certainly doesn't hit iOS or MAC

Lauriumora:
Yes, I've tested it on Win7 and WAV playout works after this commit

bear101:
It's the package tool on macOS (homebrew) which suddenly won't download a package. I guess it's a temporary GitHub issue.

PR #825, opened 2021-04-13
QT - Improve accessibility of stream dialog

Oreonan:


Client/qtTeamTalk/streammediafile.ui
```diff
@@ -42,7 +45,7 @@
            <string>Browse</string>
           </property>
           <property name="text">
-           <string>...</string>
+           <string>... &amp;B</string>
```
bear101, line 48: Should we instead change this to an icon as well (same as Refresh button)? It looks a bit weird with "...B"

Oreonan:
If you think that it's can be better, I don't see any inconvenient.

bear101:
But it's easy enough to know that "R" is a shortcut to "Refresh" although it's an icon?

Oreonan:
Currently, for "refresh" button, screenreader says: "Refresh button 
alt+R", so we know what shortcut is for what button :)

bear101:
Ok, then I change it to icon.

--
Kind regards

Bjørn Damstedt Rasmussen
www.bearware.dk

bear101:
Do we need the three dots for "R"? I.e. "...&R"

Oreonan:
As you can removed this for B, it's more logic to remove also for R.

PR #826, opened 2021-04-13
QT - Confirmation before removing an user account

Oreonan:


bear101:
@Oreonan  Hm, it's best to only update translation files on one branch.

Oreonan:
We have conflicts? Yes I'll maybe update only in master branch in the 
futur :)

bear101:
Luckily kdiff3 managed to merge everything automatically.

PR #830, opened 2021-04-14
QT - autofocus for some radio buttons

Oreonan:


Oreonan:
Ah, no, bad idea, with this change radiobutton take focus, I just wanted to find a way to place focus is checked radiobutton when we go on it using tabulation, not at everytime we switch to a tab with radiobutton in preferences for example. Some tests are required to find a better solution.

PR #836, opened 2021-04-24
Alert when intercepted from another channel

Oreonan:


bear101:
I've also fixed conflicts + some errors, so I'll merge that instead. Sorry for not notifying

Oreonan:
There is no problem. I'll work on translations :)

PR #848, opened 2021-04-28
QT - Join last channel on reconnect after drop

Oreonan:


Oreonan:
To be more precise, previously this is not working for root channel, ow it works for all channels.

bear101:
Doesn't the classic client have the same issue?

bear101:
Ah, just read the other PR

PR #855, opened 2021-05-02
QT - Fixed issues caused by #848 and 6848ebd

Oreonan:


Oreonan:
We need a method to clear m_last_channel. Currently we have an issue, reproduce like this:
- Join another channel that the root
- Leave this channel
- Now if connection is droped, when it's retrieved, client join channel previously joined, but we have leave it in the meantime
If a method exists to clear m_last_channel, I not found this.

Oreonan:
@bear101 I continued tests with this changes. For your question concerning temporary channels, my changes haven't any effect to this. I tested with 5.7.1.5031, if we create a temporary channel, and connection is droped, temporary channel isn't recreate when reconnecting.
In this PR, I improved usage of m_host.channel to use it only when it's necessary (for root channel) and continue to use m_last_channel like previously for other channels. This PR also fix the issue concerning incorrect join if we are leave a channel before dropped connection.

PR #856, opened 2021-05-02
Fix TTS bitmasks

Oreonan:


bear101:
Good you found the right conversion. Did you try building using Visual Studio?

Oreonan:
Euh, I don't really found good conversion :), I have an issue with 
latest 2 entries of tts events list in preferences, we can enable this 
by "enable all" button, but after it's impossible to change state 
anymore. And no, I don't try to build with Visual Studio.

Oreonan:
Ah, it's a compilation issue due to my system. I build with mingw32 X86. 
There is no problem when build to X64.

PR #880, opened 2021-05-13
QT - Accessible name on away spinbox

Oreonan:


qcfy1204:
How can I call the SML configuration file generated by the previous version? The current version is 5.8 QT

PR #881, opened 2021-05-13
QT - Return state of a channel using an accessible text

Oreonan:


Client/qtTeamTalk/channelstree.cpp
```diff
@@ -880,6 +880,7 @@ void ChannelsTree::slotUpdateTreeWidgetItem(QTreeWidgetItem* item)
             if (!item->data(COLUMN_MEDIAFILE, Qt::CheckStateRole).isNull())
                 item->setData(COLUMN_MEDIAFILE, Qt::CheckStateRole, QVariant());
         }
+        item->setData(COLUMN_ITEM, Qt::AccessibleTextRole, QString("%1: %2").arg(channame).arg((item->isExpanded()? tr("Expanded"):tr("Collapsed"))));
```
bear101, line 883: Can we use Qt::AccessibleTextRole on classroom checkboxes for media stream/voice/webcam? You said these options were "invisible" to the screen-reader, so maybe this way we can announce the current state

Oreonan: I don't reminder, where this checkboxes are?

PR #893, opened 2021-05-15
Windows - Propose to enable TTS if screenreader is detected at first QT startup

Oreonan:
- To make transition bitween classic and QT client more agreable :)

### DIFF
--- a/Client/qtTeamTalk/mainwindow.cpp
+++ b/Client/qtTeamTalk/mainwindow.cpp
@@ -701,7 +701,6 @@ void MainWindow::loadSettings()
 
         if(answer.clickedButton() == YesButton)
         {
-            ttSettings->setValue(SETTINGS_TTS_ACTIVEEVENTS, TTSEvents(TTS_USER_LOGGEDIN | TTS_USER_LOGGEDOUT | TTS_USER_JOINED_SAME | TTS_USER_LEFT_SAME | TTS_USER_TEXTMSG_PRIVATE | TTS_USER_TEXTMSG_CHANNEL | TTS_USER_TEXTMSG_BROADCAST | TTS_FILE_ADD | TTS_FILE_REMOVE | TTS_MENU_ACTIONS));
             ttSettings->setValue(SETTINGS_TTS_ENGINE, 2);
             ttSettings->setValue(SETTINGS_GENERAL_FIRSTSTART, 0);
         }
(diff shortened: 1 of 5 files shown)


Client/qtTeamTalk/mainwindow.cpp
```diff
@@ -682,6 +682,30 @@ void MainWindow::loadSettings()
...
+        QAbstractButton *YesButton = answer.addButton(tr("&Yes"), QMessageBox::YesRole);
+        QAbstractButton *NoButton = answer.addButton(tr("&No"), QMessageBox::NoRole);
+        Q_UNUSED(NoButton);
+        answer.setIcon(QMessageBox::Question);
+        answer.setWindowTitle(APPNAME_SHORT);
+        answer.exec();
+
+        if(answer.clickedButton() == YesButton)
+        {
+            ttSettings->setValue(SETTINGS_TTS_ENGINE, 2);
```
bear101, line 704: Don't use constants for values that have an corresponding enumeration.

I've corrected this.

Client/qtTeamTalk/mainwindow.cpp
```diff
@@ -682,6 +682,35 @@ void MainWindow::loadSettings()
...
+
+#if defined(Q_OS_WINDOWS)
+    bool tolkLoaded = Tolk_IsLoaded();
+    if (!tolkLoaded)
+        Tolk_Load();
+
+    bool SRActive = Tolk_DetectScreenReader() != nullptr;
+
+    if (!tolkLoaded)
+        Tolk_Unload();
```
bear101, line 694: I've made this correction. Previously Tolk would be unloaded since it's started by startTTS() a few lines up.

Client/qtTeamTalk/mainwindow.cpp
```diff
@@ -682,6 +682,35 @@ void MainWindow::loadSettings()
...
+
+    if (ttSettings->value(SETTINGS_GENERAL_FIRSTSTART, SETTINGS_GENERAL_FIRSTSTART_DEFAULT).toBool() && SRActive)
```
bear101, line 696: I've made a modification so the message popup only comes with a screen-reader is detected

Oreonan:
Ah yes, it's look better like this.

However, we probably have to change Setup/Client/Windows/TeamTalk5.ini 
to set first-start as true instead of 1

bear101:
You don't need to set default values in TeamTalk5.ini. The default values should be coded in the Qt client.
"first-start" is automatically set to SETTINGS_GENERAL_FIRSTSTART_DEFAULT

PR #905, opened 2021-05-22
QT - Add missing sound events - Fixed #786

### LINKED ISSUE #786: Missing default sound events
Qt Client missing sound events:
Sounds\txqueue_start.wav
Sounds\txqueue_stop.wav
Sounds\vox_disable.wav
Sounds\vox_enable.wav
Sounds\vox_me_disable.wav
Sounds\vox_me_enable.wav

Basically the following from Classic client:
[SOUNDEVENT_TRANSMITQUEUE_HEAD](https://github.com/BearWare/TeamTalk5/blob/a92ab7b6609b140f48c98ec45b225f702eb68bf9/Client/TeamTalkClassic/TeamTalkBase.h#L345)
SOUNDEVENT_TRANSMITQUEUE_STOP
SOUNDEVENT_VOICEACTIVATED
SOUNDEVENT_VOICEDEACTIVATED
SOUNDEVENT_ENABLE_VOICEACTIVATION
SOUNDEVENT_DISABLE_VOICEACTIVATION

Oreonan:


bear101:
@Oreonan Have you found a way to easily move sound events from one column to another? The left column is now 10 row and the right column is 16 rows, so it looks a bit weird.

Oreonan:
Yes, I can do this easily I think.What sounds should I move to left column?

bear101:
Just take the first three in the right column. Then there's 13 rows in
each column.

What tool do you use to do this?

--
Kind regards

Bjørn Damstedt Rasmussen
www.bearware.dk

Oreonan:
@bear101 it's better like this in latest commit?

Oh, to do this I simply copied and passed all lines in .ui file and 
manually replaced lines number. It's the more "easy" way I found :)

PR #910, opened 2021-05-24
QT - Defaults value for sound events

Oreonan:


Oreonan:
@bear101 Following this changes, should I remove default values from .ini files provide by setup program?

bear101:
Good idea! I had the same problem when I wanted to test the new sound events :)

PR #916, opened 2021-05-30
Show correct server after addition or deletion, Fixed #915

### LINKED ISSUE #915: First server in serverlistdlg.cpp is always shown
After commit 9e2f114d6ff274bb7d7a6068660a1ca925768aef it is no longer the "latest server" that is shown. Instead it's always server index 0 in server listbox.

Oreonan:


bear101:
Thanks!

PR #918, opened 2021-06-01
QT - Fix streaming of files path containing quote

Oreonan:
Very useful on Windows to use option "Copy as file path" of files explorer and past this to CTRL-S dialog.


bear101:
Maybe there should be a "#if defined(Q_OS_WINDOWS)" since quotes are ok on Linux and macOS.

PR #921, opened 2021-06-03
QT - Fix navigation using right key in channels tree

Oreonan:


Oreonan:
Apparently this PR does not compile on Linux. This is strange because the code is pretty much the same as in PR # 919 which compiles perfectly. I have no problem with Windows.

jeremyp3:
the error is:
mainwindow.cpp:6206:65: error: expected unqualified-id before numeric
constant
     if (object == ui.channelsWidget && event->type() == QEvent::KeyPress)
                                                                 ^~~~~~~~
mainwindow.cpp:6206:63: error: expected ‘)’ before numeric constant
     if (object == ui.channelsWidget && event->type() == QEvent::KeyPress)
        ~                                                      ^

Client/qtTeamTalk/mainwindow.cpp
```diff
@@ -6198,3 +6199,20 @@ void MainWindow::slotEnableVoiceActivation(bool checked)
 {
     slotMeEnableVoiceActivation(checked, SOUNDEVENT_VOICEACTMEON, SOUNDEVENT_VOICEACTMEOFF);
 }
+
+bool MainWindow::eventFilter(QObject *object, QEvent *event)
```
bear101, line 6203: Can't the event filter be implemented in ChannelsTree-class? It would be more correct to place it there.

Oreonan: I don't succeed to place it in ChannelsTree class, probably because there is no ui element to place it. My tests has finished by a crash when pressing any key on channels tree :(

bear101:
@jeremyp3 We don't see this compilation issue: https://github.com/BearWare/TeamTalk5/actions

Oreonan:
I think this issue is due to conflicts bitween X11 headers and my changes.
@bear101 you have probably right when you say that it's better to place this code in channelstree class, do this probably solve issue described by @jeremyp3 but for now I don't succeed with this.

jeremyp3:
I confirm, the compilation is successful after this modification. thanks
@oreonan

PR #923, opened 2021-06-03
TT_EnableAudioBlockEvent outside channel for TT_MUXED_USERID

bear101:
When not in a channel calling `TT_EnableAudioBlockEventEx(ttInst, TT_MUXED_USERID, STREAMTYPE_VOICE, fmt, TRUE)` will now cause TTInstance to output AudioBlock's containing silence. The AudioBlocks will use a frame size of 20 msec when 'fmt' is specified.

### DIFF
--- a/Library/TeamTalkLib/teamtalk/client/ClientNode.cpp
+++ b/Library/TeamTalkLib/teamtalk/client/ClientNode.cpp
@@ -3010,15 +3010,30 @@ bool ClientNode::EnableAudioBlockCallback(int userid, StreamTypes sts,
     {
         if (enable)
         {
-            m_audiomuxer_stream.reset(new AudioMuxer(sts));
-            // start muxer if already in channel
-            if (m_mychannel)
+            audiomuxer_t newmuxer(new AudioMuxer(sts));
+            media::AudioInputFormat infmt;
+
+            if (outfmt.IsValid())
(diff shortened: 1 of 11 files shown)


Library/TeamTalkLib/teamtalk/client/ClientNode.cpp
```diff
@@ -3010,14 +3010,30 @@ bool ClientNode::EnableAudioBlockCallback(int userid, StreamTypes sts,
...
+            media::AudioInputFormat infmt;
+
+            if (outfmt.IsValid())
+                infmt = media::AudioInputFormat(outfmt, PCM16_DURATION_SAMPLES(20, outfmt.samplerate));
+            else if (m_mychannel)
             {
                 auto codec = m_mychannel->GetAudioCodec();
-                bool startmux = m_audiomuxer_stream->RegisterMuxCallback(codec, std::bind(&ClientNode::AudioMuxCallback, this, _1, _2));
+                infmt = GetAudioCodecAudioInputFormat(codec);
+            }
```
bear101, line 3022: If not output format is specified, i.e. using `TT_EnableAudioBlockEvent` instead of `TT_EnableAudioBlockEventEx` while not being in a channel will return "false" with this modification. Should we just default to 48000Hz/Stereo?

Lauriumora: I'd be OK with 48K/stereo, as long as it's a documented behavior.

Lauriumora (approved):

PR #926, opened 2021-06-06
QT - Make classroom toggles accessible in channels tree

Oreonan:


Client/qtTeamTalk/channelstree.cpp
```diff
@@ -1489,8 +1494,19 @@ bool ChannelsTree::eventFilter(QObject *object, QEvent *event)
                 {
                     item->setExpanded(true);
                 }
-                return true;
+                else if ((item->type() & USER_TYPE))
+                {
+                    QTreeWidgetItem* parent = item->parent();
+                    int parentid = (parent->data(COLUMN_ITEM, Qt::UserRole).toInt() & ID_MASK);
+                    UserRights rights = TT_GetMyUserRights(ttInst);
+                    if (TT_IsChannelOperator(ttInst, TT_GetMyUserID(ttInst), parentid) ||
```
bear101, line 1502: Shouldn't non-operator/admin also be allowed to see if they can transmit?

Oreonan: Hmmm, maybe. But text of accessible text shouldn't be different in this case?

bear101: Doesn't have to be. Does it?

bear101: hm, maybe it should be "User can transmit..."

Oreonan: I see 2 solutions:
- We change text for all case to "User can transmit: yes/no"
- We don't change current text for operators/admins but we change for users by "transmission: allowed/disallowed"

bear101:
I've removed the event filter and instead also put accessible text on channel column. Can you manage to use it the way it is now?

Oreonan:
Yes it's OK for me.

PR #942, opened 2021-06-25
QT - Ability to choose and add sounds packs

Oreonan:


Oreonan:
We maybe have to make a function to reset sounds-pack to default value, to avoid repeat multiple times same lines, like in 906848a commit

Client/qtTeamTalk/mainwindow.cpp
```diff
@@ -714,6 +714,29 @@ void MainWindow::loadSettings()
             ttSettings->setValue(SETTINGS_GENERAL_FIRSTSTART, false);
     }
 #endif
+// Sounds pack checks
+    QString packset = ttSettings->value(SETTINGS_SOUNDS_PACK).toString();
+    QString packname = QString("%1/%2").arg(SOUNDSPATH).arg(packset);
+    QDir packdir(packname);
+    if((!packdir.exists()) && packset != "Default")
```
bear101, line 721: Shouldn't this be 'packset != tr("Default")' ?

Client/qtTeamTalk/preferencesdlg.cpp
```diff
@@ -1,1877 +1,1980 @@
...
+    break;
+    case SOUND_TAB :  //sound system
+        initDevices();
+        ui.mediavsvoiceSlider->setValue(ttSettings->value(SETTINGS_SOUND_MEDIASTREAM_VOLUME,
+                                        SETTINGS_SOUND_MEDIASTREAM_VOLUME_DEFAULT).toInt());
+        break;
+    case SOUNDEVENTS_TAB :  //sound events
+    {
+        ui.spackBox->clear();
+        ui.spackBox->addItem("Default");
```
bear101, line 568: Also tr("Default") here.

Oreonan: I prefer let this not translated because it's a pack name, all users should have same default pack name I think.

Oreonan:
I finally translated the "Default" sounds pack in commit 3c65a29 after another request from @Nardol

bear101:
Line endings were screwed up in preferencesdlg.cpp so I had to do a squash commit.

PR #958, opened 2021-07-10
Android - Separate sounds and TTS for channel messages sent and received

Oreonan:


bear101:
Do you have a sound we can use as "channel message sent"? It's a bit confusing that sent and receive is the same.

If so please add it to Setup/Client/Sounds then I can convert from .wav to .ogg. Then we'll also integrate the sound in other clients.

Oreonan:
I added, I simply used the same for receive and sent reverting the notes 
of the sound:) . In another hand if you also want a new sound for PM 
sent, it will be more complicated to find without make a totally 
different sound...

PR #963, opened 2021-07-13
QT - Fix sameHost and sameHostEntry when checking for nickname changes

Oreonan:
After commit 27f2a75 of PR #940, change nickname when connecting only update latesthost, change wasn't save to serverentry.
This issue is due to sameHost and sameHostEntry which always check nickname corresponding. This PR add a parameter to this functions to make nickname check optional.
When call dialog to change nickname, sameHost and sameHostEntry are now call without checking nickname, and issue described above is fix.

### DIFF
--- a/Client/qtTeamTalk/common.cpp
+++ b/Client/qtTeamTalk/common.cpp
@@ -1105,36 +1105,21 @@ void addTextToSpeechMessage(TextToSpeechEvent event, const QString& msg)
 
 bool HostEntry::sameHost(const HostEntry& host, bool nickcheck) const
 {
-    if(nickcheck == true)
-    {
-        return ipaddr == host.ipaddr &&
-               tcpport == host.tcpport &&
-               udpport == host.udpport &&
-               encrypted == host.encrypted &&
-               /* srvpasswd == host.srvpasswd && */ //don't include passwords
-               username == host.username &&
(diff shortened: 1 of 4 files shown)


Client/qtTeamTalk/common.cpp
```diff
@@ -1105,36 +1105,21 @@ void addTextToSpeechMessage(TextToSpeechEvent event, const QString& msg)
...
+           /* srvpasswd == host.srvpasswd && */ //don't include passwords
+           username == host.username &&
+           /* password == host.password && */
+           (!nickcheck || nickname == host.nickname) &&
+           channel == host.channel
+           /* && hosts[i].chanpasswd == host.chanpasswd*/;
 }
 
-bool HostEntry::sameHostEntry(const HostEntry& host, bool nickcheck) const
+bool HostEntry::sameHostEntry(const HostEntry& host) const
```
bear101, line 1120: @Oreonan 'nickcheck' was not used in "sameHostEntry" so I removed it. In mainwindow.cpp you passed 'false' but it had no effect

Oreonan: I just removed nickname check when serverlistdlg.cpp select a server. Now it's working as expected for me. It's OK for you?

Issue #278, opened 2017-02-12
Add Push-to-Talk lock


bear101:

tech10:
This seems like it could be a useful option, but different sounds when push to talk is enabled or disabled would enhance the usefulness of the lock. Alternatively, there could be a sound when the client is transmitting, and another sound when the client is not transmitting, but had previously been.

bear101:
User can now lock the "push-to-talk" key for extended talk when channel has "voice activation" with the ctrl-shift-a key combination, and unlock it with the same key combo, but we need a sound for each press of this key combo to confirm the "lock" changes to visual impaired users.

PR #968, opened 2021-07-14
Push To Talk lock. Issue #278

bear101:
Can you check that "push-to-talk lock" is always disabled on Linux and follows the "normal" push-to-talk?

### DIFF
--- a/ChangeLog.txt
+++ b/ChangeLog.txt
@@ -23,6 +23,7 @@ iOS Client
 -
 Server
 - Support for SIGHUP to reload server configuration
+- Error reported if unable to save configuration file (tt5srv.xml)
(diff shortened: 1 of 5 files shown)


Oreonan:
Yes, on Linux behavior is still the same as before this new option. And new checkbox is disabled.

bear101:
Should there be a separate sound for "hotkey released"?

Oreonan:
Yes, I think, if you have an idea of what sound to use. Alternatively, use same sounds as voice activation for PTT lock, but it probably a bit confused.

bear101:
I'll merge it in. The new sound event will be a separate issue.

PR #981, opened 2021-07-27
Updated the Brazilian Portuguese translations for Android and TeamTalk Classic

DalisonCPU:


Client/TeamTalkClassic/lng_classic/Portuguese_BR.lng
```diff
@@ -92,33 +92,33 @@
     <item id="191">Transmissão de vídeo permitida</item>
     <item id="192">Transmissão de tela permitida</item>
     <item id="193">Permitida a transmissão de arquivos de mídia</item>
-    <item id="194">%s mudou a subscrição de mensagens do usuário para %s.</item>
```
Oreonan, line 95: Since 5.7.1 you now should to use %s instead of %d because "0" and "1" has been replaced by "enable" and "disable".

DalisonCPU:
You can't change? in almost every one I remembered I changed it to %d before

Oreonan:
You've also deleted all submodules in your PR, merge this would cause a lot of problems. I've manually update translations with your changes.

PR #996, opened 2021-08-14
Fix behavior of enter key on mediafileComboBox

Oreonan:


Client/qtTeamTalk/streammediafiledlg.cpp
```diff
@@ -492,6 +492,11 @@ bool StreamMediaFileDlg::eventFilter(QObject *object, QEvent *event)
             ui.mediafileComboBox->lineEdit()->paste();
             return true;
         }
+        else if (keyEvent->key() == Qt::Key_Enter || keyEvent->key() == Qt::Key_Return)
```
bear101, line 495: Also convert to keyPressEvent()

Oreonan: Not succeeded for now also here.

PR #997, opened 2021-08-14
Modify some keys behavior

Oreonan:


Oreonan:
This is just a workarround, also reported to Qt because temporary solution not working in main window on X11 due to incompatibilities with X11 libraries.
https://bugreports.qt.io/browse/QTBUG-95805

jeremyp3:
i think excluding linux is not an acceptable workaround. I allow myself to post this comment so as not to forget it

Oreonan:
A possible solution find with @jeremyp3 to work with X11 libraries is to move the eventfilter function just after the main window constructor and include the X11 libraries after that.
I don't really like this solution because is don't respect anymore the normal structuration of a cpp file, but it works.
@bear101 should I apply this change?

Client/qtTeamTalk/mainwindow.cpp
```diff
@@ -6417,3 +6424,42 @@ void MainWindow::slotEnableVoiceActivation(bool checked)
 {
     slotMeEnableVoiceActivation(checked, SOUNDEVENT_VOICEACTMEON, SOUNDEVENT_VOICEACTMEOFF);
 }
+
+#ifndef Q_OS_LINUX
+bool MainWindow::eventFilter(QObject *object, QEvent *event)
```
bear101, line 6429: Isn't it simpler to override QWidget::keyPressEvent() like we do in TTSEventsTreeView-class. It seems excessive to override eventFilter() since this handles _all_ events. Also non-key press.

Oreonan: Hmmmm, with overriding keyPressEvent, we can't apparently not check what object receive the key, or I don't know do this... So it's maybe not applicable here.

Oreonan:
Ah, no, can't use enter or escape keys anymore to close preferences or user accounts dialog after commit e05f207

PR #999, opened 2021-08-16
Fix navigation using left key on columns of channels tree

Oreonan:


Client/qtTeamTalk/channelstree.cpp
```diff
@@ -1498,3 +1499,22 @@ void ChannelsTree::slotUserVideoFrame(int userid, int stream_id)
     Q_UNUSED(userid);
     Q_UNUSED(stream_id);
 }
+
+bool ChannelsTree::eventFilter(QObject *object, QEvent *event)
```
bear101, line 1503: Also here use keyPressEvent() instead of eventFilter(). Installing an event-filter will slow down application performance, since all application events will go through the installed filters.

Oreonan:
OK :), I'll update all PR

bear101:
Should we merge these before I sent out a beta2? Or wait for v5.8.2?

Oreonan:
I think it's better to include this in a beta 2

Client/qtTeamTalk/channelstree.cpp
Oreonan: Not succeeded for now...

PR #1000, opened 2021-08-16
Qt - Implement CTRL+Shift+G to speak channel state

Oreonan:
Without change log entry for now because we probably have to wait until the next version due to beta currently public.

### DIFF
--- a/Client/qtTeamTalk/mainwindow.cpp
+++ b/Client/qtTeamTalk/mainwindow.cpp
@@ -435,7 +435,7 @@ MainWindow::MainWindow(const QString& cfgfile)
     connect(ui.actionSpeakChannelInfo, &QAction::triggered,
             this, &MainWindow::slotChannelsSpeakChannelInformationGrid);
     connect(ui.actionSpeakChannelStat, &QAction::triggered,
-            this, &MainWindow::slotChannelsSpeakChannelStatisticsGrid);
+            this, &MainWindow::slotChannelsSpeakChannelStatusGrid);
     connect(ui.actionBannedUsersInChannel, &QAction::triggered,
             this, &MainWindow::slotChannelsListBans);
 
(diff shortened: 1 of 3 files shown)


Client/qtTeamTalk/mainwindow.cpp
```diff
@@ -4549,6 +4551,86 @@ void MainWindow::slotChannelsSpeakChannelInformationGrid(bool /*checked =false *
...
 
+void MainWindow::slotChannelsSpeakChannelStatisticsGrid(bool /*checked =false */)
+{
+    slotChannelsSpeakChannelStatistics();
+}
+
+void MainWindow::slotChannelsSpeakChannelStatistics()
+{
+    QString speakList, voice, mediafile, video, desktop;
+    QVector<int> users = ui.channelsWidget->getUsersInChannel(TT_GetMyChannelID(ttInst));
```
bear101, line 4562: Shouldn't this be currently selected channel in channel-tree? Or should it only apply to own channel?

Client/qtTeamTalk/mainwindow.cpp
```diff
@@ -4549,6 +4551,86 @@ void MainWindow::slotChannelsSpeakChannelInformationGrid(bool /*checked =false *
...
+    QVector<QString> voice1, mediafile1, video1, desktop1;
+    for (int i=0;i<users.size();i++)
+    {
+        User user = {};
+        TT_GetUser(ttInst, users[i], &user);
```
bear101, line 4567: Use ui.channelsWidget->getUser() instead of TT_GetUser(). TT_GetUser() runs in a different thread, so it may yield different results.

Client/qtTeamTalk/mainwindow.cpp
Oreonan: Use ui.channelsWidget->selectedChannel always return no active users if we aren't in the specified channel, so we can probably let this apply only to the channel we are currently in

PR #1002, opened 2021-08-23
Left key in channels-tree

bear101:


bear101:
@Oreonan Please check if this works the same as with event-filter.

Oreonan (approved):

Oreonan:
Yes, works as expected. Thanks.

Issue #1010, opened 2021-09-08
Subclass QComboBox to handle double Enter-press


bear101:
Related to #996 

Locate QComboBox and uses (serverlistdlg.cpp, streammediadlg.cpp, etc) where QComboBox is used and replace by subclassed QComboBox which handles the double Enter-issue.

PR #1018, opened 2021-09-13
Process Enter-key when QComboBox has focus

bear101:
@Oreonan Does this fix the Enter key issue in ServerListDlg and StreamMediaFileDlg? This is related to #1010

If it does then I can also change the other QComboBox classes


Oreonan (approved):
Tested on StreamMedia and ServerList, it works as expected.

PR #1027, opened 2021-09-19
Add Behavior tab to Preferences

Oreonan:
For now with 3 options
- Option to open a dialog box when excluded from a channel or server
- Option to close transfer dialog when a file transfer is completed (now permanent)
- Option to choose the behavior of double click (and enter key) on a channel  (approval required in particular for this option)


bear101:
Can't we stick to only having Display-tab? It seems these new options are already related to what is in the Display-tab.

Oreonan:
Yes, we can. But I'm not know the space available in display tab. I can 
move this options in display tab...

Client/qtTeamTalk/common.h
```diff
@@ -66,7 +66,7 @@
 #endif
 
 #if defined(Q_OS_WINCE)
-#define _W(qstr) qstr.utf16()
+#define _W(qstr) qstr.utf8()
```
bear101, line 69: How come this change? Windows CE is no longer supported

Oreonan: Hmmmmm, a mistake that would not have been there

PR #1030, opened 2021-09-22
Implement reload action for init and systemd

Nardol:
* Add reload and force-reload parameters to init scripts
* Add ExecReload to systemd unit files


bear101:
Thanks for this addition to the service scripts. I've also created this issue #1031 which will fix that log file is not restarted

Issue #1032, opened 2021-09-27
Sort by popularity in Qt client


bear101:
Show channels with most users first. As alternative to Ascending.

Oreonan:
Fixed by #1035

PR #1035, opened 2021-09-29
Sort by popularity option in Display-tab

bear101:
Fixes issue #1032


Oreonan (approved):
I've linked the label with the checkbox with a buddy property.

PR #1041, opened 2021-10-01
update chinese translation

hejiadong0608:


bear101:
@hejiadong0608 Should we put your name in the translator list in the About Dialog? If so, what is the name we should list?

hejiadong0608:
@bear101  Yes, i'd love to. My name is 贺稼栋.

PR #1044, opened 2021-10-03
Qt keep disconnected

bear101:
For some reason I cannot push to your repository any more. But here's the fix for keep-disconnected.


Oreonan (approved):
Thanks for your help

PR #1049, opened 2021-10-08
Mac OS - Option to speak selected item in lists

CoBC:
Only implemented in channels tree for now, will can be add to other lists more later after users comments


bear101:
@Oreonan I'm making a few changes on this branch, so hope you're not editing as well

PR #1070, opened 2021-10-25
Build universal binary for macOS (arm64 and x86_64)

bear101:
Build TeamTalk as universal binary for Darwin (macOS). Still work in progress


bear101:
@CoBC This change requires Xcode 13 in order to compile. This unfortunately means that you cannot build TeamTalk core (BUILD_TEAMTALK_CORE) on your MacBook Pro. You can, however, still build the Qt client if you download the TeamTalk SDK or build artefacts from GitHub actions.

PR #1082, opened 2021-10-30
Restore window size at startup. Fixed #506

### LINKED ISSUE #506: Let the TeamTalk main window start maximized
On Debian 10, using Compiz window manager, I set this window manager to maximize windows when they open.
But when starting TeamTalk, it looks like this is not the case, I have to maximize the window manually.

Could it be possible to implement one or some these suggestion?
- let the system automatically maximize the TeamTalk main window on start;
- Remember the last window size when closing or when it is changed;
- Add an option to automatically start TeamTalk maximized?

On Windows, it looks like working when the shortcut to TeamTalk is set to start TeamTalk maximized. I tried with others QT application as VLC and Twinkle on Linux but they start maximized as expected.

Oreonan:


Client/qtTeamTalk/mainwindow.cpp
```diff
@@ -578,6 +578,8 @@ MainWindow::~MainWindow()
 
     ttSettings->setValue(SETTINGS_DISPLAY_FILESHEADER, ui.filesView->header()->saveState());
 
+    ttSettings->setValue(SETTINGS_DISPLAY_WINDOW_MAXIMIZE, QWidget::isMaximized());
```
bear101, line 581: It should be "this->isMaximized()" in case there were more sub-classes

Client/qtTeamTalk/mainwindow.cpp
```diff
@@ -709,7 +711,9 @@ void MainWindow::loadSettings()
     }
     //set files header to last position
     ui.filesView->header()->restoreState(ttSettings->value(SETTINGS_DISPLAY_FILESHEADER).toByteArray());
-
+    // Maximize window if necessary
+    if (ttSettings->value(SETTINGS_DISPLAY_WINDOW_MAXIMIZE).toBool() == true)
+        QWidget::showMaximized();
```
bear101, line 716: Should be "this->showMaximized()"

PR #1118, opened 2021-11-05
Text-to-speech event when remote user is typing in private chat

Oreonan:
Should we add a new configurable event for this specific one in TTS preferences?
Also, should we check if focus is in private chat window to send this event.


ChrisDuffley:
I see no reason not to. Maybe make a new variable of the person who's typing in that case and add "is typing" for this to correctly announce things.

Christopher Duffley
Cell: +1 (603) 851-6971

Sent from my iPhone 12 Pro Max

On Nov 5, 2021, at 09:56, Oreonan ***@***.***> wrote:

﻿
Should we add a new configurable event for this specific one in TTS preferences?

You can view, comment on, or merge this pull request online at:

  https://github.com/BearWare/TeamTalk5/pull/1118

Commit Summary

Text-to-speech event when remote user is typing in private chat
File Changes  (2 files)
M ChangeLog.txt (1)
M Client/qtTeamTalk/textmessagedlg.cpp (1)
Patch Links:

https://github.com/BearWare/TeamTalk5/pull/1118.patch
https://github.com/BearWare/TeamTalk5/pull/1118.diff

bear101:
Do we need an enable/disable option in Preferences->Text-to-Speech for this?

Oreonan:
I thought to some possibilities:
* Add a new configurable event for this one
* Add an option to let user determine if this event is send when user is in private chat window or in all windows.
I don't know if we should have this two options, just one of them, or let this like currently, so your opinion is welcome :)
Also I don't know what can be the best behavior for this event by default, maybe just send it if private chat window has focus?

ChrisDuffley:
Personally, the more control over this, the better. Both options and focus option would help.

Christopher Duffley
Cell: +1 (603) 851-6971

Sent from my iPhone 12 Pro Max

On Nov 6, 2021, at 10:39, Oreonan ***@***.***> wrote:

﻿
I thought to some possibilities:

Add a new configurable event for this one
Add an option to let user determine if this event is send when user is in private chat window or in all windows.
I don't know if we should have this two options, just one of them, or let this like currently, so your opinion is welcome :)
Also I don't know what can be the best behavior for this event by default, maybe just send it if private chat window has focus?

Oreonan:
@bear101 If you have better texts for checkboxes they are welcome :)

PR #1134, opened 2021-11-14
Number of characters as accessible hint for public and private chat

Oreonan:


beqabeqa473:
Hello.

I think this pr should be reconsidered or reverted.

First of all, as i understand correctly, 512 is not character count but length of bytes transmitted through server.
in this case, characters could be 1 to 4 bytes? so if we send emojies, some of them which are 4 bytes, we can send only 128 characters.
Secondly, how it is done now is very annoying and spammy.
I suggest to remove this announcement from there and move to message title and indicate with sound if we exceed 512 bytes limit.

PR #1135, opened 2021-11-14
Android - Accessible hint for gender

Oreonan:


Oreonan:
Bad manipulation... No reason to close this, just my mistake.

Issue #412, opened 2019-03-24
Inform users when text message limit is exceeded


bear101:
UTF-16 strings are converted to UTF-8 and cut at 511 characters. If the string is cut then the user should be informed or indicated in the received message.

bear101:
https://stackoverflow.com/questions/35328529/stdstring-optimal-way-to-truncate-utf-8-at-safe-place

Oreonan:
On Windows when limit is exceeded, an empty line is send. We can probably send the 511 first characters.

bear101:
The problem is that the string you send on Windows is encoded in 16-bit values which are then converted to UTF-8 which can be up to 32-bit values.

bear101:
https://stackoverflow.com/questions/37623359/how-to-remove-the-last-character-of-a-utf-8-string-in-c

PR #1147, opened 2021-11-27
Change way to display number of characters for messages

Oreonan:
@bear101 If you are agree with this approach, I can also modify in main window. The GUI has probably to be reviewed also.
@beqabeqa473 Is this approach better for you?

### DIFF
--- a/ChangeLog.txt
+++ b/ChangeLog.txt
@@ -15,7 +15,7 @@ Default Qt Client
 - Status message is triggered if unable to transmit
 - Fixed "Cancel" button not working in "New client instance" dialog
 - Navigating with arrow keys now include status in TTS message on macOS
-- Number of characters in edit field of public and private chat is now indicate as accessible hint
+- Alert when characters limit exceeded for channel and private chat
 - Client-menu now has Sound Configuration submenu to quickly change sound devices
 Android Client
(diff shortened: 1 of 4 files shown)


bear101:
@Oreonan Instead of displaying the number of characters, can't we show a dialog if the number is exceeded? There's not really room for this extra label on the mainwindow

bear101:
And display exceeded by how much

Oreonan:
Did you mean add a dialog if limit is exceeded hen sending the message? 
And in case the limit is exceeded I guess we should keep the text in new 
message field?

bear101:
Yes, keep the text and not send until the user has reduced it

bear101:
Unfortunately I never managed to fix #412 so messages can be trimmed correctly. I wonder if new versions of C++ library has a way to trim UTF-8 strings.

beqabeqa473:
hi all.

It is better now, but i think displaying exceeded limit by number of bytes is not accurate, because user may don't know that emoji is 4 bytes size, if exceeded limit is 520 characters, and they are removing last char, which is sized by 4 bytes, an app wiill tell them that exceeded limit now is 516 characters which is wrong from ux point of view.

What do you think about that?

Issue #1020, opened 2021-09-14
Hidden channel cannot be auto join depending on user rights


Oreonan:
If an user hasn't right to see hidden channel, and his account is configured on server side to join an hidden channel at login, this channel is not joined, and the user is just keeping out of any channel.

PR #1156, opened 2021-12-04
Favor UserAccount.szInitChannel over user's stored join setting

bear101:
This fixes issue described in #1020.

This can easily break something but it simplifies the join process. Basically szInitChannel should take priority over user's stored setting and is otherwise treated the same with regard to channel creation. I.e. if parent existing then create a new channel. Previous szInitChannel only worked if the channel already existed.


Oreonan:
For me #1020 isn't fixed by this PR. If I edit an account to join an hidden channel, and I connect with this account (user account without right to sse hidden channel), no channel is joined.

bear101:
Ah, I didn't have a '/' in the end when I tested yesterday. This cause subchannels with no name. I've fix this now.

Oreonan:
OK, this point it's OK for me now. I'll test with temporary channels tomorrow and make a review after that.

Oreonan (approved):
Seems to be working as expected in all situations now. I guess change log need to be updated, and all will be OK.

PR #1157, opened 2021-12-04
Templates for creating issues

bear101:
I've set up two templates that people will have to fill out when creating an issue. One for bug reports and one for feature requests.

Many issues lack information, so hopefully this will make people more through before submitting issues. Also people should use Discussions more before creating an issue. Often the feature they request can be obtained in some other way using existing functionality.


Oreonan (approved):
Look good for me.

PR #1166, opened 2021-12-12
Fixed "Disconnected from server" TTS event emitted repeatedly when connection to server is lost

CoBC:


Client/qtTeamTalk/mainwindow.cpp
```diff
@@ -1913,7 +1913,8 @@ void MainWindow::Disconnect()
 
     addStatusMsg(STATUSBAR_BYPASS, tr("Logged out from %1, TCP port %2, UDP port %3").arg(m_host.ipaddr).arg(m_host.tcpport).arg(m_host.udpport));
     updateWindowTitle();
-    addTextToSpeechMessage(TTS_SERVER_CONNECTIVITY, tr("Disconnected from server"));
+    if (QObject::sender() == ui.actionConnect)
```
bear101, line 1916: I think it's more correct to use "if (timerExists(TIMER_RECONNECT))"

PR #1172, opened 2021-12-16
Do not set default nickname to noname in preferences to use noname#userID by default

### LINKED ISSUE #1171: Set empty nick name by default
### Description

For now, default configuration set nick name to No name.
So if someone connect without having set a nick name, "no name" will be displayed.
Could it be possible to set this to empty to let the server listing the user as "no name#id"?
In the case more than one user would connect without setting a nick name, they would have different name.

### Application

- [X] qtTeamTalk
- [X] TeamTalkAndroid
- [X] iTeamTalk
- [ ] TeamTalkServer

Nardol:
When running the TeamTalk client for the first time, default preferences set nickname to noname so when connecting to a server, user is called noname. This PR allows user to be named noname#userID to avoid multiple users with the same nickname if more than one user have not set their nickname.

Fixes #1171


Client/TeamTalkAndroid/src/main/java/dk/bearware/backend/TeamTalkService.java
```diff
@@ -622,8 +622,7 @@ private void login() {
 
         String nickname = ttserver.nickname;
         if (TextUtils.isEmpty(nickname)) {
-            String def_nick = getResources().getString(R.string.pref_default_nickname);
-            nickname = PreferenceManager.getDefaultSharedPreferences(getApplicationContext()).getString(Preferences.PREF_GENERAL_NICKNAME, def_nick);
+            nickname = PreferenceManager.getDefaultSharedPreferences(getApplicationContext()).getString(Preferences.PREF_GENERAL_NICKNAME);
```
bear101, line 625: You should change R.string.pref_default_nickname to blank instead of removing the code that allows a default value.

Client/TeamTalkClassic/TeamTalkDlg.cpp
```diff
@@ -276,7 +276,7 @@ void CTeamTalkDlg::Disconnect()
 void CTeamTalkDlg::Login()
 {
     int cmd = TT_DoLoginEx(ttInst,
-        STR_UTF8(m_xmlSettings.GetNickname(STR_UTF8(LoadText(IDS_DEFAULTNICKNAME, DEFAULT_NICKNAME))).c_str()),
```
bear101, line 279: Same here. Change DEFAULT_NICKNAME to blank instead of removing handling of default values.

Client/TeamTalkClassic/settings/ClientXML.cpp
```diff
@@ -454,7 +454,7 @@ namespace teamtalk {
             return false;
     }
 
-    string ClientXML::GetNickname(std::string def_nickname/* = std::string()*/)
+    string ClientXML::GetNickname(std::string def_nickname = std::string())
```
bear101, line 457: I don't think this compiles.

Client/iTeamTalk/iTeamTalk/MainTabBarController.swift
```diff
@@ -533,7 +533,7 @@ class MainTabBarController : UITabBarController, UIAlertViewDelegate, TeamTalkEv
         
         let channelsTab = viewControllers?[CHANNELTAB] as! ChannelListViewController
 
-        let nickname = server.nickname.isEmpty ? (UserDefaults.standard.string(forKey: PREF_GENERAL_NICKNAME) ?? DEFAULT_NICKNAME) : server.nickname
+        let nickname = server.nickname.isEmpty ? (UserDefaults.standard.string(forKey: PREF_GENERAL_NICKNAME)) : server.nickname
```
bear101, line 536: Again. Change DEFAULT_NICKNAME

Client/qtTeamTalk/mainwindow.cpp
```diff
@@ -1919,7 +1919,7 @@ void MainWindow::Disconnect()
 
 void MainWindow::login()
 {
-    QString nick = ttSettings->value(SETTINGS_GENERAL_NICKNAME, QCoreApplication::translate("MainWindow", SETTINGS_GENERAL_NICKNAME_DEFAULT)).toString();
+    QString nick = ttSettings->value(SETTINGS_GENERAL_NICKNAME).toString();
```
bear101, line 1922: Same change SETTINGS_GENERAL_NICKNAME_DEFAULT to blank

Nardol:
I put a blank string instead of the initial parameter to allow still have "NoName#ID" instead of only "#ID" if no nickname is specified, the "NoName" mention is important IMHO.
I was not sure to understand right or if you meant changing the default value which is translatted for now and delete "NoName" completely.

bear101:
I think it's easier that you create a new pull-request where you set the following to blank:

- TeamTalkAndroid
  - Modify translation files and strings.xml and set "pref_default_nickname" to empty string
- TeamTalkClassic
  - Set macro "DEFAULT_NICKNAME" to empty string
- iTeamTalk
  - Set variable "DEFAULT_NICKNAME" to empty string
- qtTeamTalk
  - Set macro "SETTINGS_GENERAL_NICKNAME_DEFAULT" to empty string

Client/iTeamTalk/iTeamTalk/PreferencesViewController.swift
```diff
@@ -122,9 +122,6 @@ class PreferencesViewController : UITableViewController, UITextFieldDelegate, Te
         let settings = UserDefaults.standard
         
         var nickname = settings.string(forKey: PREF_GENERAL_NICKNAME)
-        if nickname == nil {
```
bear101, line 125: A "nil" in nickname will cause a crash.

Client/qtTeamTalk/mainwindow.cpp
```diff
@@ -1919,7 +1919,7 @@ void MainWindow::Disconnect()
 
 void MainWindow::login()
 {
-    QString nick = ttSettings->value(SETTINGS_GENERAL_NICKNAME, QCoreApplication::translate("MainWindow", SETTINGS_GENERAL_NICKNAME_DEFAULT)).toString();
```
bear101, line 1922: Keep SETTINGS_GENERAL_NICKNAME_DEFAULT but set SETTINGS_GENERAL_NICKNAME_DEFAULT to an empty string

Client/qtTeamTalk/mainwindow.cpp
```diff
@@ -1988,7 +1988,7 @@ void MainWindow::showTTErrorMessage(const ClientErrorMsg& msg, CommandComplete c
                 return;
             
             addLatestHost(m_host);
-            QString nickname = ttSettings->value(SETTINGS_GENERAL_NICKNAME, QCoreApplication::translate("MainWindow", SETTINGS_GENERAL_NICKNAME_DEFAULT)).toString();
+            QString nickname = ttSettings->value(SETTINGS_GENERAL_NICKNAME).toString();
```
bear101, line 1991: Keep SETTINGS_GENERAL_NICKNAME_DEFAULT but set SETTINGS_GENERAL_NICKNAME_DEFAULT to an empty string

Client/qtTeamTalk/mainwindow.cpp
```diff
@@ -3815,7 +3815,7 @@ void MainWindow::slotClientPreferences(bool /*checked =false */)
     if((TT_GetFlags(ttInst) & CLIENT_AUTHORIZED) &&
         TT_GetUser(ttInst, TT_GetMyUserID(ttInst), &myself))
     {
-        QString nickname = ttSettings->value(SETTINGS_GENERAL_NICKNAME, QCoreApplication::translate("MainWindow", SETTINGS_GENERAL_NICKNAME_DEFAULT)).toString();
```
bear101, line 3818: Keep SETTINGS_GENERAL_NICKNAME_DEFAULT but set SETTINGS_GENERAL_NICKNAME_DEFAULT to an empty string

Client/qtTeamTalk/mainwindow.cpp
```diff
@@ -4004,7 +4004,7 @@ void MainWindow::slotClientExit(bool /*checked =false */)
 
 void MainWindow::slotMeChangeNickname(bool /*checked =false */)
 {
-    QString nick = ttSettings->value(SETTINGS_GENERAL_NICKNAME, QCoreApplication::translate("MainWindow", SETTINGS_GENERAL_NICKNAME_DEFAULT)).toString();
```
bear101, line 4007: Keep SETTINGS_GENERAL_NICKNAME_DEFAULT but set SETTINGS_GENERAL_NICKNAME_DEFAULT to an empty string

Client/qtTeamTalk/preferencesdlg.cpp
```diff
@@ -469,7 +469,7 @@ void PreferencesDlg::slotTabChange(int index)
     {
     case GENERAL_TAB : //general
     {
-        ui.nicknameEdit->setText(ttSettings->value(SETTINGS_GENERAL_NICKNAME, QCoreApplication::translate("MainWindow", SETTINGS_GENERAL_NICKNAME_DEFAULT)).toString());
```
bear101, line 472: Keep SETTINGS_GENERAL_NICKNAME_DEFAULT but set SETTINGS_GENERAL_NICKNAME_DEFAULT to an empty string

Client/TeamTalkClassic/TeamTalkDlg.cpp
```diff
@@ -276,7 +276,7 @@ void CTeamTalkDlg::Disconnect()
...
+        STR_UTF8(m_xmlSettings.GetNickname("").c_str()),
```
bear101, line 279: No need for "". It will default to this value.

Client/qtTeamTalk/mainwindow.cpp
bear101: Move "QT_TRANSLATE_NOOP("MainWindow", "NoName")" from SETTINGS_GENERAL_NICKNAME_DEFAULT to a new macro called DEFAULT_NICKNAME in common.cpp.  The DEFAULT_NICKNAME should then be used in getDisplayName().

bear101:
Hm, I deleted my last message because I can see this will create conflicts.

bear101 (changes requested):

Client/iTeamTalk/iTeamTalk/PreferencesViewController.swift
Nardol: So this functionality cannot be implemented for IOS client?

PR #1176, opened 2021-12-20
Generalize getTitle function and use it also for TrayIcon

### LINKED ISSUE #1174: QTTeamTalk: generalize updateTitle to use also in trayIcon
### Description

What do you think about generalizing updateTitle functionality to use also for trayIcon?
I am talking about that because user may have several instances of TT opened and in this case it is not a trivial task to identify which instance they want to be restored.

### Application

- [X] qtTeamTalk
- [ ] TeamTalkAndroid
- [ ] iTeamTalk
- [ ] TeamTalkServer

beqabeqa473:
fixes #1174


bear101:
Looks good! I've tested on macOS.

PR #1178, opened 2021-12-21
Return entire item text with TTS using arrow keys in channels tree on MacOS

CoBC:


bear101:
Are you sure you prefer this way? If we later want to tell TTS whether it's a channel or user item then this information cannot be retrieved.

CoBC:
Yes, you're right, channel or user isn't specified, but it wasn't before any more. Some mac users I know have requested to have all available informations announces. I guess some informations like gender, state or status message can determinate if item is an user, and number of users can help to find channels.

PR #1186, opened 2021-12-27
Don't display server name if user account is invalid

CoBC:


bear101:
Why don't you want to display server name if account is incorrect? The server name is available as soon as you open a connection to the server. E.g. try "telnet tt4eu.bearware.dk 10335"

CoBC:
If your user account is incorrect, you can't do anything on the server, 
so I thought that is not really consist to have the server name displayed.

bear101:
A simpler way to implement this is to change MainWindow::getTitle(). So on line 2436 adds a check for CLIENT_AUTHORIZED, i.e.:
`else if ((TT_GetFlags(ttInst) & CLIENT_AUTHORIZED) && TT_GetServerProperties(ttInst, &prop))`

Client/qtTeamTalk/mainwindow.cpp
```diff
@@ -1996,6 +1996,7 @@ void MainWindow::showTTErrorMessage(const ClientErrorMsg& msg, CommandComplete c
                                      _W(QString(APPNAME_SHORT)));
             if(cmdid>0)
                 m_commands.insert(cmdid, CMD_COMPLETE_LOGIN);            
+            setWindowTitle(APPTITLE);
```
bear101, line 1999: Are you sure you need this one?

PR #1206, opened 2022-01-15
Stick to OPUS v1.3.1 on Raspberry Pi

bear101:


bear101:
This should also be changed to CMAKE_SYSTEM_PROCESSOR: https://github.com/BearWare/TeamTalk5/blob/master/Library/TeamTalkLib/build/ace.cmake#L40

Issue #1212, opened 2022-01-21
when the server have more than 100 persons, various bugs ocur. all os


nunotfc:
### Description

hello,
don't know if is a server bug but it happens with all clients android, classic, qt...
when are over 100 people on a server, Team Talk presents many difficulties. Sometimes stop responding when i enter in server. 
It also happens that it seems that Team Talk gives up loading users. I can see that it has for example 4 users on a channel, but when expanding the channel, no one appears. If you enter the channel, only appear me 
 If i go to connected users, I can see that users exist, and are on the same channel as me, even if I can hear them talking to me. 
finally,i got an error. 
* The internal message queue has overflowed.

### Application

- [X] qtTeamTalk
- [X] TeamTalkAndroid
- [X] iTeamTalk
- [X] TeamTalkClassic
- [ ] TeamTalkServer

### Platform

- [X] Windows
- [X] macOS
- [X] Android
- [X] iOS
- [X] Linux

### Expected behavior

load all normally with each person in each channel when a server have more than 100 connected.

### Actual behavior

users are loaded, but channels have sometimes 0 persons. The root channel starts colapsed. Sometimes the app sops responding. The internal messsage quewe is overflowing.

### Just connect more than 100 nickanames at same time in a server.

bear101:
Enable "beta software updates" to try out the latest client with the issue fixed.

PR #1217, opened 2022-01-25
Increase event queue size so it can handle initial login

bear101:
Set queue size to a full server with max users and max channels. This
should fix issue #1212.


bear101:
Enable "beta software updates" to try out the latest client with the issue fixed.

Issue #1218, opened 2022-01-25
Make timeout for solo-transmission configurable


bear101:
Also add 'nTransmitSwitchDelay' for SOLO_TRANSMIT

_Originally posted by @bear101 in https://github.com/BearWare/TeamTalk5/issues/22#issuecomment-288632988_

PR #1220, opened 2022-01-26
Configurable delay on transmit queue

bear101:
Fixes issue #1218

### DIFF
--- a/Client/qtTeamTalk/channel.ui
+++ b/Client/qtTeamTalk/channel.ui
@@ -6,8 +6,8 @@
    <rect>
     <x>0</x>
     <y>0</y>
-    <width>734</width>
-    <height>530</height>
+    <width>993</width>
+    <height>680</height>
    </rect>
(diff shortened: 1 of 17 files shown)


Client/qtTeamTalk/channel.ui
```diff
@@ -180,11 +180,22 @@
...
+             <widget class="QCheckBox" name="singletxchanBox">
+              <property name="text">
+               <string>No interruptions (no simultaneous voice transmission)</string>
+              </property>
+             </widget>
+            </item>
+            <item>
+             <widget class="QToolButton" name="singletxButton">
+              <property name="text">
+               <string>...</string>
```
CoBC, line 194: We can maybe add an accessible text on this button to explicit for why it is, like on other buttons named "...".

bear101: Yes, good idea. I've called it "Setup".

Issue #1216, opened 2022-01-25
Configurable log levels in server properties


bear101:
Make it configurable what is being logged on the server. The server's log events are listed here:
https://github.com/BearWare/TeamTalk5/blob/master/Library/TeamTalk_DLL/TeamTalkSrv.h#L192

### Discussed in https://github.com/BearWare/TeamTalk5/discussions/1215

<div type='discussions-op-text'>

<sup>Originally posted by **NightPurrer** January 23, 2022</sup>
I know there is a possibility to disable logging private messages without disabling the log. Anyone remember the switch to do that?</div>

PR #1222, opened 2022-01-27
Configurable log levels in ServerProperties

bear101:
Fixes issue #1216


CoBC (approved):
Why not add a checkbox to quickly enable or disable all log events? All the rest look good for me.

CoBC:
Ah, finally, after tested with server generated by artifacts, it's not completely working. Try to uncheck logging of private messages for example, save the configuration, then restart the server, logging is restored to default settings.

bear101:
Ah, yes, thanks. I forgot to save it to xml-file :)

PR #1233, opened 2022-02-06
update chinese translation

hejiadong0608:


bear101:
@hejiadong0608 Can you please use "git rebase" instead of "git merge". This pull request says it has 130 changes but it should only say there is 1 change.

hejiadong0608:
@bear101  What should I do now?

bear101:
You should write "git rebase upstream/master" instead og "git merge upstream/master". That is if you're using git from console. Then your latest git-commit will be the only difference from the official TeamTalk5-repo.

hejiadong0608:
@bear101 OK.

PR #1235, opened 2022-02-09
Ability to use WASAPI to use loopback sound device

### LINKED ISSUE #413: Allow selecting loopback input devices on Windows using WASAPI
Currently, selection of Loopback Input devices on windows is not possible, however WASAPI has this feature. It would be cool to see this added to tt as it makes sharing system audio easier.

bear101:
Fixes #413

### DIFF
--- a/ChangeLog.txt
+++ b/ChangeLog.txt
@@ -4,8 +4,10 @@
 
 Version 5.9, unreleased
 Accessible Windows Client
+- Ability for WASAPI to use sound output device as sound input device (loopback audio)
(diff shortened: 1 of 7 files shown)


CoBC:
Don't understand how does it work, I don't see any new device when I build from this branch and use WASAPI

bear101:
@CoBC I've uploaded new SDKs: https://bearware.dk/test/teamtalksdk/v5.9.0.5074/

CoBC:
Thanks, it works for me now.

PR #1240, opened 2022-02-18
String handling for Python

bear101:
Introduced ttstr() in TeamTalk5.py as a helper function to convert to/from utf-8 on macOS and Linux.

On Windows no string conversion is needed for TT_* functions because TTCHAR is convertible to Python's str-class.

@beqabeqa473 Do you only use Python on Windows? Or do you also have to make workarounds for Python strings?


beqabeqa473 (approved):
Looks good.

PR #1241, opened 2022-02-20
See MOTD as tooltip and currently active users

bear101:
I plan to run a service that updates the number of currently active users and MOTD every 5 minutes. At the moment the service is not running but you should be able to see the feature anyway.

Can you look into making the server list items accessible?
https://github.com/BearWare/TeamTalk5/blob/serverlist/Client/qtTeamTalk/serverlistdlg.cpp#L107

### DIFF
--- a/ChangeLog.txt
+++ b/ChangeLog.txt
@@ -11,6 +11,8 @@ Default Qt Client
 - Duration for voice transfer can now be set on "No Interruptions" channels
 - Modify server log events from "Server Properties" dialog
 - Added "Hear Myself" menu item to "Me" menu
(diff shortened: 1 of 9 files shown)


CoBC (changes requested):
I've added accessible text. I saw an issue, when I open server list dialog, focus is in host address combobox instead of server list. Previously it was directly in the list which was more practice.

bear101:
The focus problem was due to an error in sorting. I've fixed this now.

CoBC:
Hmmmmm, I'm still experiencing the focus issue now.

PR #1242, opened 2022-02-23
Publish private server

bear101:


bear101:
@CoBC The focus issue should work on this branch

CoBC (commented):
Focus issue still present. All the rest look good for me.

bear101:
Hm, I cannot see a difference w.r.t. the focus issue. If I connect to a "local" server then the selection and focus is restored. However, if the last server was a public server then the focus is not restored (this was also the behavior in v5.8.3).

bear101:
I take it the focus issue is also present on master-branch, so I'll merge anyway.

PR #1260, opened 2022-03-06
Hotkey for show/hide main window

### LINKED ISSUE #1141: Show/hide Qt TeamTalk using hotkey (as was available in Classic client)
If this hotkey already exists in the original QT client, please let me know how to set it, if not, I suggest adding it, because then you can extract the TEAM TALK window as you wish.

bear101:
Fixes #1141


CoBC (approved):
I just saw that when window is minimized and you use shortcut to restore, window is restored but don't get focus. Otherwise look good for me.

bear101:
@CoBC The missing focus, which platform does this happen to? I've tried on Windows and macOS, where the focus is back to TeamTalk. (I've tried with both minimize-to-tray and without)

CoBC:
Hmmmm I tried on Windows...

PR #1276, opened 2022-03-20
Refactor common methods in qtTeamTalk

bear101:


bear101:
I've modified processTTMessage() so it calls separate functions instead of a long switch-case that includes implementation. This hopefully makes it easier to navigate. Just merge if it looks good to you.

PR #1278, opened 2022-03-24
Longer text messages in qtTeamTalk

### LINKED ISSUE #1236: Increase character limit in text messages
### Description

Hello!

Over the years, something that has never failed to frustrate me about TeamTalk is it's hardcode-enforced character limit for user/channel messages. While such a limit may be useful on public servers where potentially-shady people may be connecting to send messages containing 1048575 characters, this limitation is impractical and rather inconvenient on private servers, where activities such as writing projects and the like may be taking place. With that being said, I would like to suggest the following:

* The ability to raise the message character limit (or disable it entirely) at configuration time.
* Messages sent to a channel that contain newlines should be treated as 1 message, similar to how user messages operate. Currently, any message containing newlines that gets pasted into the channel message edit box splits each line up into separate messages, which ends up creating for some potentially-spammy situations as far as sounds and screen readers are concerned.
* When logging of user messages is enabled, user messages fail to log properly. Instead of retaining the format of a timestamp (if enabled) and name on 1 line and the message on another line, the logfile ends up writing out user messages with the nickname, followed by the message (no timestamps, spaces or newlines between the nickname and message are included). This has only been tested with the classic client, so I am not sure if the issue persists in the QT version as well or not.

Thank you for reading, and I look forward to hopefully having these issues addressed ASAP!

### Application

- [ ] qtTeamTalk
- [ ] TeamTalkAndroid
- [ ] iTeamTalk
- [ ] TeamTalkServer

bear101:
Fixes #1236 in qtTeamTalk.

Merge if it looks ok. Remember to build a new server as well since the TCP protocol is changed.

### DIFF
--- a/ChangeLog.txt
+++ b/ChangeLog.txt
@@ -22,6 +22,7 @@ Default Qt Client
 - TTS message when navigating with arrow keys on macOS extends to all list
 - Media recording settings are restored at startup
 - Restore window position and size in commonly used dialogs
+- Text messages can exceed 512 characters
 - Fixed issue where all users and channels not were listed after initial login
 - Fixed labels for server abuse options in server properties
 - Fixed crash issue when using Cmd+4 on macOS 12.2
--- a/Client/qtTeamTalk/chattextedit.cpp
+++ b/Client/qtTeamTalk/chattextedit.cpp
@@ -101,7 +101,7 @@ protected:
 
 QString getTextMessagePrefix(const TextMessage& msg, const User& user)
 {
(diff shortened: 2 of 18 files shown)


CoBC:
Not really working as expected: when you send a long message, notification (for example TTS event) is correctly emited but the message don't appears in chat area.
Tested on Linux and Windows with the server build by GitHub's artifacts.

CoBC:
Oh, my mistake, just replace also the library and it works. Sorry.

PR #1282, opened 2022-03-28
fix shortcuts, and fix some typos.

rexya2017:


bear101:
I renamed Indonesia.ts to Indonesian.ts when you emailed me the file. Now it looks like you're adding the same file twice. Indonesia.ts and Indonesian.ts that is.

rexya2017:
Yes you right, previously I uploaded the file to github but I forgot to do a Pull Request, instead I sent it to your email. Actually I just want to fix some wrong shortcuts  and fix some wrong words in QT client, also fixing some word and adding string in TeamTalk client android.
what should i do so that the indonesian file for QT client is not added twice?
BTW i'm sorry, i'm new on github.

PR #1313, opened 2022-05-07
Choose language at first startup

CoBC:
@bear101 If you are agree with changes in this PR, can you look to make settings choose at first startup to directly take effects? For now we have to restart TT or open and close preferences to apply settings choose at first startup.

### DIFF
--- a/ChangeLog.txt
+++ b/ChangeLog.txt
@@ -6,6 +6,7 @@ Version 5.9.2, unreleased
 Accessible Windows Client
 Default Qt Client
 - Option to show server list on startup
+- Choose language at first startup
 Android Client
 iOS Client
 Server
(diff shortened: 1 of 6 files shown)


bear101:
Hm, it's still not working with the main menu. It only works when using preferences.

CoBC:
It works if we call retranslateui after switchlanguage. I can make a commit for this, unless you have a better idea?

CoBC:
It's called in mainwindow.cpp at line 632. Not sure we can called this in switchLanguage() method directly.

CoBC:
Yes, look good for me.

CoBC:
It's called in mainwindow.cpp at line 632. Not sure we can called this 
in switchLanguage() method directly.

Issue #1139, opened 2021-11-18
TeamTalk 5.8.2+ doesn't start on any version of Windows 7.


yarodin:
Teamtak 5.8.2.5058 doesn't work at any version of Windows 7 at all.
It display message about missing **api-ms-win-core-winrt-l1-1-0.dll** and shut down.

The latest version of TeamTalk working on all versions of Windows 7 is 5.8

bear101:
This is due to a change in the TeamTalk build system. In QMake this following line was added to support Windows 7: https://github.com/BearWare/TeamTalk5/blob/master/Client/qtTeamTalk/mainwindow.pri#L56 In CMake we have to find a similar command.

bear101:
For now Windows 7 users will have to install TeamTalk v5.8:
https://bearware.dk/teamtalk/v5.8/TeamTalk_v5.8_Setup.exe
Alternatively use TeamTalk Classic in v5.8.3 which doesn't have the DLL issue.

bear101:
This workflow produces a TeamTalk v5.9.2 beta version which runs on Windows 7:
https://github.com/BearWare/TeamTalk5/actions/runs/2323131961

bear101:
This version should run on Windows 7: http://bearware.dk/beta/TeamTalk_v5.9.2.5082_Setup.exe

bear101:
Fixed in v5.10

PR #1319, opened 2022-05-15
Build Qt client for Windows in default workflow

bear101:
Before Qt client was only built in windows-qt.yaml workflow.


bear101:
Workflow artifacts in [Actions](https://github.com/BearWare/TeamTalk5/actions) will also support Windows 7. #1139

PR #1333, opened 2022-05-18
Sharing TT URLs

bear101:
@CoBC shouldn't it be possible to generate a tt-URL to any channel on a server?

At the moment it's a bit confusing that the menu item says "This Channel" when another channel is currently selected.


CoBC:
Yes, good idea. And sorry for my late reply.

PR #1345, opened 2022-06-01
CheckState on models

CoBC:
@bear101 I've let the AccessibleTextRole on Qt5 because CheckState is only accessible in Qt6.


bear101:
@CoBC it's only on Qt6 you have the notification three times or did you manage to fix it?

Maybe only Qt::CheckStateRole should be used on Qt6 and then remove Qt::AccessibleTextRole?

CoBC:
@bear101 Notification is send three times on all Qt versions with code on issue-1293 branch
I'll make a commit to only use Qt::CheckStateRole on Qt6

bear101:
Shouldn't Qt::AccessibleTextRole also be removed in Qt5 then? To avoid extra notifications.

CoBC:
No, if we remove AccessibleTextRole on Qt5 users can't know if an item is enabled or disabled.

PR #1349, opened 2022-06-04
Relay user's voice stream from Advanced-menu

bear101:


CoBC (commented):
If I understand correctly, voice activation needs to be disabled to relay voice or stream of someone else. But I think the behavior can be improved, when we enabled relay, voice activation should be disabled, and the most important for me, when we try to reenable voice activation, the relay should be disabled and our own stream should take the priority.

bear101:
I've disabled the relay audio stream menu items if PTT or voice activation is active.

CoBC:
OK, maybe just also status message if unable to enable voice activation, but all look good for me.

CoBC (approved):

PR #1353, opened 2022-06-13
Receive only one notification for long messages

### LINKED ISSUE #1293: Send one notification even for long messages
### Description

When receiving a long message, we have one notification for each piece of message.
It is the case when sending or receiving public messages, Receiving private message and maybe for broadcast messages.

### Application

- [X] qtTeamTalk
- [ ] TeamTalkAndroid
- [ ] iTeamTalk
- [ ] TeamTalkClassic
- [ ] TeamTalkServer

### Platform

- [X] Windows
- [X] macOS
- [ ] Android
- [ ] iOS
- [X] Linux

### Expected behavior

We should have only one notification for all the message

### Actual behavior

There is one notification for each part of the message so for a very long message, we have numerous notifications instead of only one

1. With TTS enabled, receive a long message

CoBC:
Supposed to fix #1293, only for channel messages for now, but notifications are send three times.

### DIFF
--- a/Client/qtTeamTalk/chattextedit.cpp
+++ b/Client/qtTeamTalk/chattextedit.cpp
@@ -24,6 +24,8 @@
 #include "chattextedit.h"
 #include "settings.h"
 #include "appinfo.h"
+#include "utilsound.h"
+#include "utiltts.h"
 
 #include <QDateTime>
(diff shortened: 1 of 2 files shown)


CoBC:
Treated in #1359**

PR #1359, opened 2022-06-22
Merge TextMessages before sending to TTS

bear101:


bear101:
This handles channel text messages and broadcast text messages. I guess user text messages should not be sent to TTS until user opens text-message-dialog.

CoBC:
Yes, it's what I saw :). We can maybe move notifications in text message dialog, but in this case I don't know how users who disable text message dialog display when receiving a message will receive notifications.

bear101:
Yes, that's the issue. It will require a complete rewrite of merging text messages.

CoBC:
Alternatively, do you think we can make another function in chatTextEdit which just return merged message, without add this to a widget, so we can use this to pass string to TTS in main window.
This potential new function will also return only merged message without any extra content, because for now, when we receive a channel or broadcast message after your changes, you also have the date and author in the notification, but normally notification should only contains message.

bear101:
I think the best way is to create a new class that handles merging text messages.

We shouldn't have user-to-user text messages ending up in the ChatTextEdit instance intended for channel text messages.

bear101:
Merge for now and fix user-to-user later. It will not be a trivial task to fix.

PR #1362, opened 2022-06-25
Option to disable voice activation when inactive

CoBC:


ChrisDuffley:
@CoBC, I'm sort of confused. Reason being that there's already an option in General preferences where you can disable voice activation when your status changes to Away after an inputted amount of time. What exactly is this new one for?

CoBC:
Currently you can automatically switch your status to away when your 
computer is idle, but voice activation is still active in this case. 
This new option allows you to disable voice activation when computer is 
idle and you switch to away mode.

Client/qtTeamTalk/mainwindow.cpp
```diff
@@ -2376,12 +2376,16 @@ void MainWindow::timerEvent(QTimerEvent *event)
                     m_statusmode |= STATUSMODE_AWAY;
                     TT_DoChangeStatus(ttInst, m_statusmode, _W(statusmsg));
                     m_idled_out = true;
+                    if (isMyselfTalking() == TRUE && ttSettings->value(SETTINGS_GENERAL_INACTIVITY_DISABLE_VOICEACT, SETTINGS_GENERAL_INACTIVITY_DISABLE_VOICEACT_DEFAULT).toBool() == true)
```
bear101, line 2379: Why do you want `isMyselfTalking() == TRUE`? If it's false initially then voice activation is not disable while idle.

Client/qtTeamTalk/preferences.ui
```diff
@@ -174,6 +174,13 @@
               </property>
              </widget>
             </item>
+            <item>
+             <widget class="QCheckBox" name="disableVoiceActCheckBox">
+              <property name="text">
+               <string>Disable voice activation in case of inactivity</string>
```
bear101, line 180: Should be "Disable voice activation when inactive"

Client/qtTeamTalk/mainwindow.cpp
CoBC: If isMyselfTalking() == false, microphone is already disable, right? So we don't have to disable it if it is already.

bear101: isMyselfTalking() is true if you're currently transmitting voice: https://github.com/BearWare/TeamTalk5/blob/master/Client/qtTeamTalk/common.cpp#L343

I guess it's not what you're interested in here.

CoBC: Hmmm, the behavior like this is what I want. More, when switching back to available status, I check if isMyselfTalking is false, so it can be confusing if voice activation is re-enable even if it not be disabled automatically before. Not sure if I'm really clear...

ChrisDuffley:
@CoBC Just catching up on this PR. I thought this option was there already? Maybe not in Qt but probably in Classic? That's why I was confused.

Client/qtTeamTalk/mainwindow.cpp
bear101: If voice is not active (i.e. TT_GetFlags() has no CLIENT_SNDINPUT_VOICEACTIVE set) when you go to idle state, then once voice level increases and toggles CLIENT_SNDINPUT_VOICEACTIVE the client will start transmitting.

Reproduce:
1. Check "Disable voice activation when inactive"
2. Enable Voice activation and set level greater than zero
3. Wait for status-away to be toggled
4. Now start talking so voice level increases
5. The client will start transmitting

The way I figure it should work is:
1. Check "Disable voice activation when inactive"
2. Enable Voice activation and set level greater than zero
3. Wait for status-away to be toggled
4. Disable voice activation
5. Now start talking so voice level increases
6. The client will _not_ start transmitting

PR #1367, opened 2022-06-28
update Indonesian translation

HongjunTew:


lbk2907:
I will continue on tomorrow (GMT+0800) to remove the unneeded translations.

lbk2907:
I'll continue to remove the unneeded translations. I'll leave 1 more comment here after my last commit later. When you receive that comment, it's mean it are ready to merge to master.

Client/TeamTalkAndroid/src/main/res/values/strings.xml
```diff
@@ -2,7 +2,7 @@
 <resources>
 
     <string name="app_name">TeamTalk</string>
-    <string name="title_activity_server_list">TeamTalk 5 Servers</string>
+    <string name="title_activity_server_list">TeamTalk</string>
```
bear101, line 5: What is the effect of this change?

lbk2907:
I just change the name from TeamTalk 5 Servers to TeamTalk.

lbk2907:
If needed, I can revert back.

bear101:
But why did you make this change?

lbk2907:
I try to explain cause of my English. A few people has ask me. After they open the app, when they swipe to the left from the new server button, screen reader will read "TeamTalk 5 Servers". Have any way to change it to match the app name that is "TeamTalk"? And I said I try to make a pull request for it in the same time I’m updating the translation.

lbk2907:
Here you go. I'll push the last commit in this contribution. It ready to be merge.

CoBC:
I'm not sure what to think about the renaming of the server list screen on the Android client. I've never heard anyone complain about it, and if you really want to change it, I think you need to use TeamTalk5 instead of just TeamTalk.
@bear101 what do you think?

lbk2907:
> I'm not sure what to think about the renaming of the server list screen on the Android client. I've never heard anyone complain about it, and if you really want to change it, I think you need to use TeamTalk5 instead of just TeamTalk. @bear101 what do you think?

It will be good in terms of speech making things smoother for screen reader
@CoBC

PR #1375, opened 2022-07-06
Only send full message to TTS, not origin and full message

bear101:
This prevents the username to be pronounced twice


jeremyp3:
hello,

Would it be possible to delete the date pronunciation?

I display it in my client to have messages time, but I do not want to hear it in TTS, as was previously.

Thanks a lot

Jeremyp3

CoBC (approved):
For now it's probably a good workaround, it should be improve later

PR #1377, opened 2022-07-07
Restore initial TTS behavior for channel and broadcast messages

CoBC:
I made a workaround to restore initial TTS behavior for channel and broadcast messages. Simply get the full string and remove display name and eventually timestamp from it.


CoBC:
Yes, for now this is clearly a hack not really clean, but it seems that 
it's more confortable for everybody to have notifications like this. So 
as you prefere, we can keep it like this for this release and improve 
later, or just wait for next release.

jeremyp3:
I also prefer with this hack, it is less difficult in case of many channels messages.

I hope it will be merged while waiting for a real solution.

Thanks!

Jeremyp3

PR #1378, opened 2022-07-07
Text message container that handles merge-messages

bear101:
Separate text message container that merges text messages prior to passing them on.


CoBC (approved):
Just missing initial commented part in new files textmessagecontainer.cpp/h. Also, a conversion is probably required on Windows because I'm not able to build like this, but behavior tested with @jeremyp3 is OK.

CoBC:
Look good for me now.

PR #1382, opened 2022-07-09
Check state of voice activation when inactive

### LINKED ISSUE #1380: Voice activation is always enabled after idle
### Description

When "Disable voice activation during idle" is checked then voice activation is always enabled even though the user has never enabled voice activation.

### Application

- [X] qtTeamTalk
- [ ] TeamTalkAndroid
- [ ] iTeamTalk
- [ ] TeamTalkClassic
- [ ] TeamTalkServer

### Platform

- [X] Windows
- [X] macOS
- [ ] Android
- [ ] iOS
- [X] Linux

### Expected behavior

If VOX is disabled and "Disable voice activation during idle" is enabled/checked then after idle VOX should remain disabled.

### Actual behavior

If VOX is disabled and "Disable voice activation during idle" is enabled/checked then after idle VOX will be enabled.

1. Connect to server and join channel
2. Set idle timeout to say 5 seconds in Preferences
3. Check "Disable voice activation during idle" in Preferences
4. Disable VOX
5. Wait for idle
6. Become active again, i.e. come out of idle
7. VOX is now enabled

The problem is in mainwindow.cpp:2383

CoBC:
Supposed to fix #1380 but still not working...

### DIFF
--- a/Client/qtTeamTalk/mainwindow.cpp
+++ b/Client/qtTeamTalk/mainwindow.cpp
@@ -2379,7 +2379,7 @@ void MainWindow::timerEvent(QTimerEvent *event)
                     m_statusmode |= STATUSMODE_AWAY;
                     TT_DoChangeStatus(ttInst, m_statusmode, _W(statusmsg));
                     m_idled_out = true;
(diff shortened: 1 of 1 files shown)


CoBC:
Replaced by #1383

PR #1405, opened 2022-07-26
Option to configure sound events volume - Fixed #1393)

### LINKED ISSUE #1393: Ability to configure notification volume in QtTeamTalk
### Description

Add the ability to configure sound events volume level in QtTeamTalk.
This will possible in Qt6 using QSoundEffect::setVolume.

### Application

- [X] qtTeamTalk
- [ ] TeamTalkAndroid
- [ ] iTeamTalk
- [ ] TeamTalkServer

CoBC:
Work only with Qt6

### DIFF
--- a/ChangeLog.txt
+++ b/ChangeLog.txt
@@ -6,7 +6,7 @@ Version 5.11, unreleased
 Accessible Windows Client
 -
 Default Qt Client
--
+- Option to configure sound events volume
 Android Client
 -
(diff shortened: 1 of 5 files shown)


bear101:
Shouldn't we call it "Sound event volume level" instead of "Sounds volume"?

Also "Sounds pack" should maybe just be "Sound pack" or "Sound event package"?

PR #1410, opened 2022-07-28
Convert overloaded nativeEvent to Qt6

CoBC:
According to the Qt documentation, nativeEvent now use qintptr* result instead of long* result. This PR make this change.


bear101:
I think you need an "ifdef" based on Qt version because it doesn't compile on Windows using Qt 5.

PR #1412, opened 2022-07-30
Ability to start new instance using the same profile

CoBC:


ChrisDuffley:
Windows users have already been able to do this by just opening a new instance of TeamTalk. macOS and Linux users will definitely like this change though.

CoBC:
Replaced by #1417

PR #1429, opened 2022-08-10
Username/password are saved to .tt file in Generate .tt File dialog

bear101:


bear101:
@djgraco now it sets 'Entry Name' to the selection in server list.

PR #1433, opened 2022-08-14
Single column for showing checked/unchecked TTS events

bear101:
@CoBC Is there any need to two columns to show checked/unchecked TTS events in Qt 5?


CoBC (approved):
I don't think. One column can be sufficient as long as the accessibleTextRole is not broken.

PR #1449, opened 2022-08-18
Option to configure Qt TTS language

CoBC:
@bear101 I tried to refresh available voices when changing language in combobox, but it crashed. Can you try to do this part?

### DIFF
--- a/ChangeLog.txt
+++ b/ChangeLog.txt
@@ -17,6 +17,7 @@ Default Qt Client
 - Only 64-bit Windows is now supported
 - MacOS changed to Qt 6.4
 - Windows changed to Qt 6.4 means Windows 7 is no longer supported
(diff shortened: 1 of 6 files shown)


bear101:
For SETTINGS_TTS_LOCALE and SETTINGS_TTS_VOICE is it ok that I change these settings from int to string. It doesn't look like the index is stable, i.e. the order changes. Also we should sort locales.

Is it ok I make these changes? Changing SETTINGS_TTS_VOICE from int to string will break compatibility

CoBC:
It's ok. Maybe just add a note in change log to inform users that they 
have to reconfigure this settings.

bear101:
@CoBC can you have a look at the changes I've made?

CoBC:
Look good.

PR #1450, opened 2022-08-19
TTS message when trying to reconnect to server

CoBC:


bear101:
Shouldn't the TTS include IP-address and port? Then it's easier to trouble shoot. I.e. "Try to reconnect to 192.168.0.1 port 10336"

bear101:
@CoBC can you have a look at the changes I've made?

PR #1465, opened 2022-08-28
Improve translations for classroom changes

CoBC:
It is required for some languages (included French) to have diferent strings for diferent contexts. So I introduced this changes here.


bear101:
I'm ok with the change, but whether it works for all languages, I don't know.

Also it will be a bit difficult for translators to know what string will become %1 in this strings: "%1 channel messages!"

CoBC:
We can maybe just add a comment for translators to explain the %1?

bear101:
I've added comments to the translator and removed the "!" after each transmit type.

bear101:
Comments are added like this: tr("%1 has to blah", "this is my comment")

CoBC:
Look good for me

PR #1466, opened 2022-08-29
Updating Russian translation

artem59k:


bear101:
Sorry the translation string for transmission has changed again :(

PR #1467, opened 2022-08-30
update chinese translation

hejiadong0608:


bear101:
Sorry the translation string for transmission has changed again :(

PR #1477, opened 2022-09-02
Status bar and TTS event when changing subscription

CoBC:


ChrisDuffley:
TTS events are already implemented I think. Its the status bar event we're focused on I believe.

ChrisDuffley:
This would actually be quite handy.

CoBC:
TTS events are implemented when someone change a subscription for you, 
this PR add an event when you change a subscription for someone else. 
Example: "Subscription "Voice" disabled for CoBC".

Client/qtTeamTalk/mainwindow.cpp
```diff
@@ -2622,6 +2622,84 @@ void MainWindow::firewallInstall()
...
+        subType = tr("Intercept desktop");
+        subTypeTTS = TTS_SUBSCRIPTIONS_INTERCEPT_DESKTOP;
+        subTypeSB = STATUSBAR_SUBSCRIPTIONS_INTERCEPT_DESKTOP;
+        break;
+    case SUBSCRIBE_INTERCEPT_MEDIAFILE :
+        subType = tr("Intercept mediafiles");
+        subTypeTTS = TTS_SUBSCRIPTIONS_INTERCEPT_MEDIAFILE;
+        subTypeSB = STATUSBAR_SUBSCRIPTIONS_INTERCEPT_MEDIAFILE;
+        break;
+    default:
```
bear101, line 2700: This will leave "subType", "subTypeTTS" and "subTypeSB" uninitialized.

Client/qtTeamTalk/mainwindow.cpp
```diff
@@ -2631,17 +2709,27 @@ void MainWindow::subscribeCommon(bool checked, Subscriptions subs, int userid/*
 
     foreach(userid, userids)
     {
+        User user;
```
bear101, line 2712: Initialise "user"

Client/qtTeamTalk/mainwindow.cpp
bear101: Remove "default" and add "SUBSCRIBE_NONE" which returns.

Client/qtTeamTalk/mainwindow.cpp
```diff
@@ -2622,6 +2622,84 @@ void MainWindow::firewallInstall()
...
+        subTypeTTS = TTS_SUBSCRIPTIONS_INTERCEPT_VIDEO;
+        subTypeSB = STATUSBAR_SUBSCRIPTIONS_INTERCEPT_VIDEO;
+        break;
+    case SUBSCRIBE_INTERCEPT_DESKTOP :
+        subType = tr("Intercept desktop");
+        subTypeTTS = TTS_SUBSCRIPTIONS_INTERCEPT_DESKTOP;
+        subTypeSB = STATUSBAR_SUBSCRIPTIONS_INTERCEPT_DESKTOP;
+        break;
+    case SUBSCRIBE_INTERCEPT_MEDIAFILE :
+        subType = tr("Intercept mediafiles");
```
bear101, line 2696: "mediafiles" should be two words.

CoBC:
It's not already the case? I fsomeone change one of your subscription 
you receive a TTS notification

amirmahdifard:
@CoBC i didn't understand what you meant to say: i'm saying, that if possible, please also add a same kind as this for transmition toggles too

PR #1482, opened 2022-09-07
User rights in tree view

bear101:


bear101:
@CoBC Did you find a way to make the tree view "read only"? I guess it should be that for non-administrators.

CoBC:
Maybe change edit triggers, or itemiseditable property, following this topic:
https://www.qtcentre.org/threads/22511-QTreeWidget-read-only

CoBC (approved):

PR #1493, opened 2022-09-11
Fix Bluetooth security exception and set PendingIntent.FLAG_IMMUTABLE

bear101:


Client/TeamTalkAndroid/src/main/java/dk/bearware/data/Permissions.java
```diff
@@ -84,7 +86,12 @@ public static boolean setupPermission(Context context, Activity activity, int pe
                 errormessage = context.getString(R.string.permission_read_phone_state);
                 break;
             case MY_PERMISSIONS_BLUETOOTH:
-                stringPermission = Manifest.permission.BLUETOOTH;
+                if (Build.VERSION.SDK_INT >= Build.VERSION_CODES.S) {
```
bear101, line 89: @beqabeqa473 Is commit also how you read to request Android permissions when targeting Android SDK 31?

PR #1498, opened 2022-09-16
Auto-operators can update non-participating channels

### LINKED ISSUE #1496: Allow channel update if auto-operator and not in channel
### Discussed in https://github.com/BearWare/TeamTalk5/discussions/1478

<sup>Originally posted by **ahmednaalik51** September  3, 2022</sup>
Hi,
We have tried to report problems with channels to devs but they don't listen.
There is a bug in TeamTalk now that if a person is moved into a channel that has password, or moved at all, then if the connnection will become drop, then it comes back and says that channel password incorect.
This also happens if you set inisial channel with password in connect dialog sometimes it happens as well.
If you want to reproduce, put a admin and a user together, and move the user to the channel that has a password. Don't join the channel, just move them.
Turn off WiFi or whatever, and turn back on after connection drop. You will get error that says incorrect password.
Remember it only happens if they move to the channel, not join, just use the move command.
Also it is only happening in QT client.
Let me know if I needed to provide more information.
Did anyone else got this problems?
Thanks.</div>

bear101:
Also kick/ban from non-participating channels.

Fixes #1496

### DIFF
--- a/ChangeLog.txt
+++ b/ChangeLog.txt
@@ -38,6 +38,8 @@ Server
 - Passwords of auto operator channel are forwarded to clients
 - Operator status is resumed after disconnect when using BearWare.dk WebLogin
 - Subscription changes are written to log file
+- Auto operators can update channels they have not joined
+- Auto operators can kick/ban from channels they have not joined
 - Only 64-bit Windows is now supported
 - Fixed bug where bans on renamed channel has no effect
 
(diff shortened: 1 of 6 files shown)


amirmahdifard:
also auto  operators should be able to make someone temperery operator  in they channel using operator password if they not in channel

PR #1507, opened 2022-09-27
Only announce classroom changes in current channel

CoBC:


Client/qtTeamTalk/mainwindow.cpp
```diff
@@ -6258,81 +6258,83 @@ void MainWindow::slotChannelUpdate(const Channel& chan)
 
 void MainWindow::updateClassroomChannel(const Channel& oldchan, const Channel& newchan)
 {
-    auto userids = ui.channelsWidget->getUsersInChannel(newchan.nChannelID);
-    userids.push_back(TT_CLASSROOM_FREEFORALL);
-    for (auto id : userids)
+    if (newchan.nChannelID == TT_GetMyChannelID(ttInst))
```
bear101, line 6261: Wouldn't it be simpler to move this check to MainWindow::slotChannelUpdate?

```
if (newchan.nChannelID == TT_GetMyChannelID(ttInst))
    updateClassroomChannel(oldchan, chan);
```

PR #1510, opened 2022-09-28
Shortcut keys in 'Online Users' dialog

bear101:
@CoBC I don't know if you already tried this but I'm trying to make shortcut keys work in "Online Users" dialog.

With the changes in this merge request it's possible to do operations with shortcuts. However, shortcut will only work if no user is selected in the main-window (i.e. menu items in "Users" are disabled). Any idea how to get around this?

### DIFF
--- a/ChangeLog.txt
+++ b/ChangeLog.txt
@@ -24,6 +24,7 @@ Default Qt Client
 - Status bar and TTS event when changing subscription
 - Checkboxes in "User Accounts" dialog replaced by tree view
 - New profile now starts with blank settings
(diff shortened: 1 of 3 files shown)


CoBC:
Have you tried to define menu actions in online users dialog constructor 
and use it to build context menu when necessary? I saw that it's the 
correct way but I never tried for now.

bear101:
@CoBC Hm, I found a way that works on macOS.

CoBC (approved):
It works also on Windows. Look good!

CoBC:
@bear101 can you use the same way to add a shortcut to remove an user or a server in "User accounts" and "Server list" dialogs, this will fix the issue with context menu accessibility on MacOS.

PR #1512, opened 2022-09-29
Use 'Delete' key to delete user account and server list entry

bear101:


CoBC (changes requested):
I suggest to use Shift+Delete instead of just Delete, to be more logical with existing deleting action for files list for example which use SHift+Delete

bear101:
I see Delete-key as the intuitive way to delete. I wouldn't expect to use Shift-Delete.

E.g. Thunderbird use Delete-key to delete focused message. Not Shift-Delete.

CoBC:
In this case, maybe change also shortcut to delete file to just delete 
instead of Shift-Delete. Just to uniformize shortcuts.

bear101:
Won't this invoke "Delete File" from "User Accounts" dialog and when editing text?

CoBC:
We can disable action associated with files in files list don't have 
focus, right?

bear101:
I don't know. I think it will cause problems if you use Delete-key directly. I think it will be processed differently if it's hooked to a menu-item as keyboard shortcut.

CoBC:
Hmmm, OK, so just keep like this for now :)

PR #1545, opened 2022-10-20
Only set focus to text message dialog if main window has focus

CoBC:


Client/qtTeamTalk/mainwindow.cpp
```diff
@@ -2839,7 +2839,8 @@ void MainWindow::processTextMessage(const MyTextMessage& textmsg)
             if(dlg)
             {
                 dlg->show();
-                dlg->activateWindow();
+                if (this->isActiveWindow())
+                    dlg->activateWindow();
```
Nardol, line 2843: IMHO for accessibility reasons getting the focus automatically should not occurs at all in this case.

PR #1553, opened 2022-10-28
Option in "Change status" dialog to display file name as status msg when streaming

CoBC:


amirmahdifard:
@CoBC can you also fix some stuf for the server side and open some Pull requests for the server too?

CoBC:
For now I'm not familiar with developement on server side.

amirmahdifard:
oh, then i suggest you to also read server side and work on it too, because only bear101 and you are actively working on this project. i really like teamtalk, so please continu developing it. thank you!

PR #1558, opened 2022-10-31
Fix sound pack changes

CoBC:
Some fixes in this PR:
- Apply "User text message sent" and "Channel message sent" when changing sound pack
- Use absolute file path when changing sound pack to avoid problems if application is moved.


Client/qtTeamTalk/settings.h
```diff
@@ -28,9 +28,9 @@
 #include <QApplication>
 
 #if defined(Q_OS_WIN32)
-#define SOUNDSPATH                                     QApplication::applicationDirPath()+"/Sounds"
+#define SOUNDSPATH                                     "Sounds"
```
bear101, line 31: Doesn't this cause the use of a relative path instead of an absolute path?

The ChangeLog.txt status: "- Use absolute file path when changing sound pack"

CoBC: Yes, you're right, I mistaken in change log. Fixed.

bear101:
If you add TeamTalk5.exe to PATH variable and then run TeamTalk5.exe from some random location then it will not find "Sounds" folder. Are you sure the behaviour is preferred?

CoBC:
Hmmm, I don't really know. I changed this because an user was worry about the behaviour if he moved the app to another computer for example (especially for portable version), but in specific case you mensioned this change can cause problem. In your opinion what is better?

bear101:
It is easier to explain that Sounds-folder must be located with TeamTalk5.exe. It is the same for TeamTalk5.ini and languages folder.

PR #1564, opened 2022-10-31
Checkable QAction on sorting context menu (Fixes #1560)

### LINKED ISSUE #1560: QT Client: bug with chekked items in context menu
### Description

thanks to you for adding seperet context menus for user accounts list, for files list, and for servers list: in all these context menus that you've added, there is an option called (sort by) and it is sub menu: when you expanding it, there is some options that you can sort that list with for example name, or with size, and whatever: but there is a problem that if you selecting any of these, the screen reader doesn't announce (chekked or something at the end of it, so you can't understand wich one are you using: as an example, in your desktop sort by submenu in windows, if you select any option from there, next time if you open that menu and chek, it will say (chekked) at the end of that option: so please make it like that, because now we can't understand wich one is  selected: QT Teamtalk vertion: 5.11

### Application

- [X] qtTeamTalk
- [ ] TeamTalkAndroid
- [ ] iTeamTalk
- [ ] TeamTalkServer

CoBC:


Client/qtTeamTalk/mainwindow.cpp
```diff
@@ -5410,10 +5410,26 @@ void MainWindow::slotFilesContextMenu(const QPoint &/* pos*/)
 {
     QMenu menu(this);
     QMenu* sortMenu = menu.addMenu(tr("Sort By..."));
-    QAction* sortName = sortMenu->addAction(tr("&Name"));
-    QAction* sortSize = sortMenu->addAction(tr("&Size"));
-    QAction* sortOwner = sortMenu->addAction(tr("&Owner"));
-    QAction* sortUpload = sortMenu->addAction(tr("&Upload Date"));
+    QAction* sortName = new QAction();
```
bear101, line 5413: To avoid memory leak "sortMenu" should be parent of QAction: "new QAction(sortMenu)"

Client/qtTeamTalk/mainwindow.cpp
```diff
@@ -5410,10 +5410,26 @@ void MainWindow::slotFilesContextMenu(const QPoint &/* pos*/)
...
     QMenu menu(this);
     QMenu* sortMenu = menu.addMenu(tr("Sort By..."));
-    QAction* sortName = sortMenu->addAction(tr("&Name"));
-    QAction* sortSize = sortMenu->addAction(tr("&Size"));
-    QAction* sortOwner = sortMenu->addAction(tr("&Owner"));
-    QAction* sortUpload = sortMenu->addAction(tr("&Upload Date"));
+    QAction* sortName = new QAction();
+    sortName->setText(tr("&Name"));
+    sortName->setCheckable(true);
+    sortName->setChecked((ttSettings->value(SETTINGS_DISPLAY_FILESLIST_SORT, SETTINGS_DISPLAY_FILESLIST_SORT_DEFAULT).toString() == "name")?true:false);
```
bear101, line 5416: The mapping from "name" in line 5416 to "name" in line 5446 is a bit unclear. It would be better to use a "const string" to emphasise the mapping. I.e.:

```
const QString name = "name";
sortName->setChecked((ttSettings->value(SETTINGS_DISPLAY_FILESLIST_SORT, SETTINGS_DISPLAY_FILESLIST_SORT_DEFAULT).toString() == name)?true:false);
...
ttSettings->setValue(SETTINGS_DISPLAY_FILESLIST_SORT, name);
```

PR #1570, opened 2022-11-03
Only play "Channel silent" sound event in channels

CoBC:


jeremyp3:
I find this correction a pity. I find the sound practical to know if someone is on the server when we connect. 

I regret to have reported the bug (without wanting it) which was there since always :(

PR #1574, opened 2022-11-04
Display sort order in "Sort by" menu

CoBC:


amirmahdifard:
@CoBC but i think you fixed the sort by option problem, so what is thisone?

amirmahdifard:
when i look at the change log.txt in this pr, i don't understand the difrents

PR #1576, opened 2022-11-05
Don't focus text message window automatically at all

Nardol:
Focusing this window automatically can cause accessibility issue.
If user write a public message and receive a private one, the end of the message will be sent to the private conversation.


bear101:
I've tested this change on macOS and it appears to work ok.

CoBC (approved):

PR #1577, opened 2022-11-06
Disable upload action when not connected

CoBC:
This should fixe the issue discussed in #1575


amirmahdifard:
@CoBC thank you! please continu activly make pr for my ishoos, so we can make the Qt client as best as possible, so we can invit more users to use Qt client and put the classic to trash. this is how i can help you to make the Qt client more bettar

PR #1578, opened 2022-11-06
Playback sound events using TT_InitLocalPlayback

bear101:
Three options to play sound events:

1. Use default OS playback
2. Use TeamTalk sound devices doing one-by-one playback
3. Use TeamTalk sound devices doing overlapping playback

### DIFF
--- a/ChangeLog.txt
+++ b/ChangeLog.txt
@@ -7,6 +7,7 @@ Default Qt Client
 - Option to change status message when idle
 - Option in "Change status" dialog to display file name as status msg when streaming
 - See selected order on "Files list", "User account" dialog and "Server list" dialog context menu
+- Sound events can be configured to play one by one in "Preferences" dialog
 - Sound event when intercepted
 - TTS events when enabled/disabled voice activation and Push-To-Talk
 - See sort order in "Sort by" menu on files list, user accounts and server list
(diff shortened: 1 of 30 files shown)


CoBC:
@bear101 The "One by one" option seems to not work as expected. For example try to quickly enable and disable voice activation when this option is selected, you only hear some sounds, and not all sounds one by one as expected.
Also, don't forget to link label_47 with sndeventPlaybackComboBox for accessibility :)

bear101:
Please try again!

CoBC:
Last thing, the volume is very different in default mode and in other modes, I set the volume to 13%, and it's much louder in default than in other modes. The rest looks fine to me now.

bear101:
Volume of TeamTalk playback is controlled by master volume. Maybe that's why there's a big difference. Also if you mute master volume then you'll not hear playback.

CoBC:
Oh yes, you're right.

CoBC (approved):

jeremyp3:
@bear101 
under linux, in "one by one" mode the sound is cut while the sound is playing

jeremyp3:
https://user-images.githubusercontent.com/30219507/200932928-bb888368-583b-41da-97a4-a82e33ff3bad.mp4

Here is an example of recording in attachment

PR #1602, opened 2022-11-18
add F6 key for switch text edit and history

jeremyp3:


bear101:
@jeremyp3 Can you also put this feature in the documentation: Documentation/TeamTalk/userinterface.md

Otherwise people have no way of knowing.

bear101:
On macOS I just get the change status dialog when pressing F6 in text message dialog

CoBC:
This only appears first time "Text Message" dialog get focus, right? If you switch back to main window and then back again to text message dialog I think this issue doesn't appear anymore.

bear101:
I get the "Change Status" dialog every time. Both on initial focus of "Text Message" dialog and switching back and forth

PR #1608, opened 2022-11-23
Switch to message edit field when pressing a printable character in history fields

CoBC:


CoBC:
@bear101 What about this one? :)

bear101:
QWidget::keyPressEvent is declared 'protected'. You should not declare it as 'public' in a derived class.

You will have to create a new method in ChatLineEdit which then calls keyPressEvent()... and even then it sounds like a hack. This sounds like something that could easily break in later versions of Qt.

PR #1609, opened 2022-11-23
Android fix bluetooth on older platforms

beqabeqa473:


bear101:
@beqabeqa473 So ChangeLog should say something like "Bluetooth headset working again on Android 11 and below"?

beqabeqa473:
done

beqabeqa473:
@bear101 is it ok to merge this pr as most of users cannot use their mics in bluetooth headsets now?

PR #1662, opened 2022-12-21
Fix keyboard navigation on "File transfer" dialog

### LINKED ISSUE #1661: Qt client: the destination is not accessible
### Description

so while downloading a file, in that window, you can press tab and review some information such as file name, file size, and throo put: but there is also a destination that shows where the file is downloading, but we can't look at this destination using tab / shift tab keys, i saw this one only with nvda object navigation: so please make this one accessible for tab and shift tab keys so we can review it if we want too: windows 10: Qt teamtalk vertion: 5.12

### Application

- [X] qtTeamTalk
- [ ] TeamTalkAndroid
- [ ] iTeamTalk
- [ ] TeamTalkServer

CoBC:
Fixed #1661

### DIFF
--- a/ChangeLog.txt
+++ b/ChangeLog.txt
@@ -4,7 +4,7 @@
 
 Version 5.13, unreleased
 Default Qt Client
--
+- Fix keyboard navigation on "File transfer" dialog
 Accessible Windows Client
 -
 Android Client
--- a/Client/qtTeamTalk/filetransfer.ui
+++ b/Client/qtTeamTalk/filetransfer.ui
@@ -61,15 +61,15 @@
        <property name="text">
         <string>Throughput:</string>
        </property>
-       <property name="textInteractionFlags">
-         <set>Qt::TextBrowserInteraction|Qt::TextSelectableByKeyboard</set>
-       </property>
       </widget>
      </item>
      <item row="2" column="1">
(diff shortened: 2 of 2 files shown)


amirmahdifard:
@CoBC is this pr also gonna fix the destination for uploading a file too?, in tt5.12, we also have destination for uploading a file, that will show us wich channel we are uploading too, and it is not accessable like the download one: so, is this gonna fix both upload and download accessibility, or only download? i would like if both are fixed and accessible.

amirmahdifard:
@CoBC can you please answer my question? thanks

Nardol:
@amirmahdifard can you please wait for the answer more than 24 hours before remind? :slightly_smiling_face: :laughing:

CoBC:
> @CoBC can you please answer my question? thanks

@amirmahdifard Yes I can but FYI I'm in holidays and I'm not in front of my computer 24 hours a day. So, as Nardol mensioned, please wait.
For your question, the answer is yes, I've fixed for download and upload.

amirmahdifard:
@CoBC @Nardol sorry, i automatikly did this because normally i don't get answer from the developers so much, only sometimes: and, @CoBC thank you for fixing for both, i only asked this question because i couldn't it my self even tho i red the code that you've made: anyway, merry crismas!

PR #1667, opened 2022-12-22
Notification when sound device is added/removed

### LINKED ISSUE #712: Periodically update sound devices
Hello,
Thank you for the new features in TT 5.7, they are very great.
if I use teamtalk, it often happens that I plug in another sound device or change the standard device. The problem is, that Teamtalk doesn't detect the changed devices. In the QT-Client, I can use the refresh sound devices button in the settings.
But I use the classic client without such a button. It would be great to check for changed sound devices with a specific interval like other messengers like Skype. So it would be possible to change the standard device in the windows settings and teamtalk would use it to if it is set to primary sound driver in the settings.
Would it be possible to add this feature?

bear101:
This fixes #712

### DIFF
--- a/ChangeLog.txt
+++ b/ChangeLog.txt
@@ -5,6 +5,7 @@
 Version 5.13, unreleased
 Default Qt Client
 - Fix keyboard navigation on "File transfer" dialog
+- Reinitialize sound devices if active sound device is removed using WASAPI sound system
 - Moved "Refresh Sound Devices" from "Input Devices" submenu to "Sound Configuration" menu
 - Fixed Upload/download button and menu item not display for admin outside channel
 Accessible Windows Client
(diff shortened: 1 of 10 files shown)


bear101:
For now this only works for Windows.

If the sound device currently in use if being removed then we should re-initialize the sound devices. Preferably to the new default communication device.

amirmahdifard:
@bear101 as mentioned in this tikit, so now are you sure that adding this pr doesn't gonna cut and break  the voices in teamtalk?

CoBC:
We can maybe refresh sound devices automatically instead of require user to do it manually? Otherwise llok good to me for now

bear101:
The problem is if we close/open sound devices then audio recordings are restarted. So it's a bit of a trade off to automatically refresh sound devices.

bear101:
so I think we should only refresh sound devices if it's the active sound device that has been removed.

CoBC:
Yes, I agree with that.

amirmahdifard:
i also agree with @bear101 because the sound devices should only be refreshed automatikly if it's needed, as @bear101 s example, if the device is currently active and it got removed: but if it is not needed, for example, if a new sound device is added, or if a sound device removed but was not currently being used by the person, hen only a notification is enough

PR #1668, opened 2022-12-22
Move "Refresh Sound Devices" to "Sound Configuration" menu

bear101:
Not sure why it ended up in "Input Devices" submenu in the first place...


bear101:
Should it also have a keyboard shortcut? It can already be enabled as global hotkey.

PR #1673, opened 2022-12-23
Ability to copy text from channel path and MOTD variables

bear101:
Applies to 'Server Properties' and 'Channel' dialog

### DIFF
--- a/Client/qtTeamTalk/channel.ui
+++ b/Client/qtTeamTalk/channel.ui
@@ -39,6 +39,9 @@
             <property name="text">
              <string notr="true">TextLabel</string>
             </property>
+            <property name="textInteractionFlags">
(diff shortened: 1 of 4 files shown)


Client/qtTeamTalk/channel.ui
```diff
@@ -42,6 +42,9 @@
             <property name="text">
              <string notr="true">TextLabel</string>
             </property>
+            <property name="textInteractionFlags">
```
CoBC, line 45: I suggest to also define an accessible name for this dialog because it's not really clear to understand what is this text when tabing to them.

bear101: I've added 2x accessibleName

CoBC: Like this, screenreader only announce the accessible name, accessible name in this case have to be define in cpp file, as I did for file transfer dialog for example :)

CoBC (approved):

PR #1677, opened 2022-12-25
Enabled upload action following user right

### LINKED ISSUE #1675: Qt client: caused another problem after disable upload action when not connected PR
### Description

so i asked you please disable upload file option only if we are not connected to the server at all, but you disabled it also when we are connected to server but not in any channel: this caused another problem because in 5.11, i could download and upload files to other channels when i was online on the server but not in any channel, but with 5.12, if i'm online on the server but not any channel, i can download files from other channels, but i can't upload because upload option is disabled because i'm connected to the server but i'm not joind any channel: please edit your commit for that, and don't afect channels: if i'm connected to the server but not in any channel, the upload file should be enabled: the upload file should only be disabled if i'm absolutely not connected to any server at all.

### Application

- [X] qtTeamTalk
- [ ] TeamTalkAndroid
- [ ] iTeamTalk
- [ ] TeamTalkServer

CoBC:
Fixed #1675

### DIFF
--- a/ChangeLog.txt
+++ b/ChangeLog.txt
@@ -6,7 +6,7 @@ Version 5.13, unreleased
 Default Qt Client
 - Fix keyboard navigation on "File transfer" dialog
 - Moved "Refresh Sound Devices" from "Input Devices" submenu to "Sound Configuration" menu
-- Fixed Upload button and menu item not display for admin outside channel
+- Fixed Upload/download button and menu item not display for admin outside channel
 Accessible Windows Client
 -
 Android Client
--- a/ChangeLog.txt
+++ b/ChangeLog.txt
@@ -6,6 +6,7 @@ Version 5.13, unreleased
 Default Qt Client
 - Fix keyboard navigation on "File transfer" dialog
 - Moved "Refresh Sound Devices" from "Input Devices" submenu to "Sound Configuration" menu
+- Fixed Upload button and menu item not display for admin outside channel
 Accessible Windows Client
 -
 Android Client
--- a/Client/qtTeamTalk/mainwindow.cpp
+++ b/Client/qtTeamTalk/mainwindow.cpp
@@ -5447,8 +5447,8 @@ void MainWindow::slotFilesContextMenu(const QPoint &/* pos*/)
     QAction* download = menu.addAction(ui.actionDownloadFile->text());
     QAction* del = menu.addAction(ui.actionDeleteFile->text());
     auto index = ui.filesView->currentIndex();
-    upload->setEnabled((TT_GetMyUserType(ttInst) & USERTYPE_ADMIN) || ((m_myuseraccount.uUserRights & USERRIGHT_UPLOAD_FILES) && m_mychannel.nChannelID > 0));
-    download->setEnabled(index.isValid());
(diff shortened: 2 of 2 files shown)


amirmahdifard:
@CoBC then in this kase, also disable download action if the user doesn't have the user can download file right access: and if a user is channel  operator in a channel, they should be able to upload to their channel: if they have not joined any channel on the server

amirmahdifard:
then may be we should also disable download action for non administrator users if they are not in channel: @bear101 please review of this pr and murj it with out any other problems.

CoBC:
I already request a review using GitHub tools, so please just wait until 
the review is done.

amirmahdifard:
you didn't consider something: if a non administrator user is stil able to upload files to other channels if they are not in that channels: see these steps:
1: with a not admin user, join to a server:
2: join a channel called test for example:
3: now you want to upload a file to a channel called Audio box while your stil join in test channel: then, focus your nvda on channel Audio box, then go to files tab, press the aplication and upload the file: i just suggest that you revert your commits completely about this bug, just let the server side decides if anyone is allowed to upload or download or not.

amirmahdifard:
i think that 5.11 was good because the server decided these.

Nardol:
You are not really clear: you said admin cannot upload but now, you say non-admin can?
So admin was not able to upload but non-admin was?

Instead of writing 3 message by hour giving orders, it could be preferable to only have one which is really clear :slightly_smiling_face:

bear101:
I think we should only check (m_myuseraccount.uUserRights & USERRIGHT_UPLOAD_FILES). Never mind admin or current channel. The same should be done for download menu item.

amirmahdifard:
@Nardol i know what i commented: this topic is really complicated for you to understand: in 5.12, it was stil possible to them to upload  files to channel audio box while you are  in channel test, but it wasn't possible to a channel if they are not in any channel: anyway, bear101 has really a good point.

amirmahdifard:
@bear101 exactly, your point is really grate! i didn't like it when he dipended the current channel: the way you said, is exactly what i expected to see: it shouldn't stop you when your no5 in any channel: it should only dipends on the user right

amirmahdifard:
@bear101 bro, what about reverting it to 5.11? in that way, only server can give error to people when they are not allowed to upload or download. what do you think?

amirmahdifard:
@CoBC whell, as now i read this your new code, really i think that server side exactly does this: if a user doesn't have the upload or download access, the server doesn't let them  download or upload: so, reverting this to 5.11 or murging it, is same: how ever, i like it to let the server to do this job, but if you like to add this, it's ok as long as it doesn't create strange bugs again like 5.12

amirmahdifard:
oh! @CoBC also you can do something for delete file menu item too, but i'm not sure if it's gonna create bugs or not: you can incloud the user can create / modify all channels user right for delete file option: but not sure then if it's gonna create bug for temp channels: or, add a new user right called user can delete files, and incloud it for delete file item: adding a new user right reqiers server side code tho

amirmahdifard:
but generally, this one now look good to me with only user rights chek and not channels.

PR #1679, opened 2022-12-25
Prevent file upload outside own channel (unless admin)

bear101:
Limit file upload to users of the channel or admins (since they're the only ones who can see the file afterwards).


amirmahdifard:
@bear101 no, they can see  the files of other channels: please just don't murg this

amirmahdifard:
also limit default users and don't allow them download from a channel they have not joined, unless channel operators may be, same for upload may be

amirmahdifard:
can you do this for download  too? default users shouldn't  download  out of channel, only admin can.

PR #1695, opened 2023-01-07
TTS event on video transmission toggled

### LINKED ISSUE #1666: Qt client: tts event for vidio transmition enable / disable using controle + shift + v or toggling throo me menu or vidio tab start webcam chek box
### Description

so before, i asked you to do this for voice transmition, so now it would be good if we have it for vidio transmition too since vidio transmition is also togglable using controle shift v or vidio tab or me menu: a new option should be added to text to speech events tab in settings, like vidio transmition mode chek box, and if it's turned on, whenever we press controle shift v toggle the vidio, we should get tts message like vidio transmition enabled and also vidio transmition disabled: qt client vertion: 5.12

### Application

- [X] qtTeamTalk
- [ ] TeamTalkAndroid
- [ ] iTeamTalk
- [ ] TeamTalkServer

CoBC:
Fixed #1666

### DIFF
--- a/ChangeLog.txt
+++ b/ChangeLog.txt
@@ -9,6 +9,7 @@ Default Qt Client
 - Moved "Refresh Sound Devices" from "Input Devices" submenu to "Sound Configuration" menu
 - Fixed Upload/download button and menu item not display for admin outside channel
 - Accessible description on more dialog to avoid potential screenreader freeze
+- TTS event on video transmission toggled
 Accessible Windows Client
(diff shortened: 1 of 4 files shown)


amirmahdifard:
@CoBC i'm not sure, may be you are correct but, you spellet transmission, but i think it is transmition more correct, i am not sure

amirmahdifard:
nvm it's ok its already murged lol

PR #1719, opened 2023-01-17
Add server name in TTS event when disconnecting

### LINKED ISSUE #1717: Qt client: make the (disconnected from server) tts event variable for server name same as (connected to server) tts event
### Description

it's good that when we are connecting to a server, it says connected to server name: but for disconnected tts event, it's not mentioning the server name, it just says disconnected from server: please make this tts event to also mention server name, example: disconnected from little friends tt. @CoBC i asume that you can do this? this tts event was developed by you at first: thanks in advance!

### Application

- [X] qtTeamTalk
- [ ] TeamTalkAndroid
- [ ] iTeamTalk
- [ ] TeamTalkServer

CoBC:
Fixed #1717


amirmahdifard:
@CoBC thanks for this! it is really useful who are flying around so meny servers

PR #1730, opened 2023-01-18
add g size in list file

### LINKED ISSUE #1728: Qt client: also add GB to files size count.
### Description

currently, we have B, KB, and MB: see this file,
Name: Windows-10-21h2-64bit.iso, Size: 4387 M, Owner: amirmahdifard, Date: 2023/01/15 09:45  collapsed  row 62  Name  column 1  level 1
please also add gb, so it would be like this, it is more eezi and more understandable that way.
Name: Windows-10-21h2-64bit.iso, Size: 4 G, Owner: amirmahdifard, Date: 2023/01/15 09:45  collapsed  row 62  Name  column 1  level 1
or something like that, sorry about my example, i couldn't properly convurt MB to GB: but anyway, it is a good enough example for you to understand this and add it if you can: this would be really handy: version: 5.12 stable and 5.13 beta

### Application

- [X] qtTeamTalk
- [ ] TeamTalkAndroid
- [ ] iTeamTalk
- [ ] TeamTalkServer

jeremyp3:
This fix #1728

### DIFF
--- a/ChangeLog.txt
+++ b/ChangeLog.txt
@@ -12,6 +12,7 @@ Default Qt Client
 - TTS event on video transmission and desktop sharing toggled
 - Add missing labels on "Font color" and "Background color" buttons on "Video settings" dialog
 - TTS event when disconnect from server now includes server name
+- Add G size in file list
 Accessible Windows Client
 -
 Android Client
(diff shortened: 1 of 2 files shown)


amirmahdifard:
@jeremyp3 thanks really, since you are doing it, can you also add tb too lol, i know we will be never get that amount of file looool

jeremyp3:
No.

Le 18/01/2023 à 18:50, Amir Mahdi a écrit :

amirmahdifard:
its ok, g is enough because even my old laptop had 1 tb space lol, so it is not needed lol

PR #1732, opened 2023-01-18
File size now converted using Qt functions to support more size 

jeremyp3:
Suite of modification for fix #1730 I had only modified the accessibility part in my previous PR


CoBC:
@bear101 to avoid this kind of mistake in the future, I think we should use text already displayed in tree columns to build accessibility text. What do you think?

jeremyp3:
thanks @CoBC for assistance!

CoBC (approved):
Look good to me!

bear101:
Superseded by #1815

PR #1745, opened 2023-01-25
update Arabic translation

Muamalaljanahi:


bear101:
Thanks for the updates!

Medusa5000:
I wanted to translate the qt client as well, but unfortunately tt5_translator is not compatible with screen readers

PR #1746, opened 2023-01-25
Add events for adding/deleting user accounts

### LINKED ISSUE #1743: Add few events in Teamtalk protocol
### Description

Hello.

While working on one project, i've discovered that i need to get a list of user accounts to check for an account availability before adding it.
But the probelm is that if server will have a large amount of user accounts and if will fetch them in every ajax call, we will faze with non-needed resource consume.

I am happy to cash user account on client start, but in this case if an admin will delete or create user account manually within the client, i won't be notified of that change.

So i suggest to add events for adding and removing user accounts, and while on it, maybe add events for connecting and disconnecting users, so we can do some stuff depending on these events.

@bear101 what do you think about that?

### Application

- [ ] qtTeamTalk
- [ ] TeamTalkAndroid
- [ ] iTeamTalk
- [X] TeamTalkServer

beqabeqa473:
fixes #1743

### DIFF
--- a/Library/TeamTalkLib/bin/dll/TTClientMsg.cpp
+++ b/Library/TeamTalkLib/bin/dll/TTClientMsg.cpp
@@ -403,6 +403,26 @@ void TTMsgQueue::OnUserAccount(const teamtalk::UserAccount& account)
     EnqueueMsg(mb);
 }
 
+void TTMsgQueue::OnAddUserAccount(const teamtalk::UserAccount& account)
+{
+    ACE_Message_Block* mb;
+    IntTTMessage* msg = MakeMsgBlock(mb, CLIENTEVENT_CMD_ADDUSERACCOUNT,
+                                     0,
+                                     __USERACCOUNT);
+    Convert(account, *msg->useraccount);
+    EnqueueMsg(mb);
+}
+
+void TTMsgQueue::OnRemoveUserAccount(const ACE_TString& username)
+{
+    ACE_Message_Block* mb;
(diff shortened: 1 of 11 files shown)


beqabeqa473:
Tested with new client, everything works.
older clients won't get these events because then we will face with protocol incompatibility.
Another approach would be to limit only non-text clients, but don't know if it is needed.

beqabeqa473:
@bear101
When i was testing yesterday on a client 5.12, everytime when i was adding/removing user accounts, client was showing me an error about incompatibility of protocols. Because of that, i've added a check for version.

bear101:
Ah, ok. I'll fix that

amirmahdifard:
so, can we also have tts events when user accounts added / removed from the user accounts? or no?

beqabeqa473:
No, these are technical events for administrators, so no reason to do that.

amirmahdifard:
so now how i'm gonna be notifyed as an administrator if a user account is added or removed on my server?

beqabeqa473:
I have concrete usecase for what i've implemented thsi feature.
How you would be notified? like you were earlier.
Anyway, i don't see a reason and won't be implementing this.

amirmahdifard:
i don't know, just asking how we will be notifyed if a user account is added or removed: unless this change is only for developers.

beqabeqa473:
This pr is ment to add events in protocol sent by server to not fetch user accounts everytime as on the servers with large amount of user accounts there could be a performance impact. I needed a way to cache user accounts and update cache on their addition/deletion.

PR #1770, opened 2023-02-23
Move User ID to last position in Online users dialog

CoBC:
I think the ID isn't an information we want to know really quickly, so this PR move ID to last position, nickname is now the first info.


bear101:
Shouldn't we use getDisplayName() for nickname column?

bear101 (approved):

PR #1771, opened 2023-02-23
Option to enable/disable users of subchannel count in channel tree

CoBC:


bear101:
Instead of having and option to have one or the other, can't we just show user count as (6/10) or (6-10) to indicate 6 in channel and 10 in total?

PR #1772, opened 2023-02-25
Option to choose output mode for TTS events (with Tolk on Windows)

CoBC:
@bear101 Maybe you'll have a better idea for option name (I'm not really satisfy by the one I defined).


Client/qtTeamTalk/preferencesdlg.cpp
```diff
@@ -1684,17 +1687,29 @@ void PreferencesDlg::slotUpdateTTSTab()
...
+        bool hasBraille = Tolk_HasBraille();
+        bool hasSpeech = Tolk_HasSpeech();
         if (!tolkLoaded)
             Tolk_Unload();
         if(currentSR.size())
             ui.ttsForceSapiChkBox->setText(tr("Use SAPI instead of %1 screenreader").arg(currentSR));
         ui.ttsForceSapiChkBox->setChecked(ttSettings->value(SETTINGS_TTS_SAPI, SETTINGS_TTS_SAPI_DEFAULT).toBool());
+        ui.ttsOutputModeComboBox->clear();
+        if (hasSpeech == true && hasBraille == true)
+            ui.ttsOutputModeComboBox->addItem(tr("Speech and Braille"), "SpeechBraille");
```
bear101, line 1707: When you use a string as a value for a setting then there's no compile check that the setting is valid. If you use an enum instead you get a compile time check that the setting is valid.

Therefore change the three string values to the following enum:
enum
{
    TTS_OUTPUTMODE_NONE,
    TTS_OUTPUTMODE_SPEECH = 0x1,
    TTS_OUTPUTMODE_BRAILLE = 0x2,
    TTS_OUTPUTMODE_SPEECHBRAILLE = TTS_OUTPUTMODE_SPEECH | TTS_OUTPUTMODE_BRAILLE,
}

When you use a switch-case on an enum you also get a warning if some values are unhandled (which typically is an error).

Client/qtTeamTalk/utiltts.cpp
```diff
@@ -51,7 +51,12 @@ void addTextToSpeechMessage(const QString& msg)
     case TTSENGINE_TOLK:
 #if defined(ENABLE_TOLK)
         Tolk_PreferSAPI(ttSettings->value(SETTINGS_TTS_SAPI, SETTINGS_TTS_SAPI_DEFAULT).toBool());
-        Tolk_Output(_W(msg));
+        if (ttSettings->value(SETTINGS_TTS_OUTPUT_MODE, SETTINGS_TTS_OUTPUT_MODE_DEFAULT).toString() == "Braille")
```
bear101, line 54: Using enum should be changed to:
`
switch(ttSettings->value(SETTINGS_TTS_OUTPUT_MODE, SETTINGS_TTS_OUTPUT_MODE_DEFAULT).toInt())
{
    case  TTS_OUTPUTMODE_BRAILLE :
        Tolk_Braille(_W(msg));
        break;
    case  TTS_OUTPUTMODE_SPEECHBRAILLE :
        Tolk_Speak(_W(msg));
        break;
....
}
`

Client/qtTeamTalk/settings.h
```diff
@@ -320,6 +320,8 @@
 #elif defined(Q_OS_WIN)
 #define SETTINGS_TTS_SAPI                         "texttospeech/force-sapi"
 #define SETTINGS_TTS_SAPI_DEFAULT                 false
+#define SETTINGS_TTS_OUTPUT_MODE                         "texttospeech/output-mode"
+#define SETTINGS_TTS_OUTPUT_MODE_DEFAULT                 "SpeechBraille"
```
bear101, line 324: enum value for SETTINGS_TTS_OUTPUT_MODE_DEFAULT: TTS_OUTPUTMODE_SPEECHBRAILLE

bear101 (changes requested):

Issue #1775, opened 2023-02-28
Enhancement for TeamTalkPy


lbk2907:
### Description

Greetings to the developers of TeamTalkPy! Your library provides valuable functionality for Python developers looking to create chat applications and scripts. However, as a developer who has utilized your library, I have faced a significant issue with the limited functionality that restricts user and channel management capabilities. I understand that this is a complex task and I appreciate the work that has already been done. Nevertheless, I wanted to share my experience and the challenges that I have faced with the hope that it could help improve the library in the future.
As a developer utilizing TeamTalkPy library, I've faced a significant issue with the limited functionality that restricts user management capabilities. Although it offers some useful functions, such as joining a channel, sending messages, connecting to a server, etc, it lacks critical functions such as creating, deleting, or moving users, making it impossible to manage user accounts or kick or move users.
The absence of such user management functions poses a major obstacle for developers who want to develop custom applications or scripts that interact with TeamTalkPy. For instance, developing a chatbot that can manage user accounts or enforce certain rules requires access to functions for creating and deleting users. Without these functions, such bots' capabilities will be limited.
Moreover, the lack of user management functions makes it challenging to develop custom client applications that enable users to manage their accounts or move to different channels. Without these functions, client applications would also be limited in their functionality.
Overall, the missing user management functions represent a significant limitation for developers utilizing TeamTalkPy. Despite the library's useful functions, the absence of these critical functions makes it difficult to develop custom applications or scripts that require advanced user management capabilities.
Apart from the missing user management functions, there are other functions that are also absent from the current list of functions, such as managing channels, creating or deleting channels, retrieving information about channels, or changing channel properties, etc. Without these functions, it's difficult to build custom applications or scripts that require advanced management capabilities.
To further illustrate the issue, consider the example of the _DoLoginEx function that allows users to log in to a server. Python provides a corresponding higher-level function called doLogin, which accepts login information and calls _DoLoginEx to perform the actual login. Similarly, other functions such as _GetVersion, _SetLicenseInformation, _DoJoinChannelByID, _DoChangeNickname, and _DoTextMessage have corresponding Python functions in the library.
However, for functions like _DoKickUser, _DoMoveUser, _DoListUserAccounts, _DoNewUserAccount, and _DoDeleteUserAccount, etc, there are currently no higher-level Python functions provided by TeamTalkPy. This means that developers are compelled to write their Python functions to call these lower-level functions, which can be a time-consuming and error-prone process.
[...]

PR #1777, opened 2023-02-28
Added missing _domoveUser function in TeamTalk5.py

DalisonCPU:
To create a function in an app, I had to use the DoMoveUser function that wasn't added, so I included it in "TeamTalk5.py".
OBS: Probably the amount of lines that show as changed is because there were carriage-return changes, of Linux to Windows.

### DIFF
--- a/Library/TeamTalkPy/TeamTalk5.py
+++ b/Library/TeamTalkPy/TeamTalk5.py
@@ -1,23 +1,3 @@
-# Copyright (c) 2005-2018, BearWare.dk
-#
-# Contact Information:
-#
-# Bjoern D. Rasmussen
-# Kirketoften 5
-# DK-8260 Viby J
-# Denmark
-# Email: contact@bearware.dk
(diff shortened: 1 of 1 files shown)


bear101:
Please look at your diff. Something has gone terribly wrong.

yukie-nobuharu:
There are other functions missing from TeamTalkPy. please check this.
#1775

DalisonCPU:
Instead of saying it's terrible, you can try to help people help you.

DalisonCPU:
trying to help fix my git problem

Nardol:
> OBS: Probably the amount of lines that show as changed is because there were carriage-return changes, of Linux to Windows.

Why not respecting these cariage returns which are probably the cause of your ugly diff?
Every good code editor let the choice to keep the actual cariage return.
Are you sure your code was really based on the last code revision in this repository?

I hop it is not from Chat-GPT (as it is the last tendance lol) because its information are not from after 2020 IIRC.

Issue #451, opened 2019-08-10
SSL certificate verification


Simon818:
We can't really add domain verification without a client update, but how about a trusted certificates list and a prompt to trust a new certificate? That way if something changes, the user will know about it and can make an intelligent decision. Right now if someone's SSL cert changed or someone were intercepting my traffic for some reason, I would have no idea.

bear101:
Server side verification of clients has been merged to master (ssl_verify branch). Client-side support for private-key, certificate and CA file(s) is still pending.

@Simon818 how would you prefer that clients get their private key and certificate? Should users send a client certificate request to you or do you plan to send private key and certificate to clients?

Simon818:
Hmm, I think this should be a system a user can put in place as a login method if they wish and server owners can enforce if they want extra security (much like SSH keys). I wasn't actually thinking about client-side verification when I wrote this, though it is important and something I'd probably put in place. I think if the client could generate--or use--its own set of keys and allow the user to send theirs to an administrator, that would work. Am I misunderstanding? You also know way more about this than I do, so I would welcome your thoughts on how it should or shouldn't work. I just want to make sure it's secure but also easy for non-administrators to deal with and set up. If you can point me at the documentation in question I can also take a look and probably be way more constructive than this.

bear101:
I've written a guide here on how you configure encryption in the tt5pro server: https://bearware.dk/test/teamtalksdk/v5.6.1.5007/docs/C-API/serversetup.html#ttsrvencrypt

As you can see it's a complicated process to do peer verification.

bear101:
This issue will not be part of v5.7 since it has to be thought more through how it should be used.

bear101:
@Simon818 please try out the new client/server verification in pull-request #1780

Simon818:
I'm kind of a GitHub noob. Is there a build available? I don't have a dev environment set up and I don't see one linked there.

CoBC:
This is probably what you have to use:
https://github.com/BearWare/TeamTalk5/suites/12323997912/artifacts/653688957

bear101:
See #1920

PR #1780, opened 2023-03-01
Set up client TLS certificate/private key and CA

bear101:
This is client-side implementation of #451 for qtTeamTalk

DONE:
- iTeamTalk
- QtTeamTalk
- Include CA, cert and key in .tt files
- TeamTalk for Android


bear101:
It will verify that the TeamTalk server's certificate is issued by Let's Encrypt. Given that you have set up the client to have Let's Encrypt CA certificate.

jeremyp3:
hello,

I continued my tests with qt5.15 under Linux, and here is what I noticed:

- when I try to import a *.cer, the teamtalk window freezes, or teamtalk crashes (I had both behaviors)

- when I add the certificate by hand in Teamtalk5.ini, it's when I click on setup that teamtalk freezes or crashes. on the other hand, the verification works very well. it's just to display the certificate, that it seems to go wrong.

Hopefully  tests help,

Jeremyp3

bear101:
@jeremyp3 If you checkout "TeamTalk 5 Encrypted Server (US)" you'll see that it has CA certificate enabled and client will verify the certificate.

Also when exporting and importing .tt files you'll get the CA, client certificate and client certificate's private key.

PR #1785, opened 2023-03-06
Check "Show MOTD variables" checkbox by default

### LINKED ISSUE #1784: please check this (Show variables  check box to be checked by default when you press f9, because it breakes the motd everytime you update the server from there.
### Description

in classic client, when you pressed f9, this show motd variables was always checked by default, so if you had any variable in your server motd, it wouldn't become broken: please check this box by default in qt client, so we don't fix our motd variables everytime after updating the server: 5.12

### Application

- [X] qtTeamTalk
- [ ] TeamTalkAndroid
- [ ] iTeamTalk
- [ ] TeamTalkServer

CoBC:
Fixes #1784


bear101 (changes requested):
When you're admin and go to Server Properties then "Show MOTD variables" is checked but it doesn't show the variables.

PR #1792, opened 2023-03-15
Set focus to nicknameEdit when opening preferences dialog

### LINKED ISSUE #1788: The first edit box to change the nickname when open  preferences should be focused instead of a tab.
### Description

This is a weird bug since normally each dialog should show the first option when opening it.

### Application

- [X] qtTeamTalk
- [ ] TeamTalkAndroid
- [ ] iTeamTalk
- [ ] TeamTalkClassic
- [ ] TeamTalkServer

### Platform

- [X] Windows
- [ ] macOS
- [ ] Android
- [ ] iOS
- [ ] Linux

### Expected behavior

When opening the preferences it should show the first option in the general tab which is the edit box to add/change the nickname

### Actual behavior

It focuses on general tab, which isn't expected.

1: by pressing f4 or by right click on client/preferences. 2: the window will show up, focuses on the general tab instead of the first option

CoBC:
Fixes #1788

### DIFF
--- a/Client/qtTeamTalk/preferencesdlg.cpp
+++ b/Client/qtTeamTalk/preferencesdlg.cpp
@@ -260,6 +260,7 @@ PreferencesDlg::PreferencesDlg(SoundDevice& devin, SoundDevice& devout, QWidget 
     m_video_ready = (TT_GetFlags(ttInst) & CLIENT_VIDEOCAPTURE_READY);
 
     slotTabChange(GENERAL_TAB);
+    ui.nicknameEdit->setFocus();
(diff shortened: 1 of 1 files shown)


bear101:
I would find it more logical to navigate the tab-control first using arrow keys. Now you need to exit the General-tab before you can move to other tabs.

But ok if you find this more logical.

bear101 (approved):

CoBC:
I'm agree with you for logical point of view. But it's the behavior of most softwares, so I don't really know...

PR #1795, opened 2023-03-26
Update UI when enabling accessibility options at first startup

CoBC:


bear101:
Shouldn't the slotUpdateUI() at line 821 be enough?

CoBC:
It's a bit strange, an user has reported that when he enabled accessibility options, microphone vumeter wasn't disabled. SO, I suggested to open preferences, enable vu meter, press OK, reopen, uncheck option and press OK, and it worked. So, I thought this can fix the issue, but if you have another idea, we can maybe try something else.

amirmahdifard:
thanks: i also have this problem

bear101:
@CoBC I think I've fixed the issue. Also notice that startTTS() is called prior to screen-reader detection and setting Tolk as default TTS engine.

bear101:
Please rebase

Client/qtTeamTalk/mainwindow.cpp
```diff
@@ -778,6 +776,7 @@ void MainWindow::loadSettings()
         {
             ttSettings->setValue(SETTINGS_TTS_ENGINE, TTSENGINE_TOLK);
             ttSettings->setValue(SETTINGS_DISPLAY_VU_METER_UPDATES, false);
+            slotUpdateUI();
```
bear101, line 779: This should be removed.

Client/qtTeamTalk/mainwindow.cpp
```diff
@@ -822,6 +821,8 @@ void MainWindow::loadSettings()
 
     slotUpdateUI();
```
bear101, line 822: This line should be at the bottom of the function.

CoBC: If I move this to the bottom, this will be after eventual display of server list, so it's better to keep it here, right?

bear101: It wouldn't matter because slotClientConnect() calls slotUpdateUI() anyway.

PR #1796, opened 2023-03-28
Ban offline user

bear101:


CoBC:
I think it can be useful to choose if we want to ban from channel or server.
And, maybe, rename "kick" and "kick and ban" for offline users, because we can"t really kick an offline user.
Otherwise looks good to me

bear101:
Yes, ban from server menu item would be good.

If you toggle "Keep disconnected users" then you can ban users who left the server. This is quite useful when someone is spamming a server.

bear101:
Also notice the way I keep selection on rows. So when model is updated the previous row is selected/highlighed again

CoBC:
> Yes, ban from server menu item would be good.
  > 
  > If you toggle "Keep disconnected users" then you can ban users who left the server. This is quite useful when someone is spamming a server.
  Yes, but I talked about "kick", for offline users we can remove this, and rename "kick and ban" to "ban". It's maybe not really important but more logical :)

bear101:
Ah, yes, that's true. Kick doesn't work for offline users. This menu item should be disabled when they're offline. Actually I think also "View User Information" will fail when you try and do that on an offline user.

CoBC:
For offline users I think we should just have: "Ban from channel" and "Ban from server", all other items aren't avilable for offline users, so "View user information", "Message", "Operator", "Kick", "Store for move" should be disabled.

PR #1798, opened 2023-03-29
Ubuntu 18 is EOL

bear101:
GitHub is no longer supporting Ubuntu 18 on April 1st 2023, so these commits removes support for building Ubuntu 18 on GitHub.


Nardol:
What doe it imply for final releases on Debian? Will you still provide prebuilt binaries?
The Ubuntu 18 release is the only version which can run on Debian Buster and Bullseye, and it looks like the same for the future Debian Bookworm because of OpenSSL.

bear101:
I'll still provide a TeamTalk Pro server for Ubuntu 18

Nardol:
> I'll still provide a TeamTalk Pro server for Ubuntu 18

And about the client?
Which has the same behavior on Debian when trying to run the Ubuntu 20.04 version.

bear101:
The client is bye bye

bear101:
I cannot provide binaries for distributions that are no longer supported. There's also users who are complaining that TeamTalk no longer supports Windows XP. It's absurd.

bear101:
PR #1780 also cannot build on Ubuntu 18 because the Qt version on Ubuntu 18 doesn't support TLS certificate information.

Nardol:
> I cannot provide binaries for distributions that are no longer supported. There's also users who are complaining that TeamTalk no longer supports Windows XP. It's absurd.

But debian Buster is LTS and Bullseye is stable), so both still maintained.
If I totally agree with you on Windows, consequences are very different on Linux.

> PR #1780 also cannot build on Ubuntu 18 because the Qt version on Ubuntu 18 doesn't support TLS certificate information.

Which version supports it?
If I am correct, on Debian Buster we have QT 5.11.3 and Bullseye provides 5.15.2.

@jeremyp3 can you confirm?

ChrisDuffley:
> I cannot provide binaries for distributions that are no longer supported. There's also users who are complaining that TeamTalk no longer supports Windows XP. It's absurd.



Yes, I agree with you here. Why on earth is anyone even still running XP with TeamTalk in the first place? That has not been supported for almost 10 years now...

As for Ubuntu 18/20, 20.04 LTS (which I run because of cPanel and WHM on my server) is not EOL until 2030. If this version is not maintained for TeamTalk, there's no way I can upgrade the binary and then run it, just with 20.04 alone, from how this looks.

Thanks.

poretsky:
Ubuntu LTS releases life has been prolonged up to 10 years.

See https://ubuntu.com/esm for more information.

PR #1799, opened 2023-03-29
Ability to sort online users list

CoBC:
Currently only Id and Nickname sort is available. @bear101 should I add more sorting options?


bear101:
Up to you. There are 7 columns. Alternatively you need a more generic way to construct the menu items

bear101 (approved):

PR #1815, opened 2023-04-12
Add G size hint to file list (#1730)

bear101:


Client/qtTeamTalk/filesmodel.cpp
```diff
@@ -98,15 +108,7 @@ QVariant FilesModel::data ( const QModelIndex & index, int role /*= Qt::DisplayR
...
-        QString result;
-        if(m_files[index.row()].nFileSize>=1024*1024*1024)
-            result = QString("%1 G").arg(m_files[index.row()].nFileSize/(1024*1024*1024));
-         else if(m_files[index.row()].nFileSize>=1024*1024)
-            result = QString("%1 M").arg(m_files[index.row()].nFileSize/(1024*1024));
-        else if(m_files[index.row()].nFileSize>=1024)
-            result = QString("%1 K").arg(m_files[index.row()].nFileSize/1024);
-        else
-            result = QString("%1").arg(m_files[index.row()].nFileSize);
+        QString result = data(index, Qt::DisplayRole).toString();
```
CoBC, line 111: Hmm, size is the same as filename with this line, for example: "Name: test, Size: test, Owner: CoBC".

CoBC (changes requested):

Client/qtTeamTalk/filesmodel.cpp
bear101: Should be fixed now

CoBC (approved):

PR #1816, opened 2023-04-12
Option to remove unknown files from file storage in server

bear101:


CoBC:
Just read the change log in the PR :)

"- Added argument -cleanfiles to remove unknown files from file storage"

amirmahdifard:
@CoBC i red it, but it doesn't explane how to do it: how to clear the unknown files: i just asked what is the command for it

CoBC:
Yes, and it was what I answered. As the change log indicates, run the 
server with -cleanfile argument, so something like tt5prosrv -cleanfile

PR #1818, opened 2023-04-12
Use headerData and data to build QAbstractItemModel AccessibleTextRole

CoBC:
@bear101 If it's OK for you, I'll do the same for other models.


bear101:
Yes, please do that. When you call createIndex() then remember to check if internalId() is used (third argument to createIndex).

PR #1826, opened 2023-04-15
Build libvpx in GitHub workflow for Windows 

bear101:


hwangsihu:
https://github.com/cygwin/cygwin-install-action

It seems like you should install Cygwin instead of MSYS2.

bear101:
Woohoo! now it build. Thanks @hwangsihu

PR #1829, opened 2023-04-17
Implement remaining not implemented do methods and their respective events in TeamTalkPy

Nardol:
Implement client/server commands which were not implemented in TeamTalkPy and their respective event when applicable.
`CLIENTEVENT_STREAM_MEDIAFILE` is also added.


bear101:
Good job! Thanks a lot.

PR #1840, opened 2023-04-19
Option to ignore bot logs in/out

CoBC:
@bear101 Maybe there is a better way to check if an user is a bot or not?

### DIFF
--- a/ChangeLog.txt
+++ b/ChangeLog.txt
@@ -26,6 +26,7 @@ Default Qt Client
 - Confirmation before auto kick and kick and ban
 - Import .tt File button moved to "Store Server Information" group box in "Server List Dialog"
 - Filter servers by name or number of users in "Server List Dialog"
+- Option to ignore bots logs in/out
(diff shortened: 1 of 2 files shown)


bear101:
This is more like a hack than a solution. Maybe it should be possible to make a list of usernames to ignore. E.g. in Preferences -> Connection tab.

CoBC:
Hehe yes, it's the goal, but not sure to know how to do it easily :(

Muamalaljanahi:
Or if the client name is not equal to the official client names such as TeamTalk5droid or ITeamTalk, It doesn’t announce login_logout. In this case, not only stats we avoid most of bots like musicBot

amirmahdifard:
i liked @bear101 ' idea: in the client, we should be able to  have a list to import people or bots username to ignor their logged in and outs events, like, i will do this for spammy people.

CoBC:
Closing for now, will be done in another PR later.

PR #1865, opened 2023-05-10
Detect screen reader on Linux and ask user to choose more accessible settings

Nardol:
A DBus property is available to check if a screen reader service is running.
While this flag doe not necessary mean a screen reader is actually running, it means one has been started in the current session.
If this flag is set to true at first start of TeamTalk, ask user to choose if optimized setting have to be applied.
Another DBus property exists and Orca unset it when stopped, but this flag is for accessibility in general, not for screen reader specifically.
And probability to have the `ScreenReaderEnabled` property set in a session where no screen reader is running looks low in my opinion.
A thank to @CoBC who helped me for this contribution :slightly_smiling_face:


bear101:
Phew, it's very unreadable with all these #ifdefs. @CoBC do you have any idea if it will work. It looks to me like macOS will be broken by this change.

bear101:
It also looks like it will break Qt 6 on Linux

CoBC:
According to the Qt doc it seams that methods used here are still available on Qt6: https://doc.qt.io/qt-6/dbus-changes-qt6.html
I'm not really sure for MacOS, I think this will not break anything, but it's a bit strange to check if it's first startup on all OS and just change the setting value, maybe we have to check the platform before checking if it's first startup.
Anyway, we also maybe have to see how all this code of first startup could be simplify...

Nardol:
> I'm not really sure for MacOS, I think this will not break anything, but it's a bit strange to check if it's first startup on all OS and just change the setting value, maybe we have to check the platform before checking if it's first startup. Anyway, we also maybe have to see how all this code of first startup could be simplify...

I did it this way to leave as it is already in mainwindow.cpp, first startup is checked for all OS.

All I can be sure about Mac OS is GitHub actions pass: https://github.com/Nardol/TeamTalk5/actions/runs/4947722457/jobs/8847483861
Unfortunately I have no machine running Mac OS to test with, I'm currently looking how I could make a VM then I'll have to change all my habits because as fare as I know everything is different especially in term of use with VoiceOver.

CoBC:
Yes, it's me who originally checked first startup for all OS, but it's 
strange.

bear101:
Closing since this pull request has only been tested on Linux and maybe Windows (with Tolk).

PR #1871, opened 2023-05-11
Reduce size of common.h/cpp

bear101:
This merge request moves TT, XML and OS related functions into their respective util-files.


bear101:
@CoBC I can see you're in Paris. so enjoy! Say hello to Macron from me :-P

PR #1882, opened 2023-05-14
Add .gitignore files for Python

Nardol:
Add .gitignore files for Python clients and library.
When running tests for example, some `__py_cache__` are created, other elements are added for prequotion.
Also add for virtual env or when specifying a Python version with Pyenv in these directory.
Maybe these last should be added in the root .gitignore?


bear101:
All I see when running python is __pycache__ in Library/TeamTalkPy. Can you include only this in your merge-request?

Nardol:
Done, looks like I should really use env.sh each time instead of complicating myself...

PR #1883, opened 2023-05-14
Add Makefile to ttsamplepy

Nardol:
There is no makefile for ttsamplepy.
This PR adds one to have the same as other example.


Nardol:
Oops, I see my mistake...
I added a symbolic link to Teamtalk SO a long time ago to modify my Team§Talk installation structure without needing to modify `LD_LIBRARY_PATH, else it would not have worked.
Now this PR only adds a Makfile for ttsamplepy.

Issue #1414, opened 2022-08-02
Use TalkBack for TTS events on Android


CoBC:
Add option to use TalkBack instead of native Android TTS. Also add an option to choose if message should interrupt the previous one or not.
### Discussed in https://github.com/BearWare/TeamTalk5/discussions/1413

<div type='discussions-op-text'>

<sup>Originally posted by **CoBC** July 31, 2022</sup>
Hi,
currently, Android only support TTS via the TTS API. A method (announceForAccessibility, introduced in API level 16) can allow usage of accessibility service like TalkBack to announce TTS events.
Using TalkBack to announce TTS events can allow users to easily interrupt the notification if necessary. I propose to add an option to allow users to use TalkBack instead of native TTS API, so the behavior will be the same as using Tolk on Windows for example.
What Android users think of that? @bear101 Do you also have any point of view of that?</div>

CoBC:
Done by @beqabeqa473 in #1887

PR #1887, opened 2023-05-20
It is now possible to send tts-messages as Accessibility announcements

beqabeqa473:
which will be cought and spoken by running accessibility service

Needs to be tested.
In my testing, while speaking, it may skip some events.
Don't know how to resolve it yet.


beqabeqa473:
Fixes #1414

PR #1894, opened 2023-05-28
Add new events from v5.13 to JNI interface

bear101:


amirmahdifard:
hello,
@bear101 you are a greate developer, and i'm proofe of this: but can you add a pull request that so many people needs and that is specifying ban duration so we don't have to ban everybody permenantly? thanks, i hope this has been in to your atention: i can even bring so many teamtalk server owners rhat they need this ability if you think i'm only one who is asking for this.

PR #1899, opened 2023-06-01
Do not allow user to move user who is not logged in

bear101:
If a user has USERRIGHT_MOVE_USERS then it was possible to move a user who had no yet logged in.


amirmahdifard:
onnistly i don't know how you can focus on someone who is not in the server and move them, i wish other bugfixes and useful suggestions being added.

PR #1911, opened 2023-06-06
Remove public servers

### LINKED ISSUE #1872: Remove public servers from server list
### Discussed in https://github.com/BearWare/TeamTalk5/discussions/1849

<sup>Originally posted by **lucaschambers** April 25, 2023</sup>
Hello there! I wanted to bring to your attention that the public servers in the team talk application are becoming less reliable since the addition of unofficial servers. As a solution, I propose that you remove the entire public server feature and give users the ability to select their preferred language or country in order to find servers within their desired communities. This way, users can find as many servers as they want, similar to other conferencing applications.
Also, the developer doesn't need to bother themselves with adding or removing servers from the public servers all the time. This is because anyone can add their servers to the unofficial section, and people can find their community easier this way.
Thank you for considering this suggestion, and please let me know if you have any questions or comments.</div>

bear101:
For now public servers still appear because when public servers are unspecified then the web-service defaults to true.

Fixes #1872

### DIFF
--- a/Client/TeamTalkAndroid/src/main/java/dk/bearware/data/AppInfo.java
+++ b/Client/TeamTalkAndroid/src/main/java/dk/bearware/data/AppInfo.java
@@ -63,11 +63,10 @@ public class AppInfo {
                 + appversion + "&dllversion=" + TEAMTALK_VERSION + "&os=" + OSTYPE;
     }
 
-    public static String getServerListURL(Context context, boolean official, boolean publicc, boolean unofficial) {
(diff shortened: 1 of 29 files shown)


CoBC:
Client/iTeamTalk/iTeamTalk/PreferencesViewController.swift should also be edited to remove references to public servers, right?

bear101:
There "Public servers" applies to Official and Unofficial servers.

CoBC:
Yes, I think, this will probably be less confusing for users.

CoBC (approved):

PR #1950, opened 2023-07-14
Store state of "Hear myself" and restore it on channel join

### LINKED ISSUE #1949: Qt client: an option to automatickly hear our own voice after login in to the server so we don't have to always enable subscription voice for our self manully, also video too if possible.
### Description

some people are got used to always turn on the voice or video subscription for them selves so they can hear them selves while talking, an option should be added in to the teamtalk to automaticly enable voice subscription for people them selves, so they don't have to always press controle + 4 or controle + shift + 3 to start hearing them selves: this suggestion has been provided to me by one of my friends who needs such option, and i also thort it's a really cool suggestion so  i told you: please work on this if possible: thanks a lot.

### Application

- [X] qtTeamTalk
- [ ] TeamTalkAndroid
- [ ] iTeamTalk
- [ ] TeamTalkServer

CoBC:
Fixes #1949

### DIFF
--- a/ChangeLog.txt
+++ b/ChangeLog.txt
@@ -13,6 +13,7 @@ Default Qt Client
 - Removed public servers from server list
 - Ability to specify max duration of voice and media file streams (time out timer) in "Channel" dialog
 - Changed to Qt 6.6 beta1 on macOS
+- State of "Hear myself" restore at startup
 Accessible Windows Client
 - Classic client is no longer available in Windows installer
 Android Client
--- a/Client/qtTeamTalk/mainwindow.cpp
+++ b/Client/qtTeamTalk/mainwindow.cpp
@@ -3006,6 +3006,8 @@ void MainWindow::processMyselfJoined(int channelid)
         updateAudioStorage(false, AUDIOSTORAGE_SINGLEFILE);
         updateAudioStorage(true, AUDIOSTORAGE_SINGLEFILE);
     }
+
+    slotMeHearMyself(ttSettings->value(SETTINGS_CONNECTION_HEAR_MYSELF, SETTINGS_CONNECTION_HEAR_MYSELF_DEFAULT).toBool());
 }
 
 void MainWindow::processMyselfLeft(int /*channelid*/)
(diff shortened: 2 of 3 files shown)


amirmahdifard:
@CoBC thanks, currently, toggeling hear my self option from the menu or using controle shift 3 has no speech report: could you also add a tts for this so it will say hear my self enabled and hear my self disabled? thanks.

CoBC:
It's not really useful I think, if you start to hear yourself, so the option has been enabled.

amirmahdifard:
it would be useful in some cases, for example, you have muted your self with controle shift m and you forgot it, or your internet is really bad and you have delay, or your microphone is broken and it doesn't work, but a tts for this option would fix that: so if you are able too, please do that: thanks a lot to you friend.

jeremyp3:
I agree with that. If it's possible to add it, that's always better, as 
long as it can be deactivated.

amirmahdifard:
yeah exactly, i would like tts events because with hearing tts event, i can make sure of something: for example, if i hear my tts says hear my self enabled, then if i still don't hear my voice, i can make sure that may be the problem is from another place that my voice is not coming.

CoBC:
If added, this specific TTS event will be add to "menu actions" event, 
so you'll not be able to disable it in particular.

jeremyp3:
just create a dedicated event for it :)

amirmahdifard:
i mean, it's not an annoying tts or anything, it just speecks it when you toggel the option: so i don't think anyone will need to disable this tts.

CoBC:
@amirmahdifard You're wrong here, for example, for me, it will be an annoying TTS event, I don't want a lot of TTS events for all actions I do, so if a TTS is added for this specific action, I'll disable it. But, anyway, if I add a TTS event for this option, it will be in "menu actions", not sure it's really a good thing to add a lot of events in preferences because users can be confused.
@bear101 What do you think about it, should we add a TTS event for this toggle?

amirmahdifard:
@CoBC yeah you are true, i just said my personal opinion: i would like tts events for the actions because they makes sence more: but yes you are true, may be someone doesn't want to have it enabled: then i have a suggestion: add another toggel to the tts settings and call it like, extra tts events, and if somone disabled that, they  won't hear these: if they enabled it, they could have tts for everything.

amirmahdifard:
@CoBC never minde about tts for that: it's already says subscription voice turned on: i will just be greateful about that if you don't wish to add tts for that: thanks.

CoBC:
In fact, this TTS event already exists. So we have nothing to do more.

PR #1953, opened 2023-07-17
Restore 'Hear myself' upon new connection

bear101:
@CoBC I've simplified it a bit by just using subscribeCommon() which is already there. Also as I've mentioned earlier I'm not a big fan of [QObject::sender()](https://doc.qt.io/qt-6/qobject.html#sender).


bear101:
Notice that the behaviour of "Hear Myself" now is a bit different because you actually hear what is really happening.

CoBC (approved):
Ah, yes, sorry for QObject::sender :)

PR #1996, opened 2023-08-22
Minor Update for iTeamTalk English Translation

lbk2907:
I make this change is because other client like QTTeamTalk and TeamTalk5Droid use the word "logged in" instead of "logged on". Feel free to reject this pull request if you think no need to change.

### DIFF
--- a/Client/iTeamTalk/iTeamTalk/en.lproj/Localizable.strings
+++ b/Client/iTeamTalk/iTeamTalk/en.lproj/Localizable.strings
@@ -381,7 +381,7 @@
 "No Voice Activation" = "No Voice Activation";
 
 /* default nickname */
-"Noname" = "Noname";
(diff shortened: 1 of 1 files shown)


bear101:
This change will break the translation in all other languages. Either we should ignore this minor error or fix up all the other translation at the same time.

lbk2907:
My opinion, maybe we can fix it for all languages to syncronise the translation between other client. Like for example, the Malay translation is using the new term for it. I need help other translator too. Give me some time. I’ll double-confirm the strings if it still have anything to change. I’ll drop an another comment to let you know if all is complete.

CoBC:
In french, translation is already aligned with Qt and Android.

lbk2907:
Thanks for the update.

lbk2907:
Ok, all set for now. At least. 😊

PR #2017, opened 2023-09-10
Doxygen update

bear101:


amirmahdifard:
@bear101 also, mention about the new argument for the server that you've added, -cleanfiles, in to the documentation, because i haven't seen that in there,

PR #2048, opened 2023-10-14
User-right USERRIGHT_TEXTMESSAGE_USER for allowing private text message

### LINKED ISSUE #1955: User-right for disallowing users to send private and channel messages
### Discussed in https://github.com/BearWare/TeamTalk5/discussions/1934

<sup>Originally posted by **amirramezani79** June 27, 2023</sup>
Hi.
When creating/editing users, We can prevent users from voice , video and media transmition.
I suggest to add option for private and channel messages to be disabled for a user.</div>

bear101:
Server's XML file is automatically converted to allow old users to send text messages.

This fixes #1955


amirmahdifard:
@bear101 it is also being said that a user right will be exist to allow channel messages too, but this is only private message: when will you do that one as whel?

bear101:
@CoBC do you see a need to also be able to block channel text messages? Already there is classroom channels where you can block channel-text messages and also you can password-protect channels.

bear101:
Also we're running low on bits in [UserRight enum](https://github.com/BearWare/TeamTalk5/blob/issue-1955/Library/TeamTalk_DLL/TeamTalk.h#L1705)

amirmahdifard:
@bear101 classroom channels are limitting for only that channel, and emajin that someone took advantij in my server, so i wish to disable their channel messages ability, for every channel, and forever: this is really needed because currently anyway we limit someone, they can still lojicly send channel messages, even when their flud prevention timer had finished.

beqabeqa473:
Speaking from a view of functionality, why not?
As for ending bits, i think it needs to be rethinked.
We may add some other user rights in the future which we didn't think
about like this, and what would be the best way of doing this?

On 10/15/23, Amir Mahdi ***@***.***> wrote:
> @bear101 classroom channels are limitting for only that channel, and emajin
> that someone took advantij in my server, so i wish to disable their channel
> messages ability, for every channel, and forever: this is really needed
> because currently anyway we limit someone, they can still lojicly send
> channel messages, even when their flud prevention timer had finished.
>
> --
> https://github.com/BearWare/TeamTalk5/pull/2048#issuecomment-1763311332
>
> Message ID: ***@***.***>


-- 
Tell: +995593454005
Email: ***@***.***
Web: https://gozaltech.org
Skype: beqabeqa473

amirmahdifard:
@bear101 ok, so now lets consider everything and then we will see.
if we disable everyone's channel messages in a classroom channel, it disables permenantly, but it does it for everyone and not only that person that we want to disable it for.
if we disable only that person's channel messages, but everyone's channel messages are enabled, if that person logs out and come back the limitation will go away so it doesn't work like that and it's not permanent.
if we password protect channel, that person won't be able to join that channel at all and will still be able to send channel messages to other channels.
if we use flud prevention, they can't send channel messages but only for some amount of time and if that time passes, they can send again, and flud prevention afects everything such as joining channels, leaving, logging in, logging out, so this is also not good way.
so as you can see, the administrator of a server has no way of permanently and definetly disable someones channel messages and they can send channel messages at some point and the admin can't do anything about this case.
so with the things i mentioned here, please add a user right for channel messages as whel, so the user advantij prevention will be complete under administrators hand: we already have prevention of voice, file upload, download, vidio, multy loggin, and all others, so this would complete it.
i guess may be the user right user can change nickname should be removed because it is stupid, user can log out and change their name and log back in again so it does nothing really if you are worryed about running low of user rights

CoBC:
In anyway I think all options to limit potential spam are welcome, so 
yes, I think it can be a good idea to also allow to block channel 
messages :)

amirmahdifard:
@bear101 seriously user right of user can change nickname makes no sens for me because i don't know what it is useful for and it doesn't bother people if someone changes their names, and people can change nicknames when they are out of the server and come back, so this is not a prevention really, because the user will still able to do what they want, but only a little difrently: but about other user rights, if you disable them, the user has no way of still doing that until you reenable it back again.

Library/TeamTalkLib/teamtalk/Common.h
```diff
@@ -219,13 +221,13 @@ namespace teamtalk {
                             USERRIGHT_TRANSMIT_VIDEOCAPTURE |
                             USERRIGHT_TRANSMIT_DESKTOP |
                             USERRIGHT_TRANSMIT_DESKTOPINPUT |
-                            USERRIGHT_TRANSMIT_MEDIAFILE,
-                               
+                            USERRIGHT_TRANSMIT_MEDIAFILE |
+                            USERRIGHT_TEXTMESSAGE_USER,
```
CoBC, line 225: Shouldn't be also a "USERRIGHT_TEXTMESSAGE_CHANNEL" here?

bear101: Yes, thanks!

CoBC (approved):

PR #2053, opened 2023-10-15
Modifications to banning

bear101:


amirmahdifard:
i really need the ban duration in the teamtalk: i hope it gets done a s a p

bear101:
@CoBC I'll import the list of public VPNs, so be aware that the ban-list is going to be huge afterwards. We can, however, delete most of the existing bans, since their IPs are already in the list.

bear101:
We can also block all data centers if we only want "real users" on the server. Should we do that as well?

CoBC:
I suggest to do that as an option, if you're OK.

beqabeqa473:
Hi Bjoern.

How about bots? if user wants to have a music bot in their channel?

On 10/19/23, Corentin Bacqué-Cazenave ***@***.***> wrote:
> I suggest to do that as an option, if you're OK.
>
>
> --
> https://github.com/BearWare/TeamTalk5/pull/2053#issuecomment-1769291297
>
> Message ID: ***@***.***>


-- 
Tell: +995593454005
Email: ***@***.***
Web: https://gozaltech.org
Skype: beqabeqa473

Muamalaljanahi:
> We can also block all data centers if we only want "real users" on the server. Should we do that as well?

I don't think this is a good idea, because we usually host MusicBots on our virtual private server

prince18d:
> > We can also block all data centers if we only want "real users" on the server. Should we do that as well?
> 
> I don't think this is a good idea, because we usually host MusicBots on our virtual private server

It is possible to bring bots via localhost or 127.0.0.1

PR #2056, opened 2023-10-21
Spambot bans sync

bear101:


bear101:
@CoBC I'll clean up the ban list on the US server so no 1-2 octet networks will be banned.

PR #2060, opened 2023-10-23
Format dates and times to user locale

CoBC:
Affect banned users dialog, files model and user accounts dialog


CoBC:
Hmm, strange, it works as expected here. Do you also reproduce the 
freeze on user accounts dialog?

bear101:
I ran in Visual Studio Debug mode when I tested. On my mac it runs slow but ok.

Also notice that sorting does not work with the change. Danish locale date is DD/MM/YYYY but then 31 of any month is last.

CoBC:
I guess we can merge it like this for now. We have to rework sorting on banned users dialog. I'll try to work on it more later.

CoBC:
@bear101 Note, sorting by size is also broken, E.G. 995Mo is shown after 1.0Go and not before.
So, we probably have to rework on sort filters for size and for this PR with date format.
Do you have any idea on how to fix it, maybe we should return diferent values for display role and user role and force sort to be done on one role, but not sure.

CoBC:
@bear101 It works as expected for me now with date.

PR #2062, opened 2023-10-23
Only return files count for admins or current channel when pressing CTRL+G

CoBC:


amirmahdifard:
@CoBC this is incorrect: because 1: as teamtalk sdk, users are able to download and view files on other channels: so please show files of other channels to normal users as whell raather than only admins, like you are shoing users of other channels to everyone: i even openned an issue and explaned this.

PR #2064, opened 2023-10-25
Ban update

bear101:


amirmahdifard:
@bear101 i think adding ban duration would be a little simple and is really apirishated: currently, when we ban a user, it has ban time that showes the time the user is banned in: please add another time, called unban time: this time should be set by administrator when banning the user, or they can set it to no time at all to ban permanently: when they setted the unban time and banned the user, whenever the user tryed to login, the server should compare the unban time of that user to the current time, and if both timers are the same or the current time is grater, server should unban that user and let them login, but if current time is less than the unban time, show the banned from server error to that person.

amirmahdifard:
@bear101 modifying the online users dialog, please also add subscription sub menu to the online users dialog as whell because in some servers some users are not joining any channel and just spamming us with text messages and we can't mute them from the online users dialog

beqabeqa473:
Please do not post unrelated info on a concrete pr/issue.

On 10/25/23, Amir Mahdi ***@***.***> wrote:
> @bear101 modifying the online users dialog, please also add subscription sub
> menu to the online users dialog as whell because in some servers some users
> are not joining any channel and just spamming us with text messages and we
> can't mute them from the online users dialog
>
> --
> https://github.com/BearWare/TeamTalk5/pull/2064#issuecomment-1779329016
>
> Message ID: ***@***.***>


-- 
Tell: +995593454005
Email: ***@***.***
Web: https://gozaltech.org
Skype: beqabeqa473

amirmahdifard:
also, administrators are unavailable to intercept user who hasn't joinned any channel and only on the server, so that user can avoid interception of text messages

bear101:
@amirmahdifard From now on you can write one comment per week. Otherwise you'll be blocked from the project. Your input is not constructive and you have contributed nothing.

PR #2069, opened 2023-10-28
Add Korean translation to Qt TeamTalk, and update other Korean translation

ajackpot:
I added Korean translation for Qt TeamTalk and improved Korean translation for TeamTalk Android, iTeamTalk, etc.
Thank you.


bear101:
Thanks for the language update. I'll add it to the new features for v5.15

PR #2074, opened 2023-10-31
Multi selection in banned users dialog

bear101:


CoBC (commented):
Looks good, however I have 2 feedback:
- Shouldn't show all IP or username of selected users in text box?
- Do you see any way to indicates in an accessible mode (maybe modify accessible text role) if an user is selected or not in the tree?

bear101:
The idea of the text box is to quickly modify an existing ban. So e.g. change "192.168.0.100" to "192.168.0.0/24" so a subnet is be banned. Spammers typically use an entire subnet.

bear101:
W.r.t. selection then the model cannot tell if an item is selected. I've tried looking in "Online Users" and "User Accounts" dialog and also there is not a way to indicate selection.

Maybe we should do a TTS event in BannedUsersDlg::banSelectionChanged ?

CoBC:
> W.r.t. selection then the model cannot tell if an item is selected. I've tried looking in "Online Users" and "User Accounts" dialog and also there is not a way to indicate selection.
> 
> Maybe we should do a TTS event in BannedUsersDlg::banSelectionChanged ?

Yes, it's probably the best way

Issue #2077, opened 2023-11-01
TeamTalk crashes on Android 14 due to Foreground Exception


bear101:
When TeamTalk v5.14.1 (app version 125) connects to a server on Android 14 it causes the following exception:

```
FATAL EXCEPTION: main
Process: dk.bearware.gui, PID: 15893
android.app.MissingForegroundServiceTypeException: Starting FGS without a type  callerApp=ProcessRecord{91b6812 15893:dk.bearware.gui/u0a192} targetSDK=34
	at android.app.MissingForegroundServiceTypeException$1.createFromParcel(MissingForegroundServiceTypeException.java:53)
	at android.app.MissingForegroundServiceTypeException$1.createFromParcel(MissingForegroundServiceTypeException.java:49)
	at android.os.Parcel.readParcelableInternal(Parcel.java:4870)
	at android.os.Parcel.readParcelable(Parcel.java:4852)
	at android.os.Parcel.createExceptionOrNull(Parcel.java:3052)
	at android.os.Parcel.createException(Parcel.java:3041)
	at android.os.Parcel.readException(Parcel.java:3024)
	at android.os.Parcel.readException(Parcel.java:2966)
	at android.app.IActivityManager$Stub$Proxy.setServiceForeground(IActivityManager.java:6761)
	at android.app.Service.startForeground(Service.java:775)
	at dk.bearware.backend.TeamTalkService.displayNotification(TeamTalkService.java:402)
	at dk.bearware.backend.TeamTalkService.onCmdSuccess(TeamTalkService.java:831)
	at dk.bearware.events.TeamTalkEventHandler$8.processTTMessage(TeamTalkEventHandler.java:223)
	at dk.bearware.events.TeamTalkEventHandler.processEvent(TeamTalkEventHandler.java:724)
	at dk.bearware.backend.TeamTalkService$3.onTick(TeamTalkService.java:697)
	at android.os.CountDownTimer$1.handleMessage(CountDownTimer.java:145)
	at android.os.Handler.dispatchMessage(Handler.java:106)
	at android.os.Looper.loopOnce(Looper.java:205)
	at android.os.Looper.loop(Looper.java:294)
	at android.app.ActivityThread.main(ActivityThread.java:8177)
	at java.lang.reflect.Method.invoke(Native Method)
	at com.android.internal.os.RuntimeInit$MethodAndArgsCaller.run(RuntimeInit.java:552)
[...]

PR #2078, opened 2023-11-01
Change foreground type to include FOREGROUND_SERVICE_TYPE_MICROPHONE

bear101:
Android 14 with targetSdkVersion=34 and
android.content.pm.ServiceInfo.FLAG_STOP_WITH_TASK causes the following exception:

android.app.MissingForegroundServiceTypeException: Starting FGS without a type  callerApp=ProcessRecord{91b6812 15893:dk.bearware.gui/u0a192} targetSDK=34 ...
	at android.app.Service.startForeground(Service.java:775)
	at dk.bearware.backend.TeamTalkService.displayNotification(TeamTalkService.java:402)

### DIFF
--- a/ChangeLog.txt
+++ b/ChangeLog.txt
@@ -29,6 +29,10 @@ Server
 - Support for banning IP-address as subnet, e.g. 192.168.0.0/24 or 2a02:09b0:402b:eed8::/63
 - Owner of ban is stored
 
+Version 5.14.2, unreleased
+Android Client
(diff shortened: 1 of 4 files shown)


bear101:
On Pixel 7 Pro with API 34 I'm able to reproduce the crash described in #2077

bear101:
@beqabeqa473 any idea if this is the right way to fix the crash issue?

PR #2096, opened 2023-11-13
Add Dependabot configuration

Nardol:
Add [Dependabot](https://docs.github.com/en/code-security/dependabot/dependabot-version-updates/configuration-options-for-the-dependabot.yml-file) for GitHub action, Gradle and Swift.

### DIFF
--- a/.github/dependabot.yml
+++ b/.github/dependabot.yml
@@ -1,0 +1,23 @@
+---
+version: 2
+updates:
+  - package-ecosystem: "github-actions"
+    directory: "/"
+    schedule:
+      interval: "daily"
+  - package-ecosystem: "gradle"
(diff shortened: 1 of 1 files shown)


Nardol:
Not the same package ecosystems except for GH actions, but [here you can find some example of Dependabot pull requests](https://github.com/Nardol/pysmsboxnet/pulls?q=is%3Apr+author%3Aapp%2Fdependabot).

Nardol:
> Do you have an example where this comes into play?

For example, you won't have to update GitHub actions manually anymore, as soon as a new version is available, Dependabot will automatically open a pull request.

bear101:
We'll update the dependencies manually. Then we're more in control if things break.

PR #2099, opened 2023-11-17
Treat more command errors in client

CoBC:


bear101:
There's a build error

PR #2101, opened 2023-11-17
Remove Change Log Duplication

lbk2907:


bear101:
This is not a duplication

lbk2907:
Sorry if I am wrong. But I think I found the same line above with the date
instead of "unrelease". Maybe you can re-check back.

bear101:
Ah, ok. I had to expand more code to see it.

PR #2105, opened 2023-11-18
enabled change status option when not connected

amirmahdifard:
now users doesn't forced to connect to a server to modify their status message.

### DIFF
--- a/ChangeLog.txt
+++ b/ChangeLog.txt
@@ -4,6 +4,7 @@
 
 Version 5.15, unreleased
 Default Qt Client
+- enabled change status option when not connected
 - PulseAudio support on Ubuntu 22
 - User right to prevent private text messages
 - User right to prevent channel text messages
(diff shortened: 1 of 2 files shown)


CoBC:
As I said previously, this change isn't sufficient to do what you want. With this change if you try to change your status without be logged to any server, radio buttons and edit field in "change status" dialog are empty and not checked, and when you log in a server, you have a status of video transmission incorrectly added.

amirmahdifard:
can you review it so i can push another commit to add your requested changes?

amirmahdifard:
@CoBC i tryed this: if anything is needed, let me know, sorry if a little bad, because it is my first contribution: also, @bear101 please review and let me know what's needed next

CoBC:
You only remove checks for automatic status changes, E.G. when idle or when video transmission become active. It's not the correct file you're trying to edit.
Anyway, I can't just make a review to explain how to do, as I said previously, your request is a bit complicate to implement because we have to rewrite all change status dialog.
Sorry, but if you can't test your changes, it's not a good thing to open PR
@bear101 For now I'm in favour of closing this

PR #2113, opened 2023-11-27
added tt5svc_manager.bat and tt5prosvc_manager.bat files

amirmahdifard:
some users are not comfert to use the tt5svc_install.bat / tt5svc_uninstall.bat files because they are not providing so much to them, they are just simple scripts that directly does something as soon as you press enter: i have coded 2 more script, that helps some people to use and manage their servers more easyly: has options, and supports all of the teamtalk server's arguments: however, i didn't remove the current files tt5 install/uninstall bat files, so that anyone who got used to them will not be interupted: ask me anything you need, and i'll tell you.

### DIFF
--- a/Setup/Server/Windows/tt5prosvc_manager.bat
+++ b/Setup/Server/Windows/tt5prosvc_manager.bat
@@ -1,0 +1,60 @@
+@echo off
+title TeamTalk 5 Professional NT Service Manager
+fltmc 1>nul 2>nul || (
+cd /d "%~dp0"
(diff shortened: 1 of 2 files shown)


amirmahdifard:
@CoBC how can i request a review from other developers, and sighn this pr to my self?

amirmahdifard:
and, @CoBC i think i asked you a question, but you ignored me. i wish someone could explane to me why everyone is like this with me

PR #2151, opened 2024-02-10
Debug workflow

bear101:


bear101:
I'm using this branch. Don't delete it

hwangsihu:
Okay, I'm sorry.

BTW, wouldn't it be better to modify the rules and merge them into master?

For example, debug-windows-* branches enable only Windows builds, debug-linux-* branches enable only Ubuntu builds, etc.

bear101:
What should the rule look like in the workflow?

hwangsihu:
HMM, Can I edit your branch? Than, I'll trying adding this rule.

PR #2161, opened 2024-02-20
fixes all client lojic checks

HanaDadfar:
in the qt client, i found some incorrect lojic checks for options, which some of them cause some Strange problems. For example, one of them were caused by this line we had in the server options section
ui.actionBannedUsers->setEnabled(me_op || (userrights & USERRIGHT_BAN_USERS));
as you can look and reproduce, with this line, even if a teamtalk user doesn't have the "user can ban users from server" right, if that user simply becomes channel operator in any channel either temporary or permenantly, that user would be able to klick "banned users" option in the server submenu, which has now been fixed. However, i think i can guess why this code mistake had happened. You were supposed to use this condition for the "banned users from channel" option, but it was rongly aplyed to the "banned users from server" option, which also now has been fixed, i have moved everything to their proper places.

### DIFF
--- a/ChangeLog.txt
+++ b/ChangeLog.txt
@@ -4,6 +4,7 @@
 
 Version 5.16, unreleased
 Default Qt Client
+- fixed some client lojic checks, such as you had access to server's banned users when you were an operator of a channel
 - Option to show server's message of the day in a welcome dialog box
 - Fixed "Gender" property not being saved in .tt files
 Android Client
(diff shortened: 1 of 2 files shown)


bear101:
It is better that users see a "Not authorized" error dialog box instead of a disable menu item. If you're not authorized to delete a file on Windows then the "Delete" menu item is not disabled. Instead you get an error dialog box when trying to do so.

PR #2163, opened 2024-02-27
Android: Include the client version in UserPropActivity

Muamalaljanahi:


bear101:
@CoBC shouldn't the new string only be included in res/values/strings.xml ? Or how do you find the new translation strings in values-fr/strings.xml ?

CoBC:
Yes it's better if the new string is only add in values/strings.xml, 
next translators just have to check changes in this file and report it 
to specific language files

bear101:
@Muamalaljanahi can you remove the "version" translation from other language files, i.e. except res/values/strings.xml

PR #2167, opened 2024-02-29
Option to announce server name for some server related TTS events

CoBC:
Currently only implemented for user logged in and out


bear101:
Shouldn't this option just be default on? I guess it's convenient when you're running multiple TeamTalk instances.

CoBC:
Yes, I agree, but I thought it's better to not change current behaviour 
when we add new option. But I can change it to default on if you think 
it's better!

bear101:
I think it should be the default behaviour without the checkbox.

I recommend calling limitText(_Q(m_srvprop.szServerName)) so the text message doesn't get too long.

PR #2170, opened 2024-03-01
fixed join channel option

HanaDadfar:
now when focused on a user and not a channel, or when not connected and accessing the channel menu from the alt menu, the label of join channel option will no longer be announced "Leave Channel", it'll be announced "Join/Leave Channel" instead.

### DIFF
--- a/ChangeLog.txt
+++ b/ChangeLog.txt
@@ -4,6 +4,7 @@
 
 Version 5.16, unreleased
 Default Qt Client
+- Fixed join channel / leave channel option
 - Option to show server's message of the day in a welcome dialog box
 - Don't display server name when disconnected from server where not authentified
 - Option to show channel topic in channel list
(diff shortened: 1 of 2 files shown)


HanaDadfar:
this pr is now ready, unless you agree with the faqt that i remove the shortcut for the option when called join/leave channel.

bear101:
I don't see why this pull request is necessary. If not channel is selected in channels tree nothing will happen:
```
void MainWindow::slotChannelsJoinChannel(bool /*checked=false*/)
{
    Channel chan;
    DoubleClickChannelAction dbClickAct = DoubleClickChannelAction(ttSettings->value(SETTINGS_DISPLAY_CHANDBCLICK, SETTINGS_DISPLAY_CHANDBCLICK_DEFAULT).toUInt());
    if (!ui.channelsWidget->getSelectedChannel(chan))
        return;

```

HanaDadfar:
please, read my point in the discription again. When no channel is selected and we are looking at channels menu from the alt menu, the option is called leave channel, which is confusing. I fixed the name so that it alwase has a correct name. Please read this again, Thanks.

PR #2171, opened 2024-03-02
Android: new option to use the old event sounds instead of the current Default soundpack

Muamalaljanahi:


HanaDadfar:
@Muamalaljanahi hi
Thanks for your works for teamtalk android which needs a lot of support currently.
Is it possible for you to also add the transmit controles and intercept subscriptions options to the user prop activity also, those are really needed, now only in windows people can manage classroom channels, or mute or unmute a user. These can be in the user prop, and anyone who have access to them will be able to toggle them.
Thanks so much! I'm only limited to add new things in qt teamtalk because for some reasons i'm not allowed to have android studio in my system. Thanks again.

bear101:
Instead of new and old audio files it should be possible for the users to choose their own audio files (sound pack).

Muamalaljanahi:
I have thought about it but there's still an issue with accessing files on Samsung devices. For instance, on my Samsung device, I'm unable to upload or stream files.

PR #2174, opened 2024-03-02
Fix online users dialog compilation warnings with Qt6

CoBC:


bear101:
it does not compile on Qt 5.15

CoBC:
Ah yes, I saw with GitHUb action. Do you think we should make a 
condition to build without warning on Qt6 and fix build on Qt5, or just 
keep as previously and let warning here?

bear101:
I think we can only avoid the warning by making a duplicate implementation based on #if-macro Qt5 vs Qt6. I've also tried to look for a solution but was unable to without duplicate code.

HanaDadfar:
why qt5 is not completely dropped when you are already up to qt6? i mean why just 1 qt version at a time is not used?

HanaDadfar:
oh, is qt6 not compatible with linux or it has any problems?

CoBC:
We're waiting for a LTS version of Ubuntu with Qt6.4 or higher

HanaDadfar:
oh, then i hope that happens soon because i hate how when i built qt teamtalk it downloaded qt5 for me also. it was pretty strange for me, and stil i don't know what to do about the tolk problem that i have, i currently can't develop any tts related features to teamtalk because i can't test. Completely it doesn't know what the tolk is, putting it in the folder or not putting it doesn't make any difrence for the teamtalk. In both cases it runs, and doesn't have tolk support. But when i remove the tolk.dll from the folder of teamtalk5.exe which is built by teamtalk, I get an error saying tolk.dll was not found. Reinstalling may fix this problem, but when i don't put the tolk.dll in the folder of teamtalk5.exe which is built by me, everything runs properly and i get no errors at all. And even in v studio when i build, i get no errors or warnings. But When teamtalk5.lib was missing, i got an error saying teamtalk5.lib was not found in the blaablaa directory. But now that my tolk is missing, i don't get any errors, so i don't know where to put the tolk.lib file which i got when entered this cmake be qt build command. And i can't get help anywhere else than teamtalk developers, and until now i don't know what to do. Keep in minde i have all of the tolk related files and sources and the tolk vcx proj file and tolk.lib file, but idk where to put them so that my teamtalk5 build steps would also consider tolk and build it as a dependency.

PR #2175, opened 2024-03-02
Android client: added operator password

Muamalaljanahi:


HanaDadfar:
also, android client really, really, really missing the ability to select an input microphone like ios and windows.

PR #2180, opened 2024-03-04
Android client: navigate up button behaviour fix

poretsky:
At first this button should switch to the channels tab. Anyway, it
definitely should not kick user out of the server if some other tab is
selected.


bear101:
I'll add a notice to the ChangeLog

PR #2181, opened 2024-03-04
Redesign hostmanager

CoBC:


HanaDadfar:
@CoBC what was the point of recoding the host manager? what has been added, what was removed and what has been changed? I'm going to build this and try it, but before building, I like to ask you what has been changed as a user facing changes?

CoBC:
Currently we can't build anything from this PR because it's a WIP and 
nothing is working for now. The final point is to simplify hostmanager 
window. I'll explain more later when it will be finish

HanaDadfar:
@CoBC oh, ok. Nice done.
yeah, it's a good idea to simplify host manager.

CoBC:
@bear101 Do you have any advice to complete the GestHostEntry method? Don't know how to do easily, especially for encryptioon properties.

bear101:
> @bear101 Do you have any advice to complete the GestHostEntry method? Don't know how to do easily, especially for encryptioon properties.

I remember the encryption part being very complicated because some settings are only kept temporarily, i.e. until selection changes. Maybe this could be solved by only allowing connect attempts to hosts that are saved.

CoBC:
> > @bear101 Do you have any advice to complete the GestHostEntry method? Don't know how to do easily, especially for encryptioon properties.
> 
> I remember the encryption part being very complicated because some settings are only kept temporarily, i.e. until selection changes. Maybe this could be solved by only allowing connect attempts to hosts that are saved.

The question here is not really for login, but for edition. In the old serverlistdlg.cpp, pressing on "Setup encryption" called GetHostEntry to check validity of selected host and than call encryptionsetupdlg.cpp to directly update host properties. But now, when we will press on "Edit" in serverlistdlg.cpp, serverdlg.cpp will be call to update host properties, and the return will be the GetHostEntry, based on the same behaviour as in mainwindow.cpp for editing a channel for example. But, don't know how to retrieve encryption settings in serverdlg.cpp and than pass it to serverlistdlg.cpp
Not sure if I'm very clear

bear101:
@CoBC sorry I haven't had time to look into this new host manager. The new Android release has taken all my time these last couple of weeks. I was planning to release the Android client as v5.15.2 that only fixed the file upload problem. However, it turned out to cause a cascade of issues that had to be fixed due to new restrictions on Google Play.

The new Android release will be released as v5.16 since we ended up making so many changes. I'll also release the Qt client while I'm at it.

bear101:
@CoBC I've rebased and push a commit that handles encryption dialog. With the setup you're doing now I don't see the need for "m_setup_encryption". The "m_setup_encryption" was only required when user had the ability to change selected server. But this can not be done in your new setup since you're only accessing one server at the time.

CoBC:
@bear101 As you can see in GitHub action, there is a lot of error like 
"usernameBox is not a member of Ui::ServerDlg" E.G. at serverdlg.cpp 
line 41.

Do you have any idea on it?

bear101:
@CoBC I've used Qt Creator's Copy/Paste to add UI items to ServerDlg on this branch: #2249
Note that it's a branch based on master. Try replacing serverdlg.ui with the .ui file you've created on "hostmanager" branch.

CoBC:
Thanks, I'll continue on this new branch

PR #2184, opened 2024-03-05
Android client: file transfer functionality fix

### LINKED ISSUE #2117: can't upload , download and stream Media files
### Description

    hello , when i try to upload or download a file or stream a media files on android it says upload or import files will be Unavailable same thing for streaming , when I click browse it doesn't show the files , I am using the programLatest version from the google play ,and my phone is samsung galaxy s23ultra

### Application

- [ ] qtTeamTalk
- [X] TeamTalkAndroid
- [ ] iTeamTalk
- [ ] TeamTalkClassic
- [ ] TeamTalkServer

### Platform

- [ ] Windows
- [ ] macOS
- [X] Android
- [ ] iOS
- [ ] Linux

### Expected behavior

_No response_

### Actual behavior

_No response_

n/a

poretsky:
This solution is not the perfect one, but it works. To use proper
approach some TeamTalk library functions API modifications are
necessary.

Hopefully fixes #2117

### DIFF
--- a/Client/TeamTalkAndroid/src/main/java/dk/bearware/gui/AbsolutePathHelper.java
+++ b/Client/TeamTalkAndroid/src/main/java/dk/bearware/gui/AbsolutePathHelper.java
@@ -24,7 +24,6 @@
 package dk.bearware.gui;
 
 import android.annotation.SuppressLint;
-import android.content.ContentUris;
 import android.content.Context;
 import android.database.Cursor;
 import android.net.Uri;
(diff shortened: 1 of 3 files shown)


beqabeqa473:
@poretsky do you mean to pass and/or get files as bytes instead of file paths?
If yes, that would be the best way and that would make working with files in more customized way not only for Android.

poretsky:
Accessing files by real path in Android is deprecated in fact. In modern
Android versions even it is not always possible to get real file
path. But there is a legal way to pass an fd (file descriptor) to a low
level function. Thus, it would be the best if such TeamTalkBase methods
as doSendFile() and so on could accept fd as an argument and use it for
local file access. The fd is passed being already opened and detached,
so a low level function must close it itself. This is the legal and
proper approach that has many benefits.

bear101:
Great @poretsky! File upload has been broken for a long time

bear101:
Any chance "Stream media file" can be fixed in the same way? When I try to stream a file it's like it doesn't detect the file path.

Muamalaljanahi:
hi @bear101 It got fixed here. You may need to regrant permissions somehow.

poretsky:
@bear101, please, consider my suggestion concerning file descriptor
usage. This solution works for now, but I think that we cannot rely on
it for a long time, because Android apparently doesn't like file access
by real path. At least, keep it in mind for future.

PR #2190, opened 2024-03-06
Fix compilation on systems using musl libc, like alpine-linux

beqabeqa473:


bear101:
Discussion here: https://github.com/DOCGroup/ACE_TAO/issues/914

PR #2200, opened 2024-03-11
Ask for accessibility settings at first startup on Linux

CoBC:
Based on the work originally did by @Nardol in PR #1865 
Tested on Windows and Linux, works as expected. Should have no effect on MacOS


Client/qtTeamTalk/mainwindow.cpp
```diff
@@ -746,18 +747,26 @@ void MainWindow::loadSettings()
     if(connect_ok)
         QTimer::singleShot(0, this, &MainWindow::slotConnectToLatest);
 
+#if (defined(Q_OS_WINDOWS) && defined(ENABLE_TOLK)) || defined(Q_OS_LINUX)
```
bear101, line 750: Isn't it enough to check for ENABLE_TOLK macro?

Client/qtTeamTalk/mainwindow.cpp
```diff
@@ -771,13 +780,17 @@ void MainWindow::loadSettings()
 
         if(answer.clickedButton() == YesButton)
         {
+#if defined(Q_OS_WINDOWS)
```
bear101, line 783: Also here the macro should be ENABLE_TOLK instead of Q_OS_WINDOWS.

Client/qtTeamTalk/mainwindow.cpp
```diff
@@ -746,18 +747,26 @@ void MainWindow::loadSettings()
     if(connect_ok)
         QTimer::singleShot(0, this, &MainWindow::slotConnectToLatest);
 
+#if (defined(Q_OS_WINDOWS) && defined(ENABLE_TOLK)) || defined(Q_OS_LINUX)
     if (ttSettings->value(SETTINGS_GENERAL_FIRSTSTART, SETTINGS_GENERAL_FIRSTSTART_DEFAULT).toBool())
     {
+    bool SRActive = false;
```
bear101, line 753: The indentation doesn't look right in this if-block

Client/qtTeamTalk/settings.h
```diff
@@ -31,8 +31,10 @@
 #define SETTINGS_VERSION                            "5.3"
 #define SETTINGS_GENERAL_VERSION                    "version"
 #define SETTINGS_GENERAL_VERSION_DEFAULT            SETTINGS_VERSION
+#if defined(Q_OS_WINDOWS) || defined(Q_OS_LINUX)
```
bear101, line 34: Also ENABLE_TOLK instead of Q_OS_WINDOWS

PR #2202, opened 2024-03-12
intercept options for android client

peter1384:


peter1384:
I deleted my teamtalk fork, Then i reforked the teamtalk again, And i only pushed the intercept related things. Is it ok now?

bear101:
Thanks for splitting up the PR. I'll merge it now.

PR #2218, opened 2024-03-18
Android. Now users can also remove channels that they are currently joined in to, Because this is handled by the server

peter1384:


peter1384:
i can't split this commit here because merging it will cause other problems

PR #2223, opened 2024-03-19
Android client foreground service fix

poretsky:
Use more advanced API where it is available.

Please, try it out. Hopefully fixes #2222 problem.


peter1384:
@poretsky currently when we are connected to a server, we always have teamtalk with our channel's name on it in the notification panle, and even clearing the notifications doesn't remove that. And klicking on it brings the teamtalk window. I don't see any usage and any need of that notification always being in there when connected to a server. Can that be removed? the only needed notification that we want teamtalk to send is notification about private messages. So, can that be removed? please really answer my question, because i'm really wundering why that notification is there, and if it has any usage. Thanks!

PR #2226, opened 2024-03-20
Remove android.permission.FOREGROUND_SERVICE

bear101:
… and bump to v5.15.2 build 128

This is a test version to see if users notice a difference when android.permission.FOREGROUND_SERVICE is no longer acquired.

### DIFF
--- a/Client/TeamTalkAndroid/build.gradle
+++ b/Client/TeamTalkAndroid/build.gradle
@@ -31,7 +31,7 @@ android {
         applicationId "dk.bearware.gui"
         minSdkVersion 21
         targetSdkVersion 34
-        versionCode 127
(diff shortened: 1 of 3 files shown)


bear101:
@poretsky I'll try and publish this version and see if users notice a difference in how the notifications are published. When I've been testing I cannot see any difference.

poretsky:
Ok, it seems principally functional. I did not discover negative effects
as well. But it is somewhat incomplete anyway. See #2227.

beqabeqa473:
@bear101 @poretsky i may be wrong but i think this pr broke microphone access in background.
Try to turn on voice transmission and turn off screen. In about five seconds microphone will be turned off.

PR #2234, opened 2024-03-31
Android Studio layout recommendations

bear101:


peter1384:
what's difrent between swich and swich compat?

bear101:
> what's difrent between swich and swich compat?

https://developer.android.com/reference/androidx/appcompat/widget/SwitchCompat

poretsky (approved):
I've added necessary corrections and now it looks ok.

PR #2235, opened 2024-03-31
Android client: foreground service restored

poretsky:
This effectively reverts #2226 since it breaks microphone functionality
when screen is locked.


bear101:
Hehe, this looks very similar to my pull request :) Can you check my comment with the version check

PR #2236, opened 2024-03-31
Reintroduce android.permission.FOREGROUND_SERVICE

bear101:
Without the android.permission.FOREGROUND_SERVICE it's not possible to press the power button and continue to listen to conversations. The connection will be dropped after a minute or so.

### DIFF
--- a/Client/TeamTalkAndroid/src/main/java/dk/bearware/backend/TeamTalkService.java
+++ b/Client/TeamTalkAndroid/src/main/java/dk/bearware/backend/TeamTalkService.java
@@ -33,6 +33,7 @@ import android.content.ComponentName;
 import android.content.Context;
 import android.content.Intent;
 import android.content.SharedPreferences;
+import android.content.pm.ServiceInfo;
 import android.media.AudioManager;
 import android.os.AsyncTask;
 import android.os.Binder;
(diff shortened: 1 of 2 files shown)


bear101:
Following the discussion here: https://github.com/BearWare/TeamTalk5/pull/2226

@poretsky this seems to fix the issue when pressing power button on the Android phone. If this look ok to you then I'll try to explain to Google why we need the android.permission.FOREGROUND_SERVICE. At least I know understand why.

On my Android 11 the connection/microphone is not dropped until after a minute or so. That's why I didn't notice it while testing.

Client/TeamTalkAndroid/src/main/java/dk/bearware/backend/TeamTalkService.java
```diff
@@ -407,9 +408,17 @@ private void displayNotification(boolean enabled) {
                     .setContentText(getNotificationText())
                     .build();
             }
-            notificationManager.notify(UI_WIDGET_TAG, UI_WIDGET_ID, widget);
+            if (Build.VERSION.SDK_INT >= Build.VERSION_CODES.Q) {
```
bear101, line 411: @poretsky Why "Build.VERSION.SDK_INT >= Build.VERSION_CODES.Q" for startForeground and "Build.VERSION.SDK_INT >= Build.VERSION_CODES.N" for stopForeground?

poretsky (commented):
We should initially start foreground mode and then only update the
notification, so it looks somewhat incorrect. And I don't understand
where these version checks come from. They are absolutely waste since
we use ServiceCompat.

I strongly recommend to drop this pr and merge #2235 instead that is
tested already by the way.

PR #2243, opened 2024-04-15
Do not reuse QSoundEffect object for next sound event

### LINKED ISSUE #2220: Sound Event Not Playing Correctly in "Default" Mode
### Description

There's an issue with the playback of sound events in the "Default" mode. The exact behavior is challenging to articulate, but generally, the correct sound isn't always triggered. Instead, the last played sound persists. For example, when User A connects to the server, the appropriate sound plays. However, subsequent events like User B sending a message to the channel result in the repeated playback of the initial connection sound rather than the expected channel message sound. This issue needs attention to ensure accurate sound playback.

### Application

- [X] qtTeamTalk
- [ ] TeamTalkAndroid
- [ ] iTeamTalk
- [ ] TeamTalkClassic
- [ ] TeamTalkServer

### Platform

- [X] Windows
- [ ] macOS
- [ ] Android
- [ ] iOS
- [ ] Linux

### Expected behavior

I expected the sound events in the "Default" mode to play accurately and consistently according to their respective triggers. Specifically, when User A connects to the server, I anticipated hearing the designated sound associated with that event. Subsequently, when other events occur, such as User B sending a message to the channel, I expected to hear the corresponding sound without any interference from previously played sounds. Overall, I expected the sound playback to align precisely with the actions happening within the application, providing clear auditory feedback to users.

### Actual behavior

In the "Default" mode, the sound events are not playing as expected. When User A connects to the server, the designated sound associated with this event plays correctly. However, subsequent events, such as User B sending a message to the channel, do not trigger the intended sound. Instead, the sound that was played during User A's connection persists and is heard again, despite the occurrence of a different event. This results in inconsistent and inaccurate sound playback, which can lead to confusion among users regarding the actual events happening within the application.

I am unsure since which version of TeamTalk this issue has existed, but my current TeamTalk version is 5.15.2.5137 - Unreleased. All that is needed to reproduce the problem is to change the playback mode of the sound event to "Default".

bear101:
This prevents the same sound from being played again and fixes #2220

### DIFF
--- a/Client/qtTeamTalk/utilsound.cpp
+++ b/Client/qtTeamTalk/utilsound.cpp
@@ -582,7 +582,9 @@ void PlaySoundEvent::playDefaultSoundEvent(const QString& filename)
 #if QT_VERSION < QT_VERSION_CHECK(6,0,0)
     QSound::play(filename);
 #else
-    static QSoundEffect* effect = new QSoundEffect(ttSettings);
+    static QSoundEffect* effect = nullptr;
+    delete effect;
+    effect = new QSoundEffect(ttSettings);
     effect->setSource(QUrl::fromLocalFile(filename));
     effect->setVolume(ttSettings->value(SETTINGS_SOUNDEVENT_VOLUME, SETTINGS_SOUNDEVENT_VOLUME_DEFAULT).toInt()/100.0);
     effect->play();


CoBC (approved):

bear101:
@CoBC we could also make a list so QSoundEffect can play until the end. With this chance we cut the sound event short.

CoBC:
It's not already what is done by playback mode one by one?

bear101:
One-by-One uses TeamTalk Sound Device. I've made another pull request with a queued version

lbk2907:
> One-by-One uses TeamTalk Sound Device. I've made another pull request with a queued version

Is it  possible for "Overlapping" be implemented to play using "default playback mode" instead of using "TeamTalk Sound Device" as well?

lbk2907:
> One-by-One uses TeamTalk Sound Device. I've made another pull request with a queued version

@bear101 @CoBC
Is it  possible for "Overlapping" be implemented to play using "default playback mode" instead of using "TeamTalk Sound Device" as well?

bear101:
@lbk2907 for that you can try out this merge request: #2244

lbk2907:
Ok.

PR #2244, opened 2024-04-15
Place QSoundEffect in a queue until playback is finished

### LINKED ISSUE #2220: Sound Event Not Playing Correctly in "Default" Mode

bear101:
This fixes #2220

### DIFF
--- a/Client/qtTeamTalk/utilsound.cpp
+++ b/Client/qtTeamTalk/utilsound.cpp
@@ -582,10 +582,21 @@ void PlaySoundEvent::playDefaultSoundEvent(const QString& filename)
 #if QT_VERSION < QT_VERSION_CHECK(6,0,0)
     QSound::play(filename);
 #else
-    static QSoundEffect* effect = new QSoundEffect(ttSettings);
+    for (qsizetype i=0;i<m_playingQueue.size();)
+    {
+        if (m_playingQueue[i]->isPlaying())
+            ++i;
+        else
+        {
+            delete m_playingQueue[i];
+            m_playingQueue.erase(m_playingQueue.begin()+i);
+        }
+    }
+    QSoundEffect* effect = new QSoundEffect(this);
     effect->setSource(QUrl::fromLocalFile(filename));
     effect->setVolume(ttSettings->value(SETTINGS_SOUNDEVENT_VOLUME, SETTINGS_SOUNDEVENT_VOLUME_DEFAULT).toInt()/100.0);
     effect->play();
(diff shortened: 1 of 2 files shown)


bear101:
@CoBC you can choose which one to merge

CoBC:
I prefer this one but in this case shouldn't remove current one-by-one mode, merge #2243 to fixe #2220 and convert one-by--one mode to this method?

bear101:
Default mode plays using Qt whereas the two other (Overlapping and One-By-One) plays on your selected TeamTalk sound device. I think we should stick to calling this new "Default".

bear101:
The previous behaviour of Default works like #2243

CoBC:
OK but in this case I think we should merge #2243 instead, because some 
users have habit with the way "default" mode play sounds, and can be 
confused if "default" goes to play like "one by one"

PR #2251, opened 2024-04-30
update chinese translation

hejiadong0608:


CoBC:
I just add 2 new string I forgot previously :(
Can you update translation again, so we will merge it with complete work :)

hejiadong0608:
@CoBC  OK. i've finished.

Issue #2252, opened 2024-04-30
qt teamtalk. New host manager's problems


peter1384:
### Description

1: in the context menu of laytist hosts, when i klicked the clear laytist hosts option, my teamtalk hard crashed.
2: the items inside laytist hosts list are just announced as list item for screen readers.
3: When i didn't had any servers saved in my saved servers list, and i klicked the edit option in the context menu, my teamtalk crashed. Context menu options should be grade out when you don't have any server selected.

### Application

- [X] qtTeamTalk
- [ ] TeamTalkAndroid
- [ ] iTeamTalk
- [ ] TeamTalkServer

CoBC:
Point 1 and 3 are fixed by PR #2253  and #2254 
Don't understand second point. List items are correctly labelled with latest host name.

peter1384:
@CoBC point 2: For example, you have 5 servers in the laytist host list. When you scrole throo them with up and down arrow keys, nvda just says list item, list item, list item.

CoBC:
I can't reproduce

PR #2253, opened 2024-04-30
Ensure server is selected to enabled context menu actions

CoBC:
#2252

### DIFF
--- a/Client/qtTeamTalk/serverlistdlg.cpp
+++ b/Client/qtTeamTalk/serverlistdlg.cpp
@@ -733,11 +733,11 @@ void ServerListDlg::slotTreeContextMenu(const QPoint& /*point*/)
     QAction* genTTServ = menu.addAction(tr("&Generate .tt file"));
     QAction* publishServ = menu.addAction(tr("&Publish Publicly"));
     auto srcIndex = m_proxyModel->mapToSource(ui.serverTreeView->currentIndex());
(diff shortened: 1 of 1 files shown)


peter1384:
@CoBC also, please move the connect button to the context menu, because the connect is related to specific server. Only add new server button should be stayed in the main dialog as a button because that button is not a server or anything related button.

Client/qtTeamTalk/serverlistdlg.cpp
```diff
@@ -733,11 +733,11 @@ void ServerListDlg::slotTreeContextMenu(const QPoint& /*point*/)
     QAction* genTTServ = menu.addAction(tr("&Generate .tt file"));
     QAction* publishServ = menu.addAction(tr("&Publish Publicly"));
     auto srcIndex = m_proxyModel->mapToSource(ui.serverTreeView->currentIndex());
-    if (srcIndex.isValid())
-    {
-        delServ->setEnabled(m_model->getServers()[srcIndex.row()].srvtype == SERVERTYPE_LOCAL);
-        publishServ->setEnabled(m_model->getServers()[srcIndex.row()].srvtype == SERVERTYPE_LOCAL);
-    }
+        delServ->setEnabled(srcIndex.isValid() && m_model->getServers()[srcIndex.row()].srvtype == SERVERTYPE_LOCAL);
```
bear101, line 736: This doesn't look right. Is there an 'if' missing?

bear101: ah, I was confused by the indentation. I'll fix it

PR #2255, opened 2024-04-30
people who have the user can make others channel operator right will no longer be asked to enter the operator password in auder to grant/revoke channel operator

peter1384:
In teamtalk classic, when you have the user can make other users channel operator right enabled, you don't need to enter the password in auder to grant/revoke operator, because you have the access to it even if you don't know the operator password, and even if you enter the password incorrectly or don't enter anything, it would work for you, so just used the function in android client as wel.

### DIFF
--- a/Client/TeamTalkAndroid/src/main/java/dk/bearware/gui/MainActivity.java
+++ b/Client/TeamTalkAndroid/src/main/java/dk/bearware/gui/MainActivity.java
@@ -1434,6 +1434,8 @@ private EditText newmsg;
 
     @Override
     public boolean onMenuItemClick(MenuItem item) {
+        UserAccount myuseraccount = new UserAccount();
+        ttclient.getMyUserAccount(myuseraccount);
         AlertDialog.Builder alert = new AlertDialog.Builder(this);
         switch (item.getItemId()) {
         case R.id.action_banchan:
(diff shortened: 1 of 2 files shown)


peter1384:
@bear101 this pull request is ready to be merged.

Client/TeamTalkAndroid/src/main/java/dk/bearware/gui/MainActivity.java
```diff
@@ -1434,6 +1434,8 @@ public boolean onItemLongClick(AdapterView< ? > l, View v, int position, long id
 
     @Override
     public boolean onMenuItemClick(MenuItem item) {
+        UserAccount myuseraccount = new UserAccount();
```
bear101, line 1437: should be moved to `case R.id.action_makeop:`

peter1384:
@bear101 i didn't move this to the case makeop because this way we can retreave account rights in all cases if we need, raather than only case makeop. This was my reason but if you stil want, i can move it to there.

peter1384:
@bear101 so what's up? do you expect anything else from me in this pr?

peter1384:
@bear101 this pull request is ready, but you've marked it as draft. Can i know the reason please? Thanks

bear101:
Please  make the changes I wrote earlier.

peter1384:
@bear101 done, but as I said before, i'm not really a big fan of this way, because this way if we need to check some user rights for other cases, we have to right this for all of them, but that way we only needed to say which right we want to retreave.

peter1384:
@bear101 Can you please check up in this pr and merge it if possible, because my fork of this repozitory is messed up and i need to delete my fork and fork this repozitory again,  but i cannot do that as long as this pr is open because the branch of this pr is coming from my fork.
Thanks

peter1384:
@bear101 Thanks!
I have a question
Now that i'm also a contributor of teamtalk5, can I create a new branch in the teamtalk5 repozitory it self and open my pull requests instead of doing it in my own fork?

bear101:
So far you've added 131 lines. It is not enough to become a contributor. When you've added a couple of thousand lines of code then you become a contributor.

PR #2260, opened 2024-05-03
Ability to customize event messages

CoBC:
Add a new tab in preferences to customize status and TTS messages for events.


bear101:
@CoBC The label in the tab takes up too much space. Shouldn't we just have the text in the edit box and reset it to default if it's blank?

CoBC:
Don't understand which label takes too much place? Which label do you 
want to remove?

bear101:
The "Available variables..." label in "Events Tab"

bear101:
Alternatively these Event messages could show up in a QInputDialog when the Text-To-Speech message is checked in "Text to Speech" tab.

bear101:
If you later want to have more customizable status messages then there's no more space.

CoBC:
Yes, I also thought of the alternative option to customize message when 
checking each event in TTS tab, but what about message use in TTS and 
status (chat history)?

bear101:
Status messages already have a separate dialog in Display->Show statusbar events. So their customisation could happen there.

CoBC:
OK, so I'll try and make another PR more later. I'll close it for now as 
the current implementation is not retained

CoBC:
No problem :), we can try and find a better solution :)

peter1384:
@CoBC anyway, i don't think customising these events would have a lot of points and usage, the default customisation is fine i guess, so your time could be spent to add other useful things instead.

PR #2269, opened 2024-05-08
added connect action to the host manager's context menus

peter1384:
tested them and made sure there is no problem and they function properly


peter1384:
@CoBC ok. Now that the connect action is in the context menu and our point is to simplify the host manager and clean it so it has everything in their correct places, I can do 2 options in a new pull request. I'll provide the options, and please you tell me what do you think we should do.
1: Just remove the connect button from the main screen, and users will be able to connect by just pressing enter or use the context menu if they want to, (This is recommended by me),2: I'll open a pull request and fix this button to get it in to working because as you even said, activating this button does nothing.
so we should remove this button or fix it to work, but as i said before, having it in there is pointless.
What do you think We should do?

PR #2273, opened 2024-05-11
Use locale file size in "File transfer" dialog

CoBC:


Client/qtTeamTalk/utilui.h
```diff
@@ -148,4 +148,5 @@ bool restoreWindowPosition(const QString& setting, QWidget* widget);
 QStringList extractLanguages();
 bool switchLanguage(const QString& language);
 QString getFormattedDateTime(QString originalDateTime, QString format);
+QString getFormattedFileSize(int filesize);
```
bear101, line 151: This should be an int64

PR #2274, opened 2024-05-12
Implement "Move Users" dialog

### LINKED ISSUE #1833: Add Move users dialog to Qt client
### Description

Currently, the only way to move users in Qt client is to select users on channel tree than move it to channel we wish.
In classic client, a dialog was present to allow moving users quickly. A similar dialog should be added to Qt client with a list of checkboxes to choose users to move, and a combobox to choose channel.

### Application

- [X] qtTeamTalk
- [ ] TeamTalkAndroid
- [ ] iTeamTalk
- [ ] TeamTalkServer

CoBC:
Fixes #1833
@bear101 Maybe you prefer to have an ui file. If yes, can you create it please? :)

### DIFF
--- a/Client/qtTeamTalk/moveusersdlg.cpp
+++ b/Client/qtTeamTalk/moveusersdlg.cpp
@@ -1,0 +1,100 @@
+/*
+ * Copyright (C) 2023, Bjørn D. Rasmussen, BearWare.dk
+ *
+ * This program is free software: you can redistribute it and/or modify
+ * it under the terms of the GNU General Public License as published by
+ * the Free Software Foundation, either version 3 of the License, or
+ * (at your option) any later version.
+ *
+ * This program is distributed in the hope that it will be useful,
+ * but WITHOUT ANY WARRANTY; without even the implied warranty of
+ * MERCHANTABILITY or FITNESS FOR A PARTICULAR PURPOSE.  See the
+ * GNU General Public License for more details.
+ *
+ * You should have received a copy of the GNU General Public License
+ * along with this program.  If not, see <https://www.gnu.org/licenses/>.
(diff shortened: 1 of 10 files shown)


Client/qtTeamTalk/moveusersdlg.cpp
```diff
@@ -0,0 +1,100 @@
...
+
+void MoveUsersDlg::setupUi()
+{
+    setWindowTitle(tr("Move Users"));
+    setWindowIcon(QIcon(APPICON));
+
+    QVBoxLayout* mainLayout = new QVBoxLayout(this);
+
+    QGroupBox* usersGroup = new QGroupBox(tr("Users"), this);
+    QVBoxLayout* usersLayout = new QVBoxLayout(usersGroup);
```
bear101, line 40: Clang-Tidy says "usersLayout" may memory leak. I just can't figure out why. Can you? It appears to be the case if userIds.size() is 0. However, if I put a check for userIds.size()>0 then it still complains :(

bear101 (commented):
Btw, I've added one commit

Client/qtTeamTalk/moveusersdlg.cpp
CoBC: Maybe just add a QLabel if userids is empty to indicates there is no user available. This QLabel will never be visible but maybe clang-tidy will be happy. Not sure :)

bear101: Hm, that also didn't work (with the QLabel). I'll merge. It looks like a false-positive.

bear101:
btw, .ui files are not good for dynamic content.

PR #2275, opened 2024-05-13
Ability to join latest joined channel

### LINKED ISSUE #1941: qt client: add the last joind channel check box to the connect to a server dialog like android client.
### Description

in android client, in the connect to a server dialog, we have a check box called last joined channel: when this option is checked, in that server any channel you join itt will remember the path and the password, and next time you log in it will lead you directly to that channel, wich is a useful feature for me a lot: please also add this ability to the qt client, thanks a lot.

### Application

- [X] qtTeamTalk
- [ ] TeamTalkAndroid
- [ ] iTeamTalk
- [ ] TeamTalkServer

CoBC:
Fixes #1941

### DIFF
--- a/Client/qtTeamTalk/mainwindow.h
+++ b/Client/qtTeamTalk/mainwindow.h
@@ -260,6 +260,7 @@ private:
     void relayAudioStream(int userid, StreamType st, bool enable);
     void enableVoiceActivation(bool checked, SoundEvent on = SOUNDEVENT_VOICEACTON, SoundEvent off = SOUNDEVENT_VOICEACTOFF);
     void updateClassroomChannel(const Channel& oldchan, const Channel& newchan);
+    void moveUsersToChannel(int chanid);
 #if defined(Q_OS_LINUX)
     void executeDesktopInput(const DesktopInput& input);
 #endif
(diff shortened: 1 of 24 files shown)


Client/qtTeamTalk/mainwindow.h
```diff
@@ -154,6 +154,7 @@ class MainWindow : public QMainWindow
 
     //current host
     HostEntry m_host;
+    bool m_latesthost = false;
```
bear101, line 157: shouldn't we put 'm_latesthost' into the HostEntry-struct? It could also simply other functions (like setServerEntry, deleteServerEntry)

CoBC: I didn't see a real utility to put this in HostEntry struct, we already have all we need to compare latesthost and saved host in common.cpp. Also if we add it to HostEntry struct we probably have to store it in config file, so update config at first startup of this version. And, I'm not sure it's really useful to store this in config as we already have different sections for latest host and server entries.
So, as you want :), I can move it to HostEntry struct if you find it useful, but I think we can keep as currently.

safe-acid: ok

peter1384:
@CoBC the boxes should not be cleared when the checkbox is unchecked because what if someone wants to keep the current path in the boxes but they just want to uncheck the checkbox so the path doesn't get any update anymore?

PR #2279, opened 2024-05-14
Move latesthost to HostEntry-stuct

bear101:


bear101:
@CoBC I was not a big fan of additional member variables in MainWindow-class, so I've move the latesthost to HostEntry and changed a bit how a hostentry gets this information.

CoBC (approved):

Issue #2131, opened 2023-12-20
bug in qt teamtalk. Moving users between channels is not possible using the Move User(s)  Ctrl+Alt+V context menu option


HanaDadfar:
### Description

currently, it is possible to do every action throo the option, and also shortcut
but this is not the case for moving users between channels, because the Move User(s)  Ctrl+Alt+V option is located in users context menu > advance sub menu, But this option should be located in channels context menu. Then, it will be possible to move users between channels using the context menu options and not only forced to use the shortcuts, here's the possibility.
1: select all users you wish to move, then go to users context menu>advance sub menu, and klick on Store User(s) for Move  Ctrl+Alt+X option
2: focuse on destination channel, go to channels context menu, and klick on Move User(s)  Ctrl+Alt+V option to finish the moving action.
but currently this possibility doesn't work because both store user for move and move user options are in the users context menu.
store for move option should be in users context menu which already is so no touching needed, and move users option should be moved to channels context menu.
doing this will solv this problem and enables the possibility explaned abuv, there for enabling the possibility for users to move users using the context menu options as whel, not only limited to shortcut
also this is a Technical bug, because the Move User(s)  Ctrl+Alt+V option is in the users context menu, but its shortcut which is ctrl+alt+v works on the channels.
affect version/s: 5.15 and older versions

### Application

- [X] qtTeamTalk
- [ ] TeamTalkAndroid
- [ ] iTeamTalk
- [ ] TeamTalkServer

PR #2280, opened 2024-05-14
moved the move user option to the channels menu which is its proper place

peter1384:
some users and also me have reported that we can only use ctrl+alt+v shortcut to do the move action, but we can't use the option which this shortcut is assighned for it, because this option is not in its proper place. For the more, this option was in the users menu, but we needed to focus on a channel when we wanted to press ctrl+alt+v, so now that this option has been moved to channels menu, we can press enter on it and complete the move action

### DIFF
--- a/ChangeLog.txt
+++ b/ChangeLog.txt
@@ -11,6 +11,7 @@ Default Qt Client
 - "Check for Update" in "Help" menu
 - Ability to join Last Joined Channel for each server individually
 - Use Qt checkable flag in user rights list to allow screenreader to see if an item is checked
+- Moved "Move User" option from users menu to channels menu, to allow the option usage as wel, raather than only shortcut usage
 Android Client
 -
 iOS Client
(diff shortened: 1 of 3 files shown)


peter1384:
this pull request adresses this issue
https://github.com/BearWare/TeamTalk5/issues/2131

CoBC:
I'm not really in favor of it, because this action affects users, also 
it can be strange to have to switch from a menu to another one to select 
and than move users.

peter1384:
@CoBC 1: How this action will effect users? pressing ctrl+alt+v shortcut stil works like always.
2: yes people are gonna rarely use the option and swich menus to move users, but the correct place for this option is channels menu because the job of this action aplys to channel, not user
3: look at teamtalk android as an example. The "Select For Move" option is in the users popup menu when you long press a user, and the "Move Hear" option is in the channels popup menu when you long press a channel. So, this is the same hear now. You select a user by focusing on a user and pressing the "Store For Move" option to store that user, and then you focus on a channel and press the "Move Users" option, to move the selected users to a channel.

peter1384:
actually with out this pull request some people inclooding my self was in troble sometimes because sometimes i'm unable to use shortcuts for some reason, but with this pull request there is no problem and this pull request won't interupt other users who were using the shortcuts, and so we needed this option.

bear101:
I don't see how this change improves usability

PR #2282, opened 2024-05-15
Customizable TTS messages

CoBC:
@bear101 Can you take a look and see if it's OK for you?
If it's OK, I'll continue to add more events and maybe add it to status messages also.


CoBC:
@jeremyp3 As I know, you're interested by this feature, so, you can also test and see what you think.

CoBC:
@bear101 After all these new commits, can you make a first review? If changes are needed, I'll make changes than I'll do the same for status events.

bear101:
If I put in "foo" as the event message then how do I revert to the original message? I think a "revert" feature would be nice if you make a mistake

CoBC:
Maybe just keep edit field blank so the default message will be use, 
right? Otherwise should I add a "Revert" button to the custom input dialog?

bear101:
I think clearing the message should revert. Maybe put this information in the input dialog’s description. On 16 May 2024, at 22.07, Corentin Bacqué-Cazenave ***@***.***> wrote:﻿
Maybe just keep edit field blank so the default message will be use, 
right? Otherwise should I add a "Revert" button to the custom input dialog?

CoBC:
Hmmmmm, it's already the case, if editfield is blank, so the setting value will be empty and default value will be use.
So, should we precise it to the input dialog's description, or do I missed understood something?

bear101:
Try putting in "foo" as "User logged in (Customizable). Then it's not possible to get back the original message. If you put in a blank message then the message will still be "foo".

bear101:
Can't we remove the "(Customizable)" postfix? It's already quite clear which are customizable (the items with text in the second column).

bear101:
@CoBC you have more for this one? Otherwise I'll mark as ready and squash

CoBC:
I'll do the same for status events. Do you prefer another PR?

bear101:
Yes. Make another PROn 18 May 2024, at 13.57, Corentin Bacqué-Cazenave ***@***.***> wrote:﻿
I'll do the same for status events. Do you prefer another PR?

CoBC:
OK, so you can merge this one and I'll make another one from master 
after this

ChrisDuffley:
@CoBC: When customizing TTS message at least, and resetting to the default value, it stays at the message that was previously customized rather than keeping it at its default as you shift+tab to return to the table to look at more messages. Using latest commit from #2324.

ChrisDuffley:
I have not done status message customizations as of yet but have experimented with this and found this issue certainly in Text-To-Speech preferences.

CoBC:
Thanks a lot, status bar has the same issue. Fix by PR #2325

PR #2284, opened 2024-05-15
Display current message in extra column

bear101:
@CoBC change ttseventsmodel.cpp:257


CoBC:
@bear101 Done. Is it OK for you?

PR #2286, opened 2024-05-17
Redesign user accounts dialog

CoBC:
Suggested by @peter1384 in #2285


peter1384:
@CoBC Thanks! is this complete already? can I build? LOL

CoBC:
It's ready and working, probably just the GUI is visually broken, but you can build, try, and see what you think

peter1384:
@CoBC really nice! everything is as expected so far, but, there is something that needs to be done as wel in this case also.
when you set the abuce prevention to custom, as you know, a diallog will popup, allowing you to configger the number of commands to allow, and number of seconds.
but currently, when you set one of the spin buttons, you have to press the ok button, then the second dialog will come up, allowing you to set the second spin button. Also, the spin buttons are unlabled, and screen reader just says spin button 0 or whatever.
These dialogs should be simplifyed like the following.
when you set the abuce prevention to custom, a dialog should pop up, contaning 2 spin buttons, with an ok and a cansle button, also the spin buttons should be labled, one as number of commands, and one for the seconds to allow respectivly, as you know, raather than 2 dialogs with a single spin button in each one.

CoBC:
@peter1384 Please test again with last commit

peter1384:
@CoBC build started. Configuration: releace x64

peter1384:
@CoBC yes, it has been made as expected, but 2 small problems
1: when the dialog opens, any item inside the dialog doesn't get in focus of the screenreader, which the first spin button should become in focuse
2: the spin buttons has got discription lable instead of name lable. When you navigate between spin buttons,  screenreader says spin button commands 60, and when navigating up and down, it says command 59, command 60, etc, which only numbers should be announced when going up and down and changing spin button. If you don't know how to fix the lables of these spin buttons, look at users spin button of filter servers groupbox of host manager, or look at the server prop's spin buttons.

CoBC:
1: Which Qt version are you using?

2: I know hot to fix it, thanks, but I just did it differently as usual, 
because I use prefix and suffix properties of spinbox, so screenreader 
read entire values as each step change.

peter1384:
@CoBC i'm using qt 6.7.0
also, the dialog which opens when you want to add a new user or edit a user or view a user, doesn't focuse anything inside the dialog, just opens the dialog. But i don't hacve this focus problem in any other dialog, for example settings or host manager's add a server dialog.
2: ok i got your point, but this way is not good at all when you want to change the numbers to so faar away, such as from 0 to 200 or 300, then you'd everytime hear command 30, command 60, command 20, command 172, etc.

CoBC:
1: It's just an issue with Qt6.7, focus moves to "OK" button by default, 
sometimes NVDA announces it, sometimes not.

2: If you want to set values to 200, it's probably better to type 200 on 
your keyboard instead of pressing arrow up 199 times.

Btw: no need to mention me on each comment, I received all notifications 
from this repo

peter1384:
@CoBC ok then this works fine in these cases, but again 2 things.
1: i understood that you want the screenreader to announce the discription up on step change of the spin button, but also, please add a short title for the spin buttons as wel, so the screenreader doesn't say spin button at the very first, and also people who are scroling the dialog with tab key, can hear the spin button title instead of spin button at the first, hear is an example.
for the first spin box, when we focuse on it, instead of hearing something like this.
spin button  Command Limit: 60
we should hear
command spin button  Command Limit: 60
or any other title that you want, just anything short is better than nothing
2: you should change the logik of how this dialog opens. Currently, the only way to open this dialog is to press up arrow once and a down arrow once to reach the custom specifyed, which is not a problem really, the problem is that wen you set a custom limit and next time you open this dialog and you want to look at what is your custom command limit and your inteval to edit it or something, you can't. When you press up arrow and down arrow once to open the dialog, the command limit and inteval will be both set to 60, because that one is the last one that was provided.
so, please add a new button called set button or something, which only appears when the custom specifyed option is being selected in the combo box, which this set button will always open the dialog.
Thanks.

peter1384:
oh sorry to mention you again, i was just used to it.

peter1384:
I mean, when you press up arrow and down arrow, the dialog should open like before stil and this logic shouldn't be touched, just a set button should be added beside that combo box when the custom is selected, to allow the dialog open with out missing the custom limit commands settings. So this button would be really needed, useful and apreciated, Thanks
And the last but optional, a close button should be added beside new user button, to close the user account dialog.
the rest is fine! Thanks for everything!

peter1384:
oh! forgot something! Also context menu actions should only be enabled if an item is selected in the list like the server list, to prevent functions on non selection and stuf.

peter1384:
it is now like this
Limit issued commands  combo box  Custom (25 commands per 60 seconds)  collapsed  Down
but, there is no set button besides it to open the dialog and edit it, raather than pressing up and down arrow to open the dialog but mis the settings  which is 25 commands per 60 seconds in this case.

CoBC:
Don't understand why you want a button to open the dialog to see a 
setting you can already see in the combobox, if you open dialog to 
change setting, I don't understand the issue with set setting from a 
previous value.

peter1384:
there is no issue to change settings from the previous value, but just this is more quicker and the standard habit. But, you can do as how you wish. Thanks
Just i was thinking that it is strange that you can open a dialog by pressing up arrow and down arrow, also keep in minde that the dialog will not open at all when you press space in the combo box to expand the combo box and then press up arrow and down arrow at all, so a set button to open this dialog is needed anyway. I hope you've got my point correctly.

bear101:
@CoBC you have more for this one? Otherwise I'll squash and merge

peter1384:
also, when in the accounts tree view and selected an account, pressing enter should open the edit for that account, but currently enter key always activates new account button, and only space key activates edit account when in the tree view. Enter key also should activate edit account when selected an account in the list, and should activate the new account button only when focused on it.

CoBC:
You can merge :)

peter1384:
oh! an important thing. When it's the case to only view user information, instead of having ok and cansle buttons in the dialog, like the view server information dialog, we should only have a close button in the dialog.

CoBC:
I commited new changes in reply to Peter's comments.
Hopefully it's ready this time

peter1384:
very nice, and good job. Now i actually like the qt teamtalk because the things are no longer messy so much like classic client. My last request to this pr is to add a close button beside the new user account button, to close the dialog like other dialogs such as the old user accounts system, private messages dialogs, etc.
Thanks!

CoBC:
You can just press escape or use system menu to close the dialog, there 
is no "Close" button in host manager for example, so not sure this is 
really required

peter1384:
yes there is no close button for host manager, but some other dialogs have this button in teamtalk, and some dialogs not. The standard ui is to have a close button for every dialog for example look at nvda's dialogs, configuration profiles dialog, addon store dialog, my opinion is add it to this dialog, and then add it to all other dialogs which do not have this button, which are not so much, i think only online users dialog doesn't have this dialog, and the host manager. Other dialogs either have an ok and a cansle button, or a close button which are fine. So, i say we add it.

CoBC:
In this case, maybe we'll add it to all other dialogs, but in another PR 
more later

peter1384:
i say you add it to this dialog in this pr because it's related to this pr, and later you add to other dialogs in another pr

PR #2288, opened 2024-05-18
Change UI to customize TTS message

CoBC:
I'm really sorry, after saying it was ready this morning, I finally found an even more practical way for customization, instead of having a dialog that opens for each checked event, a field changes dynamically for personalization.
If this new interface is OK, I will personalize status messages in the same way


CoBC:
@bear101 Do we still need column to display current message in model as we can now see message directly in edit field, or should I finally remove them from TTS model and status bar model in my next PR?

jeremyp3:
I think it's good that the message is also in the list, so you don't have to  tab/shift tab to see the message for each event.

CoBC:
Ah yes, you're right. We'll keep it

PR #2291, opened 2024-05-19
Ability to customize status bar messages

CoBC:


Client/qtTeamTalk/statusbardlg.cpp
```diff
@@ -51,9 +61,100 @@ void StatusBarDlg::slotStatusBarEventToggled(const QModelIndex &index)
...
+    auto eventMap = UtilUI::eventToSettingMap();
+    StatusBarEvents eventId = static_cast<StatusBarEvents>(index.internalId());
+
+    if (eventMap.contains(eventId))
+    {
+        const StatusBarEventInfo& eventInfo = eventMap[eventId];
+        QString paramKey = eventInfo.settingKey;
+        QString defaultValue = UtilUI::getDefaultValue(paramKey);
+        QString currentMessage = ttSettings->value(paramKey, defaultValue).toString();
+        if (eventInfo.eventName.size() > 0)
```
bear101, line 92: When "subscription change" and "classroom change" events are called then this label is set to the event message. Why this behaviour? Also the label becomes too long. Shouldn't this just be removed?

CoBC: I think it's more clear to understand that the message for each subscription change or classroom is the same, without this behaviour I didn't find really clear the fact that each subscription or each classroom change available in list don't have his own message.
Maybe we can just choose a shorter event name?
We also have the same behaviour in TTS tab so if a change is needed we'll have to do the same in TTS tab also

bear101: The problem is that the "Message for Event" stays. There should be an "else" to reset to default behaviour.

CoBC:
Thanks for the report, fixed by commit edbbcf0

I don't understand your idea, do you mean an extra TTS event in which 
all status messages will be read? Or can you explain more?

PR #2292, opened 2024-05-19
Redesign "Sound events" tab

CoBC:


peter1384:
please also fix the groupe boxes. The edit box for the file edit, and the brouse button are in the enable/disable sound events groupebox which are should be moved out of there.

CoBC:
It seams to still have something wrong, when you go to "Sound events" tab, all default values are reset, but this should be done only if pack sound change. Don't understand why for now :(

CoBC:
Fixed. So the major part of Sound Events tab is now saved dynamically.
Maybe one day we can move all preferences to dynamic saving, like a lot of other apps now

peter1384:
not sure i've correctly understood what does dynamic saving means. In the connect to server dialog, the "Include servers  checkbox" and users spin button are saved dynamicly for example? because when you set them and close the dialog and the app and next time you open the app and this dialog, they are as you modifyed it last time? or dynamic saving is something else?

CoBC:
Yes, it's exactly it, but, in preferences dialog, we have some options 
with dynamic saving (like new TTS customization or this new Sound events 
tab) and other options in manual saving (when pressing OK)

peter1384:
oh cool! but then as i understand, some options cannot be saved dynamicly, for example, in the general tab, nickname wrighting cannot be saved dynamicly because for example as you are changing your nickname, any leter you type in your keyboard it will constantly send event to the server with your new nickname just that one more leter has been added or removed lol. Is it true?

CoBC:
If we move to dynamic saving for all settings, we'll have to do it 
correctly, so, for example, don't save nickname each time you type a 
character but only when you leave edit field

peter1384:
then if you change your nickname and leave the text feald but you press the cansle button, what will happen?

CoBC:
If we switch to dynamic saving for all srttings ehre will be no OK and 
Canbel buttons anymore but just a Close button.

This will require a discussion more later, it's initially not the topic 
of this PR

bear101:
Good idea @CoBC. This makes it much easier to add new events.

bear101:
@CoBC do you agree with commit a49f6edb623617d0385869a8116bfda271647f47 ? I found it quite confusing that I could not go back to the default audio file.

PR #2296, opened 2024-05-21
Ability to search in "Banned Users"

CoBC:
Don't understand why it's crashing when searching something which don't exist :(
@bear101 Do you have an idea?

### DIFF
--- a/Client/qtTeamTalk/bannedusersdlg.cpp
+++ b/Client/qtTeamTalk/bannedusersdlg.cpp
@@ -21,6 +21,7 @@
 #include "utilui.h"
 
 #include <QPushButton>
+#include <QRegularExpression>
 #include <set>
 
 extern QSettings* ttSettings;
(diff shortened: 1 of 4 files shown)


peter1384:
would you just remake banned users dialog, so instead of having 2 tree views one for unbanned users and one for banned users, we only have a banned users treeview, and if we want to unban a user, we can press application on a banned user and remove it.
or, please do something.
add context menu to both tree views banned users and unbanned users, and that, is, we can move a user from one tree view to another. Example.
when selected a user in banned users tree view and pressed application, we have an option, called move this user to unbanned tree view.
And the same for unbanned tree view. That is,  When selected a user in unbanned tree view and pressed application, we have an option called move this user to banned users tree view, so, these context menu actions would do the work of these 2 buttons.
>  button  Remove banned user
<  button  Put back to banned users
Because these buttons are confusing, I normally get confused when for example i want to unban a user what to do and which user i'm unbanning, but context menu would fix this for me.
I wish you do this second one because then we stil have the unbanned users tree view and if we unbanned a user by mistake, we stil have that user in the unbanned users tree view and we can move it back to banned users tree view.

peter1384:
@CoBC I can do the context menus. Do you wish to do it or can I do it?

CoBC:
It's not the goal of this PR, which is still in draft as you can see. 
So, I'll do it but more later.

peter1384:
ok then you can do it, I knoe you would surely know this, but i'll just give you a hint so you don't end up bothering yourself so much than you have to
When you added the context menus to both tree views and added the actions to it, you don't need to write any void for the context menu actions as we already have the voids which we want these context menu actions to execute, just assighn "slotBanUser" and "slotUnbanUser" to the actions.

Client/qtTeamTalk/bannedusersdlg.cpp
```diff
@@ -214,6 +246,15 @@ BannedUsersDlg::BannedUsersDlg(const bannedusers_t& bannedusers, const QString&
...
 
     ui.bannedTreeView->header()->restoreState(ttSettings->value(SETTINGS_DISPLAY_BANNEDUSERS_HEADERSIZES).toByteArray());
+    connect(ui.filterButton, &QPushButton::clicked, this, &BannedUsersDlg::filterChanged);
+    ui.bannedTreeView->setFocus();
+}
+
+void BannedUsersDlg::filterChanged()
+{
+    QString filterText = ui.filterEdit->text();
+    static_cast<BannedUsersFilterProxyModel*>(m_bannedproxy)->setFilterText(filterText);
```
bear101, line 256: I don't see why this cast is necessary

bear101:
> Don't understand why it's crashing when searching something which don't exist :( @bear101 Do you have an idea?

I think I've fixed it.

peter1384:
and when you added the context menus, like other context menus of other tree view places, also add Sort By options, to be able to sort banned users as wel

PR #2298, opened 2024-05-21
Confirm server deletion on iOS

CoBC:
It seams to build correctly according to GitHub action checks


bear101:
The CI pipeline does not compile iTeamTalk. It only builds libraries.

CoBC:
Oh OK, I didn't remember it.
So if this PR doesn't compile, I guess we can just close it, as I can't test changes myself.

PR #2301, opened 2024-05-21
Optional sound event for user typing private message

### LINKED ISSUE #1121: Play continuously a sound when user is typing a private message
When the person we privatly chat with is typing, there is an indication.
Could we also have an option to play a sound all the time the user is typing so playing it in loop?

CoBC:
Fixes #1121


bear101:
@CoBC I've noticed a bug in Sound Events tab:

1. Select an item in the list
2. Change sound pack
3. Change selection in the list
4. Now the sound event of the previous sound pack replaces the sound event

CoBC:
Fixed

PR #2303, opened 2024-05-22
Only use QLabel in User info dialog

CoBC:
Maybe usage of readonly QLineEdit was wanted, but I didn't see the real advantage of it.
So, I just replaced it by static QLabel

### DIFF
--- a/ChangeLog.txt
+++ b/ChangeLog.txt
@@ -16,6 +16,8 @@ Default Qt Client
 - Redesign "Sound Events" tab in Preferences
 - Ability to replace existing file on upload
 - Menu to insert MOTD variables in "Server Properties" dialog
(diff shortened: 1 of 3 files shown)


CoBC:
We can use arrow keysto navigate in a QLabel, but it's not read by 
screenreader, right. It's an issue which have to be reported to Qt.

However, note that we can use screenreader option to read character by 
character, E.G. with NVDA, numpad 3 and numpad 1.

bear101:
The reason for QLineEdit was to be able to copy/paste user information. But it looks like it also works with QLabel.

CoBC:
Yes it works, I find it more appropriate when we just display info

jeremyp3:
@bear101 I find that this pr degrades the accessibility of this dialog box, as indicated in my first message, I don't understand the merge.

PR #2304, opened 2024-05-22
Support Import/Export entire Server List

CoBC:


Client/qtTeamTalk/serverlistdlg.cpp
```diff
@@ -397,36 +399,45 @@ void ServerListDlg::slotImportTTFile()
...
             tr("Failed to load file %1").arg(filepath));
         return;
     }
 
-    QDomElement rootElement(doc.documentElement());
-    QString version = rootElement.attribute("version");
-
-    if(!versionSameOrLater(version, TTFILE_VERSION))
+    QDomElement rootElement = doc.documentElement();
+    if (rootElement.tagName() == "teamtalk")
```
bear101, line 411: Use TTFILE_ROOT instead of "teamtalk"

Client/qtTeamTalk/serverlistdlg.cpp
```diff
@@ -619,6 +630,110 @@ void ServerListDlg::saveTTFile()
...
+    if (filename.isEmpty())
+        return;
+
+    ttSettings->setValue(SETTINGS_LAST_DIRECTORY, QFileInfo(filename).absolutePath());
+
+    QDomDocument doc;
+    QDomProcessingInstruction xmlDecl = doc.createProcessingInstruction("xml", "version='1.0' encoding='UTF-8'");
+    doc.appendChild(xmlDecl);
+
+    QDomElement root = doc.createElement("teamtalk");
```
bear101, line 677: Use TTFILE_ROOT instead of "teamtalk".

Client/qtTeamTalk/serverlistdlg.cpp
```diff
@@ -619,6 +630,110 @@ void ServerListDlg::saveTTFile()
...
+    root.setAttribute("version", "5.0");
```
bear101, line 678: Use TTFILE_VERSION instead of "5.0"

peter1384:
also, the way we choose the single server per file or export all servers in one file is a bit strange I think.
When we pressed the export server list to .tt file button in the host manager, we should get a combo box and an ok button and a cansle button, like the ban a user dialog. In the combo box then we select we want a single file or one file per each server. The rest is grate!

CoBC:
I don't really like multiplication of dialogs with just one widget, it's 
why I choose this method with menu

peter1384:
oh! I looked at this properly and now i know This is a context menu, grate grate, don't touch it it's fine, You are grate to go!

peter1384:
there is a bug with export all servers in one file. When you klick on it, in the save file dialog wich opens and allowes you to save the tt file, the file type is set to tt file *.tt which is correct, but the filename edit feald is blank, so when we try to press the save button to save the tt file in to the selected directory, the save button doesn't do anything and doesn't save anything because the filename is blank, so this means this feature is broken with this bug.

peter1384:
it should work as when you select a file in the files tab and you download it, in the save file dialog that opens, the file name that you are downloading is in the filename edit feald so you can continue

peter1384:
is this random number in the filename is expected? edit  Alt+  n  selected TeamTalk5Servers_20240523_121206.tt

CoBC:
It's not a "random" number, just current date and time

peter1384:
oh I see, but shouldn't it be like this instead?
2024-03-05_17-06-43
for example?

CoBC:
I did it exactly as the same way of recordings filenames.

peter1384:
yes true, but i think add a / or - simble between year/month/day, such as 2024/5/23
because currently, screenreader just reads out all the numbers like 20245236

PR #2306, opened 2024-05-23
Context menu to ban list

CoBC:
* Ability to sort banned list
* Option to move users bitween list
* Ban time converted to local format

### DIFF
--- a/ChangeLog.txt
+++ b/ChangeLog.txt
@@ -19,6 +19,9 @@ Default Qt Client
 - Optional sound event for user typing private message
 - Replace readonly edit fields by statis label in "User info" dialog
 - Only show available info in "User info" dialog
+- Ability to sort Ban List
+- Context menu to move banned/unbanned users to banned/unbanned list
+- Ban time displays in local format
 Android Client
 - users who have the "User can make other users channel operator" right will no longer be asked for the operator password for granting/revoking channel operator
 iOS Client
--- a/Client/qtTeamTalk/bannedusersdlg.cpp
+++ b/Client/qtTeamTalk/bannedusersdlg.cpp
@@ -323,7 +323,7 @@ void BannedUsersDlg::slotUnbannedContextMenu(const QPoint& /*point*/)
 {
     QMenu menu(this);
     QAction* banUser = menu.addAction(tr("&Move Selected User to Banned List"));
-    auto srcIndex = m_bannedproxy->mapToSource(ui.bannedTreeView->currentIndex());
+    auto srcIndex = m_unbannedproxy->mapToSource(ui.unbannedTreeView->currentIndex());
     banUser->setEnabled(srcIndex.isValid());
     if (QAction* action = menu.exec(QCursor::pos()))
     {
--- a/Client/qtTeamTalk/bannedusersdlg.cpp
+++ b/Client/qtTeamTalk/bannedusersdlg.cpp
@@ -254,9 +254,24 @@ void BannedUsersDlg::cmdProcessing(int cmdid, bool active)
 
 void BannedUsersDlg::keyPressEvent(QKeyEvent *e)
 {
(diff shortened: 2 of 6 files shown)


peter1384:
nice! Then, these buttons
> button remove banned user
and
<button put back to banned users should be removed to simplyfy the dialog and context menu actions does the work of these buttons.

peter1384:
the option in the unbanned tree view context menu stays unavailable even when an option is selected in it, lol because in the void of unbanned context menu, you say
    auto srcIndex = m_bannedproxy->mapToSource(ui.bannedTreeView->currentIndex());
    but you should say
        auto srcIndex = m_bannedproxy->mapToSource(ui.unbannedTreeView->currentIndex());
        lol

CoBC:
Good try, but nope, it should be:
auto srcIndex = m_unbannedproxy->mapToSource(ui.unbannedTreeView->currentIndex());
Btw, I'm glad if my mistakes are so funny

peter1384:
oh yeah! that unbanid one too
lol no, your mistakes are not funny, i just laughed because i felt we need some fun while working and stuf, and things happen.

peter1384:
actually, a very nice suggestion! The ability to unban all banned users! hear's how to
these >button and <buttons should not be removed, instead their job should be difrent.
When klicking this >button remove banned user, all items inside banned tree view should be moved to unbanned tree view
and when klicking this <button put back to banned users, all items inside unbanned tree view should be moved to banned tree view, wether they have only 1 item or 10000 items in them, these buttons should move them all
so this is an ability for server owners who have banned a lot of users from their servers and they just want to unban everybody.
But you'll need to wright 2 new voids for these buttons because currently we don't have any void that moves all items of one tree view to another tree view.

CoBC:
Just press CTRL+A in the tree view and than press the button

peter1384:
also, currently, pressing enter key in tree views just exits the dialog, but double klick in an item in banned tree view moves that user to unbanned tree view, and double klick on an item in unbanned tree view moves that item to banned tree view, So, enter key should do that as wel in tree views. Thanks!

peter1384:
like the user accounts dialog, wich enter key does the double klick's job instead of doing unexpected things

peter1384:
i don't think i have the ability to push commits to this branch to help you, if i could i'd do some things that i know how to do
currently, we have a place where it says pressing enter on the text box slot new ban, can we add 2 more elses to that lojic for tree views, or we need to wright it difrently, because it's nice and quick to press enter on an item in the tree view to move that item to the other tree view, also that's what double klick does see this
    connect(ui.bannedTreeView, &QTreeView::doubleClicked, this, &BannedUsersDlg::slotUnbanUser);
    connect(ui.unbannedTreeView, &QTreeView::doubleClicked, this, &BannedUsersDlg::slotBanUser);

CoBC:
You don't have right to push on this repo
I've applied your suggestion

peter1384:
ok then thanks. Then if anything else was good as a suggestion i'll let you know hear so you can aply

peter1384:
it doesn't work. I just built the last commit. pressing enter on the tree views just exits the dialog, actually not only tree views, pressing enter anywhere exits the dialog.
may be there is a bug somewhere else, because for example when we pressed enter on user accounts dialog it didn't exit the dialog.

CoBC:
I'm not a big fan of event filter but it seams we don't have choice for trees :(

bear101:
There's a good deal of conflicts with the "search banned users" PR

CoBC:
I think I've fixed it

bear101:
The search function using "Enter" key is no longer working with this change

peter1384:
oh, i understand why. The enter key we modifyed it so that it'll add or remove ban in tree views doesn't incclood that search thing hear, so it'll be fixed if you also inclood that when pressed enter on search text box do that

bear101:
Replaced by #2312

PR #2310, opened 2024-05-24
Chat history in a list

CoBC:
A lot of apps use list instead of readonly edit field to display chat history now, so I thought it's time to do the same in TT
It seams to works as expected with keyboard and screenreader.
Really not sure of the behaviour visually and with a mouse
I hope it's not completely broken

### DIFF
--- a/ChangeLog.txt
+++ b/ChangeLog.txt
@@ -23,6 +23,7 @@ Default Qt Client
 - Ability to copy all User Info to clipboard in "User Info" dialog
 - Ability to search by IP or username in "Banned Users" dialog
 - Support Idle Time on Linux (X11)
+- Chat history in a list
 Android Client
 - Toggle channel operator does not show password dialog if user has user-right to change channel operator
 iOS Client
--- a/Client/qtTeamTalk/chattextedit.cpp
+++ b/Client/qtTeamTalk/chattextedit.cpp
@@ -169,6 +169,20 @@ void ChatTextEdit::clearHistory()
     clear();
 }
 
+void ChatTextEdit::copyAllHistory()
+{
+    QString allText;
+    for(int i = 0; i < count(); ++i)
+    {
+        QListWidgetItem* item = this->item(i);
+        if(item)
+        {
+            allText += item->text() + "\n";
+        }
+    }
+    QApplication::clipboard()->setText(allText);
+}
+
 void ChatTextEdit::limitText()
 {
     while(count() > 1000)
@@ -216,8 +230,6 @@ void ChatTextEdit::mouseReleaseEvent(QMouseEvent *e)
 
 void ChatTextEdit::keyPressEvent(QKeyEvent* e)
 {
-    QListWidget::keyPressEvent(e);
-
     if (e->key() == Qt::Key_Enter || e->key() == Qt::Key_Return)
     {
         QListWidgetItem* item = currentItem();
@@ -228,21 +240,20 @@ void ChatTextEdit::keyPressEvent(QKeyEvent* e)
             }
         }
     }
+    QListWidget::keyPressEvent(e);
 }
 
 void ChatTextEdit::contextMenuEvent(QContextMenuEvent *event)
 {
     QMenu *menu = new QMenu(this);
-    menu->addAction(tr("Copy"), [this]() {
+    menu->addAction(tr("&Copy"), [this]() {
         QListWidgetItem* item = currentItem();
         if (item) {
             QApplication::clipboard()->setText(item->text());
         }
     });
-    menu->addAction(tr("&Clear"), this, &ChatTextEdit::clear);
-    QAction* chosen = menu->exec(event->globalPos());
-    if (chosen && chosen->text() == tr("&Clear")) {
-        clearHistory();
-    }
+    menu->addAction(tr("Copy &All"), this, &ChatTextEdit::copyAllHistory);
+    menu->addAction(tr("&Clear"), this, &ChatTextEdit::clearHistory);
+    menu->exec(event->globalPos());
     delete menu;
 }
--- a/Client/qtTeamTalk/chattextedit.cpp
+++ b/Client/qtTeamTalk/chattextedit.cpp
@@ -28,7 +28,7 @@
 #include <QTextCursor>
 #include <QUrl>
 #include <QMessageBox>
-#include <QClipBoard>
+#include <QClipboard>
 
 extern TTInstance* ttInst;
 extern QSettings* ttSettings;
--- a/Client/qtTeamTalk/chattextedit.cpp
+++ b/Client/qtTeamTalk/chattextedit.cpp
@@ -28,74 +28,26 @@
 #include <QTextCursor>
 #include <QUrl>
 #include <QMessageBox>
+#include <QClipBoard>
 
 extern TTInstance* ttInst;
 extern QSettings* ttSettings;
 
-QString urlFound(const QString& text, int& index, int& length)
+ChatTextEdit::ChatTextEdit(QWidget * parent/* = 0*/)
+: QListWidget(parent)
 {
-    QRegularExpression httpExpression("(http[s]?://\\S+)", QRegularExpression::CaseInsensitiveOption);
-    QRegularExpressionMatch httpMatch;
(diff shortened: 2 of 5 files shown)


peter1384:
Thanks! I love this! I was tired of read only edit boxes that screenreader would act stupidly with it. I'm going to try it now, if it's not the case, please also do it for private messages history as wel. Thanks!

peter1384:
it looks like some messages are being sent as difrent peaces or something, but sometimes screenreader says list item. Look at below lines, i'll paced the items of my history list so you can understand, i'm pressing down arrow and my screenreader is nvda
2024-05-25 01:35:53  * Connecting to 127.0.0.1 TCP port 10333 UDP port 10333
2024-05-25 01:35:53  * Connected to 127.0.0.1 TCP port 10333 UDP port 10333
list item
2024-05-25 01:35:53 Server Name: my pro server
2024-05-25 01:35:53 Message of the Day: hello
also, when i join a channel, the logged history messages are as follows, I mean it's 2 or 3 messages instead of one, so idk if it's expected or not, but i'll show what i mean see below.
list item
2024-05-25 01:42:47 Joined channel /Stereo/
Topic: 
Disk quota: 20000000 KBytes
also, another thing that i really need to be done and it's really bad, please display the date and time after the events raather than before because the event is more important information that we need to hear first, as an example, when we receave a private message or event, it's like this.
2024-05-25 01:46:47 <Peter>how are you
but it should be like this
<Peter> how are you 2024 blaablaablaa
and for all other events, channel messages, and stuf
please seriously do this. Thanks!

peter1384:
actually so many of my friends and anyone was complaning about the history system, but by moving it to list all those problems has been greately resolved

bear101:
This change does not work well for people who use a mouse since you cannot select text and use Copy/Paste.

peter1384:
you can coppy the message by the context menu option, then paste it in the edit box, and then select part of it and remove it or stuf

peter1384:
and read only edit text does not work wel for us with keyboard and screenreader. So, may be add an option in display tab and when checked, we use list instead of read only edit text. Just an opinion

jeremyp3:
I also use a screen reader and I don't see how the list is better than the read-only text box.

in the read-only text box, you can select a specific part of the text without having to select the whole line, and cut it out in a notepad, for example.

peter1384:
in list all item is being announced all at once, but in read only text box when you receave an event that is long or has 2 lines and stuf, you have to press down arrow to read the half of that event so it's seperated, but in list, when you receave one event, focusing on that event reads all of it, from start to end, and it doesn't need you to press down arrow because just in case it's split up in to 2 lines, also, some people report that when pressing up or down arrow to view all of their history items, some events are missing for them, but with list that problem no longer exist

jeremyp3:
I've never had the problem of the history entry disappearing when it was 
in read-only mode.

but anyway, I don't think it's worth debating the subject any further.

peter1384:
yes, if some people really wants the read only edit box, an option in the display tab or another solootion by @CoBC would fix that, so

CoBC:
We can't select a piece of text with keyboard as well.

So, I suggest the following: add the ability to show message in a 
dialog, with potential more info, E.G. the dialog can display timestamp, 
which can be useful for users who decided to don't display this info in 
the history but want to see timestamp of a specific message. In this 
dialog message will be shown in a edit field so we can select text and copy.

I also have the idea to allow deleting one item of the list, for users 
who don't want to clear all history but just remove unwanted messages.

What do you think?

CoBC:
Yes, we also have a serious problem with navigating in the readonly edit 
field with NVDA for example, some characters are repeated more and more, 
other characters are skiped, multilines messages are read very strangely 
with blank lines ignored. Accessibility of list is really better 
compared to readonly edit field.

The real problem is if mouse interactions are broken, but I made a 
suggest for it, we'll see

peter1384:
yes, exactly, those are the problems i forgot to mention. So, I strongly want the list for the history as a teamtalk user instead of read only edit box.

jeremyp3:
@peter1384 I think everyone understands your point.

beqabeqa473:
I vote against this change. If someone wants this, this should be as an option disabled by default.
Personaly i, and most of the people don't have problems with readonly editbox.

CoBC:
Let me know if I should implement this as an option, but it will be a 
bit complicate to do

peter1384:
@CoBC I don't really want to loos this feature. So, if it's realy the case, i'd want it as an option then, and i'm happy about the faqt that you understand the problems of the read only edit boxes

CoBC:
I don't really understand how we can't have problem with this readonly 
editbox, multilines fields are really problematic with Qt, especially 
when a lof of blank lines are present.

So, I also want to keep this feature, but I'm not the only one to decide :)

CoBC:
And, my last thought on this topic, majority of chat apps (at least on Windows) use this type of list for message, E.G. WhatsApp, Messenger, Phone Link (for messages from Android or iOS), and probably some other.
I know some apps like Discord, Skype, use readable content, but in a webpage, not just a readonly editbox.
I find apps with list really more practical, and I don't know if they allow to copy piece of text.
That's all for me, final choice not depend on me :)

peter1384:
github desktop is using list for the history as wel, and yes as you said, I don't know how they don't have any problem with readonly editbox. everyone around me are complaning about it that it has problems

CoBC:
One more advantage for the list I forgot yesterday, imagine you're receiving a very long message with an URL in the middle of this message, currently you have to read all the message with arrow keys for keyboard user to be able to press enter on the link and open it, with the list, you just have to press enter on the message and the link will open, and if multiple links are present, a menu will ask you what link you want to open.

Muamalaljanahi:
List is a lot better especially for NVDA users

valiant8086:
Hi.

In regard to not being able to select messages with the mouse, I would have expected the mouse selection to work like multi-selection of files in explorer does, e.g. holding right mouse button and dragging it down the messages list, or something similar to that.

amirmahdifard:
I also hate read only edit box for the history, and because of that i never used history for reading my messages, I always used my nvda speech history to read my messages, but with list i will use my teamtalk history from now on always. So, i also really want the list

peter1384:
@CoBC can you please add an action to context menu to only delete the current item, sometimes I don't wish to clear my hole list, I just wish to delete some unwanted items from my list and keep the others that I need. So, This context menu action is needed, So please add it. Thanks.

CoBC:
I'll don't make anymore commit in this PR for now until I'm sure the PR 
idea is accept.

peter1384:
Can you Please also merge the master branch to this branch, because I want to use it with laytist heehee

peter1384:
oh, ok then. Sad

CoBC:
I opened a pool
Please vote for your choice in #2313
I'll switch this PR to draft until the week-end and than, following pool results, switch back to ready or close this

CoBC:
Current results of poll (#2313) are in favor of the list, I also received 2 positive comments for this list from users who aren't on GitHub.
If we implement the list, some adjustments are however needed:
1. Ability to display message in a dialogbox (E.G. to show more message info and copy part of text)
2. Move date after message, or change to a column based list with ability to reorder column
3. Ability to delete one item of the list
@bear101 final advices and choice are yours :)

beqabeqa473:
if this will be implemented, i vote for it to be an option.

peter1384:
@beqabeqa473 somethings cannot be option because they will be implemented depending on it's better for all users. For example, I didn't liked the new q table, but I didn't request it to be option because it was depended on what moste users want and like. So, you are not the only person to say what to happen, because so many people are using teamtalk and all of those people cannot be ignoared just because of what you want. So please, don't complane and get used to whatever that is. Thanks.

CoBC:
@peter1384

You're always requesting something and something and something else 
again, with some reasons like "we need it so please add it", so please, 
be respectful with other users and contributors, if Beqa prefer this to 
be an option, even if it's very difficult to implement it as an option, 
it's his choice, we have to hear and understand all users. Please be 
careful when you comment.

peter1384:
@CoBC yes i agree with you, I just ment that we should do what users want. so, sorry for the miss understanding

CoBC:
Replaced by #2548

PR #2311, opened 2024-05-24
Use QTableView on model based widget

CoBC:
@bear101 Is this one something OK for you?
If yes, I'll change it on other table model
I find this more logic to use QTableView for multicolumn widget
More, this widget doesn't require accessible text role anymore, we can navigate with arrow keys like any real table, which is more efficient for me.

### DIFF
--- a/ChangeLog.txt
+++ b/ChangeLog.txt
@@ -27,7 +27,7 @@ Default Qt Client
 - Context menu to move banned/unbanned users to banned/unbanned list
 - Ban time displays in local format
 - Ability to leave current channel from everywhere in main window
-- Table instead of tree on Online Users, User Accounts, Banned Users, Server List, File List, and Preferences
+- Table instead of tree on Online Users, User Accounts, Banned Users, Server List, File List, Server Properties, and Preferences
 Android Client
 - Toggle channel operator does not show password dialog if user has user-right to change channel operator
 iOS Client
--- a/ChangeLog.txt
+++ b/ChangeLog.txt
@@ -27,6 +27,7 @@ Default Qt Client
 - Context menu to move banned/unbanned users to banned/unbanned list
 - Ban time displays in local format
 - Ability to leave current channel from everywhere in main window
+- Table instead of tree on Online Users, User Accounts, Banned Users, Server List, File List, and Preferences
 Android Client
 - Toggle channel operator does not show password dialog if user has user-right to change channel operator
 iOS Client
--- a/Client/qtTeamTalk/CMakeLists.txt
+++ b/Client/qtTeamTalk/CMakeLists.txt
@@ -90,6 +90,7 @@ if (Qt5_FOUND OR Qt6_FOUND)
     serverlogeventsmodel.h mytreeview.h textmessagecontainer.h
     useraccountsmodel.h encryptionsetupdlg.h utiltt.h utilxml.h
     utilos.h serverdlg.h moveusersdlg.h useraccountdlg.h soundeventsmodel.h
+    mytableview.h
     main.cpp mainwindow.cpp preferencesdlg.cpp uservideowidget.cpp
     channelstree.cpp channeldlg.cpp userinfodlg.cpp
     bannedusersdlg.cpp useraccountsdlg.cpp videogridwidget.cpp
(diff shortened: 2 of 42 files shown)


bear101:
The row height is quite big in QTableView compared to QTreeView but I can probably fix that. So go ahead and change to QTableView

peter1384:
I don't know if this is expected for this for now because not complete or something, but with this my banned users items are announced like this with nvda when I press up and down arrow.
data item  selected  row 1
data item  selected  row 2

CoBC:
It's just because the column you're navigating in is empty, so press 
right arrow and you'll see

peter1384:
oh no, so for seeing all of the information of a ban such as creater, ban type, ban time and all others we have to press right arrow now? this is not good at all!

CoBC:
Yes, it's exactly the goal of a table, on all tables I know, you have to 
navigate in it to see info.

Personally, I find it really more convinient to be able to quickly see 
the info I want, E.G. ban type, instead of have to hear all previous info.

CoBC:
I just pushed for online users. Unfortunately context menu actions seams to be broken :(
I'm searching why but if anyone has an idea before me, it will be really good :)

peter1384:
then in this case personally i disagree with this and i don't know if i'd be able to use it properly because this is gonna be replaced by all tree views

CoBC:
Sorry, but I don't really understand your problem with these changes

peter1384:
my problem is that i have to press right arrow to view other info, but in tree view i could see everything
now i have to press right arrow to see username, ip, nickname, ban type, and this is gonna be replaced with everything, it's not comfurt, I didn't know the point of tables are this.

CoBC:
And, if, for example, I want to see the ID of an online user, currently, 
I have to heard nickname, status message, IP, version, and more, before 
the ID, I have the time to drink a cup of tea before hearing the info I 
want.

CoBC:
All will be read on first column (accessible text role), and we can continue to navigate with arrow keys to only read other info
Hope everyone is happy with this behaviour

ChrisDuffley:
Yes. Other than left/right arrow keys not navigating each column then this is exactly like how a list view would behave. At least by what you're saying here. I'll still have to figure out how I'll test this when I'm back at a computer again.

Thank you for this change, I also may want to hear only one column most of the time. Does first/multi-letter navigation still work in these elements and if so, what column is it based on?

CoBC:
Yes, one letter key navigation works. It's based on first column, depending on table.

CoBC:
Done for me, just online users context menu is still broken, but don't understand why for now

ChrisDuffley:
Will that method change depending on sort (via Sort By...) or no, always by the first column?

CoBC:
The based column is always the same, but depending on sort result order will change

bear101:
@CoBC can you rebase with master branch to get a clean commit history?

CoBC:
I've rebased. Not sure why Git is still showing 7 commits. Maybe I'm not a rebase expert :(

bear101:
@CoBC Column 0 in server list shows "Name", "Users" and "Country" as rows. And also many empty rows. "Name", "Users" and "Country" are also in header of columns 1, 2, 3.

bear101:
@CoBC Also header sizes are not correct width initially. I have to adjust each of them manually. Is there a way to have horizontalHeader() to manually set preferred sizes if the size is not already stored?

CoBC:
Hmmmm, other dialogs don't have the same issue? Strange if server list 
is the only one with this issue.

For horizontalHeader, I have to check again the doc :)

CoBC:
@bear101 what about resizeSection? Is not this emthod you want to set 
default size?

CoBC:
I pushed a new commit, hope this helps to fix server list

bear101:
@CoBC I'll clean up the commit history and do a force push

bear101:
@CoBC Your old branch is available as tableview-old

bear101:
@CoBC when you select a server in "Server List" dialog and press "Connect" button it has no effect.

CoBC:
This issue is also present in master branch, right? In all case this 
button doesn't have any effect for me when I select an item on the tree, 
tab to connect button and than press space. So, does it works in master 
branch with mouse?

bear101:
Hm, yes, on master branch it also has this issue. The issue is the same with and without mouse.

I guess it makes sense that the focus check fails when none of the lists are focused :(

CoBC:
Yes, you're right. But, do you see any other easy way to connect the 
"Connect" button to an action without this focus check? Or maybe we have 
to add a 2nd "Connect" button?

bear101:
Maybe the "Connect" button should check selection instead. If one list has an item selected then it chooses that one.

CoBC:
Should I make this change in this PR now? Or are you working on 
something else on this PR and I'll open another PR for this fix later?

bear101:
Should we just remove the "Connect" button? It's a bit ambiguous when both list views have selection.

Please make the change from master branch. I'm currently review the QTableView changes.

CoBC:
Yes, I thought to the same thing, this "Connect" button is not really useful anymore, we have double click, enter key and context menu to connect, I'll remove it.

bear101:
Is it ok that I enable row selection in all the table views? It's more logical than cell selection

User Accounts dialog is already like this:
```
         <property name="selectionBehavior">
          <enum>QAbstractItemView::SelectRows</enum>
         </property>
```

CoBC:
Yes, it's OK

PR #2312, opened 2024-05-26
Rebased bannedusers_contextmenu branch

bear101:


peter1384:
@CoBC Now that i'm testing this, As I said before also, the enter key in user accounts tree view works as expected with out event filter, also all others are working fine with out event filter, but why only banned users dialog doesn't work with out event filter? what's wrong with only that dialog?

CoBC:
I've removed event filter in the PR with table instead of tree

peter1384:
@CoBC but currently user accounts is stil tree and it doesn't have the event filter and it works properly

CoBC:
Because keyPressEvent was implemented better. It's why, as I said 
previously, I removed event filter in another PR

PR #2314, opened 2024-05-27
Ability to leave channel from everywhere in main window

CoBC:


peter1384:
i'm sad that with this pr the join channel action is no longer a togle option and we have a new action leave channel, but ok

bear101:
When you select your own channel then you only have the "Leave Channel" menu item. If you select a user in your channel then you have both "Join Channel" and "Leave Channel". I figure it should show the same in both cases.

CoBC:
Ah yes. What is better, probably to hide "Join channel"?

peter1384:
@CoBC can you just remove JoinChannel and LeaveChannel actions and just make a JLChannel or something else action, and make it toggle between do leave channel and do join channel when it's the case like before?

PR #2318, opened 2024-05-28
Remove MyTreeView

CoBC:
Forgot to remove but not requires anymore after switch to QTableView


peter1384:
@CoBC sorry, I mentioned you because i'm talking to you.
Currently, tables has a fue problems, That I don't know you can fix them or qt devs should fix that.
Currently, after the anouncement of what we want, NVDA always says this word data item, data item, data item.
2: The server list table is kind of broken.
first, Even the local servers has the users, cuntry and motd columbs in them, but local servers should not have these.
Then, When pressing right arrow on a published server, the server's motd is announced for all of the columbs "users, cuntry, and motd it self", so we can't see the number of users in the users columb, instead, we hear the server's motd in all columbs.For Now I figgered these.

bear101:
@CoBC this doesn't compile

CoBC:
Oh, wow, with all modified files commited it will probably be better :)

PR #2319, opened 2024-05-28
Remove macOS TTS speak on list option

CoBC:
Following switch to QTableView, and Qt6 providing native accessibility for trees on MacOS, this option seams to be useless now.


CoBC:
Doesn't compile because MyTreeView is still present in this PR, it will be fix when #2318 will be merge

PR #2321, opened 2024-05-29
Accessible text role on username column in user accounts dialog

CoBC:


peter1384:
user accounts accessible text role is like this incorrectly.
Username: player, Password: player, User Type: Default User, Note: , Channel: , Modified: 5/18/2024 6:58 PM  data item  selected  Modified  column 6

peter1384:
currently as i'm testing, files list and online users are ok and doesn't have any problem and their entire accessible text role is at the first columb which should be, but server list and user accounts list is a bit broken and incorrect.

CoBC:
It's exactly what I changed in this PR, so don't understand your comment 
here

peter1384:
also, banned users doesn't have the nickname columb, and pressing enter on banned or unbanned table just exits the dialog.

CoBC:
1. This PR fix accessible text role on user accounts dialog
2. Server list behaviour is fix in PR #2322 
3. Enter key on Banned users dialog is fix in PR #2324 
4. Don't understand the issue with nickname column in banned users dialog, all column are present

peter1384:
wel, when I press right arrow in banned users table, I see username, ip, ban type, time etc, but no nickname, unless the very first columb which has the accessible text role in it is the nickname columb.

CoBC:
Yes, first column is the nickname column, as it was also the case 
previously, nothing has changed.

This "data item" is just an announcement by NVDA because it's the type 
of control reported by Qt, not sure why it's really a problem, the info 
you want is says firstly so I don't see a real issue here

peter1384:
ok i understood now, but i have an idea. Instead of accessible text role being on first columb like nickname columb, create an empty columb before the first columb, and put the accessible text role contaning all information in it, so then we can see nickname columb as wel if we press right arrow, and this should be done for all tables. This is my suggestion.

PR #2325, opened 2024-05-29
Fix "Reset to default value"

CoBC:
Only check if message is different from current message, don't check default value

### DIFF
--- a/Client/qtTeamTalk/preferencesdlg.cpp
+++ b/Client/qtTeamTalk/preferencesdlg.cpp
@@ -1357,8 +1357,10 @@ void PreferencesDlg::saveCurrentFile()
     const SoundEventInfo& eventInfo = eventMap[eventId];
     QString paramKey = eventInfo.settingKey;
     QString text = !ui.soundEventFileEdit->text().isEmpty()?ui.soundEventFileEdit->text():UtilSound::getDefaultFile(paramKey);
-    if (text != ttSettings->value(paramKey))
-        ttSettings->setValue(paramKey, text);
+    if (!text.isEmpty())
+    {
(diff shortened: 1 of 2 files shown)


Client/qtTeamTalk/preferencesdlg.cpp
```diff
@@ -1812,7 +1812,7 @@ void PreferencesDlg::saveCurrentMessage()
         QString paramKey = eventInfo.settingKey;
         QString text = ui.TTSMsgEdit->text();
 
-        if (!text.isEmpty() && text != UtilTTS::getDefaultValue(paramKey) && text != ttSettings->value(paramKey))
+        if (!text.isEmpty() && text != ttSettings->value(paramKey))
```
bear101, line 1815: shouldn't `ttSettings` clear the value if `text == UtilTTS::getDefaultValue(paramKey)` ?

So instead:
```
if (!text.isEmpty())
{
    ttSettings->setValue(paramKey, text == UtilTTS::getDefaultValue(paramKey) ? "" : text);
}
```

Client/qtTeamTalk/statusbardlg.cpp
```diff
@@ -134,7 +134,7 @@ void StatusBarDlg::saveCurrentMessage()
         QString paramKey = eventInfo.settingKey;
         QString text = ui.SBMsgEdit->text();
 
-        if (!text.isEmpty() && text != UtilUI::getDefaultValue(paramKey) && text != ttSettings->value(paramKey))
+        if (!text.isEmpty() && text != ttSettings->value(paramKey))
```
bear101, line 137: Same here

PR #2326, opened 2024-05-29
Ability to Reorder Column of "User Accounts" and "Online Users" dialog

CoBC:


Client/qtTeamTalk/mytableview.cpp
```diff
@@ -48,3 +53,47 @@ void MyTableView::keyPressEvent(QKeyEvent* e)
         QTableView::keyPressEvent(e);
     }
 }
+
```
bear101, line 56: Can this commented code be removed?

Client/qtTeamTalk/mytableview.cpp
```diff
@@ -17,9 +17,14 @@
 
 #include "mytableview.h"
 #include <QAccessible>
+#include <QShortcut>
 
 MyTableView::MyTableView(QWidget* parent/* = nullptr*/) : QTableView(parent)
 {
+    QShortcut* shortcutLeft = new QShortcut(QKeySequence(Qt::CTRL | Qt::Key_Left), this);
```
bear101, line 24: Shouldn't this be documented somehow? Or is it a generally know shortcut?

CoBC: Yes, it could be better to document this, but where can we do it?

CoBC:
Hmmm, it seams headers don't follow column reorder, E.G. move username to column 2 instead of 3 in online users dialog, NVDA is still announcing username as column 3, however the column is correctly place in 2nd position in the table.
Don't know if this issue is also visually present :(

Client/qtTeamTalk/mytableview.cpp
bear101: In https://github.com/BearWare/TeamTalk5/blob/master/Documentation/TeamTalk/userinterface.md section "User Accounts dialog".

CoBC: OK, done! We should also update doc with new dialogs, but probably in a future PR or commit

Client/qtTeamTalk/useraccountsmodel.cpp
```diff
@@ -23,6 +23,11 @@ UserAccountsModel::UserAccountsModel(QObject* parent)
 {
 }
 
+void UserAccountsModel::setTableView(QTableView* view)
+{
```
bear101, line 27: The view class should not be passed to the model. Then the model cannot be used by other classes. What you need in the model has to be inside the model.

Client/qtTeamTalk/useraccountsmodel.cpp
```diff
@@ -79,14 +84,23 @@ QVariant UserAccountsModel::data(const QModelIndex & index, int role /*= Qt::Dis
...
+            if (tableView && index.column() == tableView->horizontalHeader()->logicalIndex(0))
+            {
+                QString accessibleText;
+                int columnCount = tableView->model()->columnCount();
```
bear101, line 92: Here call this->columnCount(index) instead.

Client/qtTeamTalk/useraccountsmodel.cpp
CoBC: Do you know how to retrieve horizontalHEader or something similar directly inside the model? I don't see how to do it without pass the tableView to the model :(

bear101: hm, not quite sure. Maybe you have to tell the model of the column index order, since it should be irrelevant to the model.

CoBC: I tried something, but the behaviour is totally broken :(
So, for now, not sure how to do, I have to do more experimentation

CoBC:
I didn't really understand the issue to pass view to model? Which other class needs to use affected models?

bear101:
Models should be independent of the view. Otherwise a chosen view cannot be replaced by another. You can read about model/view programming here: https://doc.qt.io/qt-6/model-view-programming.html

What you're doing is a hack since you hardcode outside behaviour inside the model. The model should not know anything about the outside world. I'm aware it works what your doing but it's an incorrect (i.e. hacked) approach to models.

The UserAccountsModel class should not be concerned about rearranged columns. The model should only be concerned about providing data through a standard interface (QAbstractTableModel) and is therefore independent of what some might have changed in the outside world.

CoBC:
Sorry, but I can't achieve a working result any other way. So if you know and have the time to do it differently, you can do it. Otherwise, we'll probably have to close this.

bear101:
Also please rebased instead of merging. It's hard to read the history after merging.

CoBC:
OK, thanks. I'll see how you did it. So if it's OK for you, it's OK for me, let me know if you want something else from me in this PR :)

CoBC:
Hi @bear101 
we have an issue with user accounts dialog, accessible text role isn't present, first visible column only have display role so screen reader only announce data for this specific column. I tried to understand why, but I didn't find anything :(
Note also that the horizontalHeader->setSectionMovable(true) is missing for this dialog, so we probably have to add it, but anyway this addition doesn't fix the issue mentioned.
Can you take a look?
Thanks.

bear101:
"Online Users" dialog doesn't have this problem?

CoBC:
No, only user accounts dialog have this problem :(, all works fine in 
online users dialog

PR #2333, opened 2024-06-01
Ignore limit text size in channels tree for screen readers

CoBC:


peter1384:
would you also please explane this? thanks

bear101:
Doesn't the change remove the number of users part of the channel name? The parantesis part, e.g. "(4/6)"

CoBC:
No, it doesn't, channel name, users count, channel topic, and potential 
emojis are added before accessible text role definition

CoBC:
OK, I finally understood the issue, you're right, count was removed to display role with this change, so, I made a different string for display and accessible text roles

peter1384:
i've built this, and i haven't noticed any changes in the channel tree with my screenreader. Can you please explane what does this pull request do? i'd like to know

CoBC:
By default, text in channels tree (including channel name, channel 
topic...) are limited to 50 characters, this limit can be increase to 
512 characters in preferences. So, in this PR, I removed this limit for 
accessible text role, so SR always ahve entire item text available, and 
the limit only concerns display

peter1384:
oh very nice! can't this 512 limits be removed or the ability to increace it to more than 512 characters for display users as wel? I think it would be a good idea

CoBC:
Not sure a lot of screens support more than 512 characters per line

PR #2334, opened 2024-06-01
Ability to customize timestamp format

CoBC:


peter1384:
what is timestamp? can you discribe please?

CoBC:
Date and time in chat history, files list, user accounts, banned users

bear101:
@CoBC should the QLineEdit and Variables button be disabled if "Timestamp" checkbox is unchecked?

CoBC:
Probably not, because checkbox effect is only for chat history, but 
timestamp format is apply for all date and time (E.G. in file list, 
banned users dialog, user accounts dialog...)

CoBC:
@bear101 Should we add a button to reset to default format? And, if yes, do we have place to add one more button in this tab? Display tab is very big :(

bear101:
@CoBC can't blank field be default format?

CoBC:
OK for blank field for default format, so it's already the case

PR #2338, opened 2024-06-05
Check for active events before puting event to queue

CoBC:
Fix a bug where sound events are played even if sound event is disable

### DIFF
--- a/Client/qtTeamTalk/utilsound.cpp
+++ b/Client/qtTeamTalk/utilsound.cpp
@@ -540,8 +540,11 @@ PlaySoundEvent::PlaySoundEvent(QObject* parent) : QObject(parent)
 
 void PlaySoundEvent::queueSoundEvent(SoundEvent event)
 {
-    m_playbackQueue.enqueue(event);
-    playSoundEvent();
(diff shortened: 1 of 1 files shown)


peter1384:
i've noticed you have some small spelling errors, and that even aplyed to teamtalk's menus. Sometimes, you need to put double t on your word, such as.not puting, it's putting

not emited, it's emitted
and some others.

CoBC:
applied, it's not aplyed. I think you're really not welcome to correct 
my spelling

peter1384:
yes I have spelling errors also, but something I know that happened in teamtalk, In text to speach variable menu, You say
Server's name from which event was emited
But It should be
Server's name from which event was emitted
If you don't beleave this, You can look at dictionaries and stuf

Client/qtTeamTalk/utilsound.cpp
```diff
@@ -540,8 +540,11 @@ PlaySoundEvent::PlaySoundEvent(QObject* parent) : QObject(parent)
 
 void PlaySoundEvent::queueSoundEvent(SoundEvent event)
 {
-    m_playbackQueue.enqueue(event);
-    playSoundEvent();
+    if (ttSettings->value(SETTINGS_SOUNDEVENT_ACTIVEEVENTS, SETTINGS_SOUNDEVENT_ACTIVEEVENTS_DEFAULT).toULongLong() & event)
```
bear101, line 543: Shouldn't it be placed in `void playSoundEvent(SoundEvent event)` instead?

PR #2344, opened 2024-06-09
Replace some texts by emojis in channel list

CoBC:
If someone has any idea for channel operator and administrator, maybe we can also replace "channel operator" and "administrator" by emojis. This make channel item shorter


peter1384:
may be 👑 can be used instead of administrator

PR #2346, opened 2024-06-10
Change to QLabel instead of ReadOnly QLineEdit in "Server Statistics" dialog

CoBC:


peter1384:
thanks, @CoBC Will you also simple the invalid account dialog as explaned? If you are not convinced why it is useful, i'll explane here
lets say you've entered your username, and you pressed ok button and the password dialog appears, but then you've remembered that you've entered the username incorrectly. Currently you can't go back and fix your username edit box, but if my idea is implemented, You can press shift tab to correct all other entrys, like when you want to create an account. This is important because some servers administrators configger that if you are logged in 2 or 3 times with an invalid info  you'd get banned, Max login attempts before ban  spin button  2
So, Will you do this in this pull request, or will you do it in another one?
Thanks for your valuable work

CoBC:
I'm working on it, just be patient

PR #2349, opened 2024-06-10
Add slider in main window and global shortcut to adjust media file volume

CoBC:
@bear101 Do we have place for this new slider?
I'm also wondering if we can group sliders in groupbox (E.G. input volume and output volume)


bear101:
I think this looks confusing. Most people don't use the extra slider. I think it should be placed in the stream media file dialog instead.

CoBC:
Why in "Stream Media" dialog? It's not a bit strange to adjust output 
volume from a dialog to play a sound? Also, some users want a simple and 
quick way to configure this media volume without go to preferences dialog

CoBC:
Also note that some users has asked for the ability to hide some elements of main window, so I suggest to add a "View" menu where we can show/hide some sliders, this new stream volume slider can be hide by default and enabled by users who want it.
What do you think?

bear101:
Can we make a Hotkey shortcut to increase/decrease media stream volume?

bear101:
The Hotkey shortcuts in Preferences dialog are user defined (global hotkeys).

bear101:
There is still space in the "Display"-tab but we should probably rearrange some checkboxes so similar ones are grouped together.

CoBC:
Can you update the UI file to rearrange checkboxes and a the new ones to 
main window items? I'll do the rest

CoBC:
@bear101 We should consider reorder shortcuts tab to group similar hotkeys, E.G. master volume, media file volume and microphone gain
Can you do it?

bear101:
Shortcuts-tab only has 9 items.

I think it would be better to turn it into a table-view, so it's easy to add new shortcuts.

CoBC:
Yes, good idea. DO you have time to do this, or do you want I work on it?

bear101:
I don't have time to look at it at the moment, so if you have time please go ahead.

CoBC:
OK, I'll see. I'll make another PR when it will be ready, so you can 
review this one

bear101:
@CoBC would it be an idea to make a "Media Playback" tab in the Chat/Desktop/WebCam tab-control on Main window? This tab could should progress, volume control, etc.

CoBC:
Oh yes, sounds good

PR #2353, opened 2024-06-11
Remove "Close" button from "Server Statistics" dialog

CoBC:


bear101:
The layout of many of the dialogs get screwed up by these close buttons. Why do we need them? Can't users just press Escape?

CoBC:
This was an user request, but I'm OK to keep escape to close dialog

However, maybe remove the "Close" button from "Server Statistics" dialog 
to align dialogs behaviour

PR #2354, opened 2024-06-11
Translate installer

CoBC:
Just add French to see if it works. If it's OK, we can add more languages

### DIFF
--- a/ChangeLog.txt
+++ b/ChangeLog.txt
@@ -4,6 +4,7 @@
 
 Version 5.17, unreleased
 Default Qt Client
+- Windows installer translatable
 - "Connect to a Server" and "User Accounts" dialog simplified with multiple sub dialogs
 - Table view instead of tree view in "Online Users", "User Accounts", "Banned Users", "Server List", "File List", "Server Properties" and "Preferences" dialogs
 - "Move Users" dialog from "Users" -> "Advanced" menu item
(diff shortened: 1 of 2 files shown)


CoBC:
@bear101 Maybe we should find a way to integrate translations for component names
Do you think it's easy for translators to understand how to add translation directly to this InnoSetup script? Or should we ask known translators to send us translations and integrate the changes ourselves?

bear101:
Yes, it works. I wonder if there's a way to just include all the languages that are automatically included with Inno Setup.

CoBC:
Inno Setup examples include language one by one. Also we don't support 
all languages in TT, E.G. Inno Setup provide translation for Armenian 
but we don't have Armenian translation in TT, so IMO it make no sense to 
provide an installer in a language don't supported by the app

CoBC:
I can add all languages we support in TT in the installer when they're 
available in Inno Setup, but how translators will translate E.G. 
components name? Also languages not supported by Inno Setup (E.G. 
Chinese) can be translated by our translators

bear101:
Hm, yes, ok

CoBC:
Done! So translators can give fr.Client and fr.Full as example to see how to translate components name.

Nardol:
> Also languages not supported by Inno Setup (E.G. Chinese) can be translated by our translators

And will these language actually not supported by Inno Setup be contributed upstream?

PR #2355, opened 2024-06-11
Replace language selection by auto detection

CoBC:


bear101:
Shouldn't we make a "show once" dialog that asks the user to choose language?

CoBC:
It's exactly the previous behaviour, but I found it more user friendly 
to set language based on system locale

DO you prefer the previous behaviour with dialog to choose language?

bear101:
Yes, I prefer the previous way. Creating a mapping just adds more maintenance when new languages are added.

CoBC:
OK, closing for now.
However I think to a potential alternative way, we can rename language files to use language codes and so the process will be automatic without any more maintenance needed.
We just have to update settings at first version startup
I still think automatic detection is a more common approach, especially if #2354 is approve, it seams confusing to have an installer in user language and than an app which ask to choose a potential different language

Issue #2351, opened 2024-06-11
bug in qt teamtalk. It's not possible to add a laytist host to the server list with the provided option


HanaDadfar:
### Description

When klicking on the add to saved hosts action on the context menu, the view server information dialog appears. But the edit server dialog should appear allowing us to edit the wanted info such as server name and some other info, and then press save and close button to add it to the server list

### Application

- [X] qtTeamTalk
- [ ] TeamTalkAndroid
- [ ] iTeamTalk
- [ ] TeamTalkClassic
- [ ] TeamTalkServer

### Platform

- [X] Windows
- [ ] macOS
- [ ] Android
- [ ] iOS
- [ ] Linux

### Expected behavior

when klicking add to save hosts, the edit server dialog should appear, allowing us to interact with the informations

### Actual behavior

When klicking on add to saved hosts, the view server information dialog appears

### Steps to reproduce problem

press application on a host in the laytist hosts list, and then press enter on add to saved hosts action

CoBC:
Can't reproduce

HanaDadfar:
ok, I've figgered out the problem, But it's really hard to explane
The problem is that, the add to saved hosts will open the dialog exactly as how last time it was opened, such as edit server or view server, let me explane
0: go to server list table view
1: select an unofficial or official server
2: open context menu and press edit
the view server dialog will open
3: close the dialog, and go to laytist hosts list
4: select a host, open context menu, and press add to saved hosts
the view server dialog will appear
5: Close the dialog, and go to server list table view
6: select a local server, open context menu, and press edit
the edit server dialog will appear
7: close the dialog, and go to laytist hosts list
8: select a host, open context menu, and press add to saved hosts
the edit server dialog will appear
If you follow these steps exactly one by one in your teamtalk, you'll understand what do I mean

CoBC:
OK, the issue is when a server is selected in server table view, because we check if selected server from table view is local or not to determine dialog type to display. I'll see

CoBC:
Fixed in PR #2352

CoBC:
Fixed by #2357

PR #2356, opened 2024-06-11
removed server read only mode

HanaDadfar:
* fixes https://github.com/BearWare/TeamTalk5/issues/2351 as standard
* code cleanup
* the ability to edit hole information of an unofficial and an official server, and press the save button to save it to your local server list for your later use
* the ability to specify nickname for the server while saving it in to your local server list https://github.com/BearWare/TeamTalk5/discussions/2276 like the old host manager that didn't have this limit
* built the app with these changes and tested everything


HanaDadfar:
@bear101 I didn't have the option to request a review because i'm not a contributor, But i'm requesting a review from you in comment.
Thanks

bear101:
fixed by #2357

PR #2352, opened 2024-06-11
Override server dlg type when calling from latest host

CoBC:
Fixes #2351

PR #2357, opened 2024-06-11
Alternative way for #2352

bear101:
When calling getSelectedHost() we get a HostEntryEx from the models and the HostEntryEx defaults to SERVERTYPE_LOCAL when a latest host is selected.

### DIFF
--- a/Client/qtTeamTalk/serverlistdlg.cpp
+++ b/Client/qtTeamTalk/serverlistdlg.cpp
@@ -449,7 +449,7 @@ void ServerListDlg::slotImportTTFile()
 
 void ServerListDlg::slotConnect()
 {
-    HostEntry host, latestHost;
+    HostEntryEx host, latestHost;
     if (getServerEntry(0, latestHost, true) || ui.hostListWidget->count() == 0)
     {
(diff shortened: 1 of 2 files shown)


HanaDadfar:
@bear101 in my pull request, I fixed a lot of user's complanes, They say in the old manager they could edit the hole public server information and even specifying the nickname and pressing save to server list would save the server in to their local server list, but they cannot do that in the new host manager, So I've made it so that anyone doesn't have these kind of problems with the new host manager, so, do you think to merge my pull request instead? Because the view server information which opens when we try to save a published server is unwanted in this new host manager by a lot of people and they want to use the old host manager for this reason, So, what do you think?

peter1384:
what is even the point of server readonly mode? I mean, What is the point of viewing a server's information but not being able to do anything with it? so, I also agree with @HanaDadfar here

peter1384:
the readonly mode makes sence for the new user accounts dialog system because default users cannot edit a user, so they must be able to view their user account info, but this readonly mode for the host manager does not have any point

CoBC (approved):

PR #2358, opened 2024-06-11
Set default user rights when creating new user account

bear101:
Previously default user rights were USERRIGHT_NONE.


bear101:
@CoBC if you edit a user account and change the username then the old user account is still active.

CoBC (approved):
I've addressed your comments

PR #2363, opened 2024-06-12
Remove unused config keys from config file at first version startup

CoBC:
Some TTS config keys are no longer use and can be remove from config file


bear101:
Can you move this to MainWindow::loadSettings() where it cleans up old configuration properties?

CoBC:
Hmm, not sure to understand, it's already in MainWindow::loadSettings()

bear101:
Ah, sorry. I didn't expand long enough on GitHub

PR #2364, opened 2024-06-12
Split Display-tab in Preferences dialog into group boxes

bear101:


CoBC:
@bear101 Shouldn't also hide label_4?
Also, do you see other widgets we can allow to show/hide?

bear101:
Yes, you can hide the voice activation label (label4). Maybe also rename it to something more expressive.

If VU-meter updates are disabled then we can also hide that label.

bear101:
@CoBC you can mark it is ready when you've finished your changes.

CoBC:
I think we can already merge this one. Some work will be done in #2349
Also, do you think it can be good to allow hide some tabs like desktop and video?

bear101:
I'd rather keep desktop and video tabs since they provide easy access to start/stop video/desktops.

CoBC:
OK

PR #2366, opened 2024-06-13
Support Qt Accessibility Announcement for TTS events

CoBC:


CoBC:
As the object pointer is main window, it should works, but, it doesn't work.
I tried to understand with a Qt dev why, but without any success for no.
Any idea is welcome.

CoBC:
Maybe we can change "Qt Accessibility Announcement" by something more expressive in TTS engine combobox.
I can suggest "Running Screen Reader (via Qt)" or just "Running Screen Reader".
@bear101 Do you plan to switch to Qt 6.8 beta for 5.17 release to include this change?

amirmahdifard:
@CoBC can my compilation error be related because i'm stil using qt 6.7.0 and not 6.7.1? but I didn't have problem until yesterday. If I update my qt do you think it will be resolved?? if yes, then i'll just wait until you update to 6.8 and then i'll update

CoBC:
You should be able to build this PR on all Qt versions

amirmahdifard:
@CoBC no, i'm not only talking about this pr, i'm talking about having compilation error when building the laytist master, after the mytab has been added, but I don't have problem with this pr

CoBC:
I already replied in #2375
Re-run CMake, and if you're still using QMake to build, switch to CMake, I can't say more
Btw, it's not really the appropriate place to talk about build issues

amirmahdifard:
@CoBC I'm sorry because i talk about build here, but, is it really no longer possible to build teamtalk with qmake? i'm following the commands on qt teamtalk/readme.md, but it is no longer possible.

bear101:
@CoBC Didn't I switch to Qt 6.8-dev at some point where you discover a critical accessibility bug?

CoBC:
The issue is fixe in 6.8 beta released WednesdayEnvoyé de mon iPhoneLe 15 juin 2024 à 07:15, Bjørn Damstedt Rasmussen ***@***.***> a écrit :﻿
@CoBC Didn't I switch to Qt 6.8-dev at some point where you discover a critical accessibility bug?

bear101:
@CoBC the checked/uncheck issue in QTableView has been fixed in Qt 6.8

CoBC:
Are you talking about the one discussed in #2373? If yes I guess we can revert my commit which fixed this

bear101:
Ah, sorry. I didn't see that it had been merged to master. It's still necessary :(

Client/qtTeamTalk/utiltts.cpp
```diff
@@ -124,9 +125,12 @@ void addTextToSpeechMessage(const QString& msg)
             .arg(noquote));
 #endif
         break;
+    }
     case TTSENGINE_QTANNOUNCEMENT:
 #if QT_VERSION >= QT_VERSION_CHECK(6,8,0)
         QAccessibleAnnouncementEvent announcementEvent(announcerObject, msg);
+        if (ttSettings->value(SETTINGS_TTS_ASSERTIVE, SETTINGS_TTS_ASSERTIVE_DEFAULT).toBool() == true)
+            announcementEvent.setPriority(QAccessible::AnnouncementPriority::Assertive);
```
bear101, line 133: Isn't there an "else" missing to reset to default?

bear101: ah, ok, it reverts to QAccessible::AnnouncementPriority::Polite

Client/qtTeamTalk/mainwindow.cpp
```diff
@@ -4247,6 +4250,12 @@ void MainWindow::slotClientPreferences(bool /*checked =false */)
         ttSpeech = nullptr;
     }
 #endif
+#if QT_VERSION >= QT_VERSION_CHECK(6,8,0)
+    if (ttSettings->value(SETTINGS_TTS_ENGINE, SETTINGS_TTS_ENGINE_DEFAULT).toUInt() == TTSENGINE_QTANNOUNCEMENT && announcerObject == nullptr)
+        QObject* announcerObject = this;
```
bear101, line 4255: This looks incorrect

CoBC: Why incorrect?

bear101: You create a local variable `announcerObject` pointing to `this`.

CoBC: Yes, because we need a QObject representing a object-class to use QAccessibleAnnouncementEvent, so main window is extern to UtilTTS. I didn't saw a better way, do you see one?

bear101: but you just create an unused variable.

Isn't it the global variable `announcerObject` that you want to assign to this?

CoBC: Ohhh, yes. It should be better after 067b045

PR #2372, opened 2024-06-13
Subclass QTabWidget to override keyPressEvent

CoBC:
If #2370 is accept, we'll have to replace QTabWidget by MyTabWidget also on Server dialog


bear101:
Good idea to sub-class it to avoid duplicate code. Maybe this shortcut change should be upstreamed to Qt project. Any idea why they don't have this feature?

CoBC:
I opened an issue on the Qt bug tracker longtime ago, but they still 
haven't implement this:

[QTBUG-95805] Home and End keys not working on a tab bar - Qt Bug 
Tracker <https://bugreports.qt.io/browse/QTBUG-95805?filter=-2>

Accessibility improvements are always very long to be implemented in Qt :(

PR #2376, opened 2024-06-14
Ability to send toast notifications in addition to TTS events

CoBC:


CoBC:
@bear101 I see build errors from MacOS workflow. Do you know if GitHub workflows have issue with Objective C++, or if my code have real problems for MacOS?

amirmahdifard:
@CoBC as I see in the changed files, you've updated cmake lists.txt, and added the newly added files, but you forgot to update the mainwindow.pri. So please also update the mainwindow.pri to prevent compilation errors.

CoBC:
Replaced by #2542

PR #2382, opened 2024-06-15
Use system language by default and ask to continue

CoBC:
Based on the previous PR #2355 with following improvements:
* Ask user to continue with detected language or to switch to another one at first startup
* Display language in preferences with their original names
* Make all process automatically (no more maintenance required for us)

I think it can be a good thing to display in user language at first startup for users who are really not confortable with English. More, after the merge of #2354 it seems a bit confusing to have an installer program available in more languages and than an app which talk in english


bear101:
@CoBC The Polish language has been move to an incorrect folder. Also is it really supposed to be called lapl.ts?

bear101:
@CoBC dammit, I think this ini-file version check has never worked. You can see that `SETTINGS_GENERAL_VERSION` is only called with` setValue() `if a migration takes place. And `ttSettings->value(SETTINGS_GENERAL_VERSION, SETTINGS_GENERAL_VERSION_DEFAULT )` defaults to latest version. Doh!

CoBC:
It seems to work, E.G. in TT5.16.1 version is set to 5.3, so at first 
5.17 startup migration will be done. However it doesn't work bitween 
beta versions, because 5.4 has already been set as version

CoBC:
@bear101 Ah, yes, you are right. I was going to suggest setting the 
version on first startup if the config key doesn't exist, but I see you 
already did that :), looks like we had the same idea :)

bear101:
@CoBC should we just specify in v5.17 that if no version is set then we assume it's 5.3?

CoBC:
Yes, good idea to at least ensure all migrations for 5.17 will be done.

bear101:
@CoBC  I've added this commit to avoid blank languages in preferences: 4af6fd5c3962e1163815db714badc3e837d7bddb
And set ini-file to 5.0 if it's not specified: 4c5acdc

CoBC:
Also, shouldn't remove old language files? Is this something we can do by Inno Setup during installation? Or maybe do it when migrating to 5.4 config?

bear101:
The language files have to be deleted by the installer because user otherwise do not have administrator rights. I don't know how to do it though in Inno Setup (if it's even supported)

CoBC:
@bear101 We should probably use the [InstallDelete] of Inno Setup script, however I didn't manage to get it working for now.
I think we should however also remove language files in the app if they are present, because for protable install, or on other platforms, we should delete them as well.

bear101:
Notice that the macOS installer has the same issue.

Client/qtTeamTalk/utilui.cpp
```diff
@@ -131,6 +131,42 @@ void migrateSettings()
...
+            "/Thai.qm",
+            "/Turkish.qm",
+            "/Vietnamese.qm"
+        };
+        for (const QString &fileName : oldFiles)
+        {
+            QString filePath = TRANSLATE_FOLDER + fileName;
+            QFile file(filePath);
+            if (file.exists())
+                file.remove();
```
bear101, line 168: You cannot delete these files because they're created by administrator.

CoBC:
Is this page helpful?

macos - osx - remove files as part of pkg installation - Stack Overflow 
<https://stackoverflow.com/questions/35060900/osx-remove-files-as-part-of-pkg-installation>

Client/qtTeamTalk/utilui.cpp
CoBC: Yes, but in portable installation or on Linux we can delete files using this method

bear101: But the portable TeamTalk doesn't have the old language files. It's always a "clean" install.

CoBC: I think that most users copy the files of the new version over the old ones, whether on a portable Windows version or on Linux where the operation is the same. That's why I left this part.
If you think it's unnecessary, we can remove it

bear101:
I remove the commit where old language files are deleted from inside the Qt client.

Issue #2383, opened 2024-06-15
compilation error is not resolved


peter1384:
### Description

@bear101 you've closed the other issue regarding the compilation as completed, but it's not resolved, and I didn't see any commit from you In master about fixing it, and I tryed now again with the steps provided by you yourself, and these are the exact errors I get from vs 2019 when compiling. I don't have any hope that you will investigate this as you normally don't, but lets see anyway, as having compilation error in an app is really important.
1>mainwindow.obj : error LNK2019: unresolved external symbol "public: __cdecl MyTabWidget::MyTabWidget(class QWidget *)" (??0MyTabWidget@@QEAA@PEAVQWidget@@@Z) referenced in function "public: void __cdecl Ui_MainWindow::setupUi(class QMainWindow *)" (?setupUi@Ui_MainWindow@@QEAAXPEAVQMainWindow@@@Z)
1>preferencesdlg.obj : error LNK2001: unresolved external symbol "public: __cdecl MyTabWidget::MyTabWidget(class QWidget *)" (??0MyTabWidget@@QEAA@PEAVQWidget@@@Z)
1>useraccountdlg.obj : error LNK2001: unresolved external symbol "public: __cdecl MyTabWidget::MyTabWidget(class QWidget *)" (??0MyTabWidget@@QEAA@PEAVQWidget@@@Z)
1>release\\TeamTalk5.exe : fatal error LNK1120: 1 unresolved externals
1>Done building project "TeamTalk5.vcxproj" -- FAILED.
========== Build: 0 succeeded, 1 failed, 0 up-to-date, 0 skipped ==========

### Application

- [X] qtTeamTalk
- [ ] TeamTalkAndroid
- [ ] iTeamTalk
- [ ] TeamTalkServer

peter1384:
for now don't do anything, I think I know what corentin forgot to do that's causing this. If it was the case, I'll open a pr to fix this.

peter1384:
This is now worce

PR #2385, opened 2024-06-15
fixed compilation error

peter1384:
when corentin added the mytab files, he forgot to update the mainwindow.pri file and add the mytab cpp and h files in it, and that's why we got those compilation errors. It now has been fixed. So from now on, please when you add or remove a file, take care in mainwindow.pri also, So we don't get compilation errors Thanks

### DIFF
--- a/Client/qtTeamTalk/mainwindow.pri
+++ b/Client/qtTeamTalk/mainwindow.pri
@@ -24,7 +24,7 @@ HEADERS    = mainwindow.h preferencesdlg.h uservideowidget.h \
              serverlogeventsmodel.h textmessagecontainer.h \
              useraccountsmodel.h encryptionsetupdlg.h utiltt.h utilxml.h \
              utilos.h serverdlg.h moveusersdlg.h useraccountdlg.h soundeventsmodel.h \
-             mytableview.h
+             mytableview.h mytabwidget.h
 
 SOURCES    = main.cpp mainwindow.cpp preferencesdlg.cpp uservideowidget.cpp \
              channelstree.cpp channeldlg.cpp userinfodlg.cpp \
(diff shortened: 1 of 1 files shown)


peter1384:
fixes https://github.com/BearWare/TeamTalk5/issues/2383

CoBC:
@bear101 As we previously discussed, QMake is deprecated, and mainwindow.pri should now contains only files relevant for translations, if I remember correctly.
So, I suggest to update readme.md with instructions to build using CMake instead of QMake.

peter1384:
@CoBC
As I remember, Me And some people wer getting errors with cmake when getting the dependencys, And i'm not using cmake, I'm using qt creater to build teamtalk, and even with that out of the way, I don't know the cmake command to get teamtalk5.vcxproj in qt teamtalk folder, I follow these steps because I use qt creater.
Setup Qt Framework in PATH so we can use *qmake.exe* to generate
project files:

`# set PATH=C:\Qt\6.7.0\msvc2019-64\bin;%PATH%`

Generate project files for TeamTalk Qt-based client:

`# qmake -tp vc CONFIG+=release teamtalk5.pro`

Now open *TeamTalk5.vcxproj* in Visual Studio 2019 and build.
so, If qmake is removed, we'll be in troble. But if you really want to make things simpler for yourself, I'll make mainwindow.pri in a way that you don't need to update the files in it anymore, it will update its self automaticly, We won't call the files names anymore and will remove all of the files names from mainwindow.pri, and just say,
*.h, *.cpp, *.ui, /*.cpp, etc, As you did with the langs in your pr,
This way, We will be able to stil build with qmake, and you won't need to update the files in mainwindow.pri anymore.

peter1384:
cmake will not work for those who use qt creater, So I disagree

bear101:
qmake support has been removed. Please use CMake. The README.md file has been updated in QtTeamTalk folder

peter1384:
heh. I'm sorry for this kind of development

PR #2387, opened 2024-06-15
Switch to QTableView model-based for shortcuts tab

CoBC:
@bear101 There is a strange issue, when double clicking on an item, keycompdlg appears and disappears immediately, if I add a QMessageBox before opening keycompdlg, all works fine.
Do you have any idea?

### DIFF
--- a/Client/qtTeamTalk/preferencesdlg.cpp
+++ b/Client/qtTeamTalk/preferencesdlg.cpp
@@ -92,37 +92,37 @@ PreferencesDlg::PreferencesDlg(SoundDevice& devin, SoundDevice& devout, QWidget 
     connect(ui.statusbarToolButton, &QAbstractButton::clicked, this, &PreferencesDlg::slotConfigureStatusBar);
     connect(ui.updatesChkBox, &QAbstractButton::clicked, this, &PreferencesDlg::slotUpdateUpdDlgChkBox);
     connect(ui.betaUpdatesChkBox, &QAbstractButton::clicked, this, &PreferencesDlg::slotUpdateUpdDlgChkBox);
-    m_TSFVarMenu = new QMenu(this);
-    connect(ui.TSFVarButton, &QPushButton::clicked, this, [this]()
+    connect(ui.TSFVarButton, &QPushButton::clicked, this, [&]()
     {
-        m_TSFVarMenu->exec(QCursor::pos());
+        QMenu tsfVarMenu(this);
+        QHash<QString, QString> tsfVariables = {
+            {"d", tr("The day as a number without a leading zero (1 to 31)")},
+            {"dd", tr("The day as a number with a leading zero (01 to 31)")},
+            {"ddd", tr("The abbreviated day name ('Mon' to 'Sun').")},
+            {"dddd", tr("The long day name ('Monday' to 'Sunday').")},
+            {"M", tr("The month as a number without a leading zero (1 to 12)")},
+            {"MM", tr("The month as a number with a leading zero (01 to 12)")},
+            {"MMM", tr("The abbreviated month name ('Jan' to 'Dec').")},
+            {"MMMM", tr("The long month name ('January' to 'December').")},
+            {"yy", tr("The year as a two digit number (00 to 99)")},
+            {"yyyy", tr("The year as a four digit number.")},
+            {"h", tr("The hour without a leading zero (0 to 23)")},
+            {"hh", tr("The hour with a leading zero (00 to 23)")},
+            {"H", tr("The hour without a leading zero (0 to 23)")},
+            {"HH", tr("The hour with a leading zero (00 to 23)")},
+            {"m", tr("The minute without a leading zero (0 to 59)")},
+            {"mm", tr("The minute with a leading zero (00 to 59)")},
+            {"s", tr("The whole second, without any leading zero (0 to 59)")},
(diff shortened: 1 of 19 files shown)


CoBC:
Note: the issue described in the PR description is due to the release of key space (key release 32) after keycompdlg was opened.
The only way I cna find to fix it is to add a QTimer of E.G. 50ms before opening keycompdlg, but it seems to be an hack more than something else.

bear101:
@CoBC Here's how I fixed the initial-dialog-close issue on macOS: e68b3cd4b96ab5ba39ba99d88be42dc9240018ee

The issue seems to be the same on Windows.

CoBC:
Yes, issue is same on all platforms

bear101:
The issue should be fixed now

CoBC:
Yes, it works

bear101:
@CoBC if I enable push-to-talk key in General-tab then it's not updated in "Shortcuts"-tab and vice versa.

CoBC:
@bear101 Should we remove HOTKEY_PUSHTOTALK from shortcuts model, as it was not present here previously, or fix the issue and let this one in model also?

amirmahdifard:
in my opinion keep push to talk in both places for eas of access, and fix the updating problem.

bear101:
I think we should fix the issues. Push-to-talk key is a Shortcut that works in the same way as the other shortcuts. I know this becomes a bit more complicated to maintain state in Preferences dialog but it's more logical.

CoBC:
I think I've fixed the issue

bear101:
I'll try and fix the conflicts

bear101:
I've fixed the conflicts and rebased with master.

Client/qtTeamTalk/preferencesdlg.cpp
```diff
@@ -1005,35 +952,31 @@ void PreferencesDlg::slotSaveChanges()
...
+        while (ite != m_hotkeys.end())
         {
             saveHotKeySettings(ite.key(), *ite);
 #ifdef Q_OS_WIN32
             TT_HotKey_Register(ttInst, ite.key(), &(*ite)[0], INT32(ite->size()));
 #endif
             ite++;
         }
+        Hotkeys activeHotkeys = 0;
+        for (auto it = m_shortcutsmodel->getShortcuts().begin(); it != m_shortcutsmodel->getShortcuts().end(); ++it)
```
bear101, line 974: This code is crashing. You get a copy of shortcuts in `.begin()` and another copy of shortcuts in` .end()`. This means that iterators will never match.` getShortcuts(`) has to return a const reference:

`const hotkeys_t& ShortcutsModel::getShortcuts()`

bear101:
@CoBC I think colon should be part of the translation string. I'm not sure colon makes sense in Chinese or other non Latin languages: [1af4e06](https://github.com/BearWare/TeamTalk5/pull/2387/commits/1af4e06ce9d33b063eb6366f08326faf40b86811)

bear101:
Hm, I'm seeing some strange behavior:

1. Start Prefences dialog
2. Check "push to talk" in General-tab
3. Setup a hotkey.
4. Click Ok.
5. Start Preferences dialog again.
6. Now "push to talk" is unchecked.

CoBC:
OK for colon in translation strings. Note however that we also have some 
similar strings in other dialogs like server properties, server stats, 
or user info, I didn't include colon in translation when migrating from 
QLineEdit to QLabel to avoid to retranslate strings just to add a colon. 
Should we update all translation strings to include colon?

CoBC:
This strange behaviour should be fixes by 54e3902d9

bear101:
Is there a reason to keep SETTINGS_GENERAL_PUSHTOTALK after introduction of SETTINGS_SHORTCUTS_ACTIVEHKS?

CoBC:
Hmmm, right, probably not. I'll remove it

bear101:
I've made the Key Combination dialog show up which is being configured.

bear101:
I'm still seeing some strange behavior.

1. Set PTT to F5 in General tab
2. Press OK
3. Go back to General tab
4. Change to F6
5. Press OK
6. Now PTT is still listed as F5 in the status bar

CoBC:
Hmm, strange, I don't reproduce it here :(

bear101:
I've fix part of the problem here: [a8d217f](https://github.com/BearWare/TeamTalk5/pull/2387/commits/a8d217fc9c0a00940e1b1779444b6e322d3b3027)

Note the description of std::map::insert() is:
`Inserts element(s) into the container, if the container doesn't already contain an element with an equivalent key.`

bear101:
When the check state is toggled on PTT:
`ui.pttChkBox->setChecked(true);`
Then it triggers an event that reloads PTT:
```
        hotkey_t ptt;
        if (loadHotKeySettings(HOTKEY_PUSHTOTALK, ptt))
            m_hotkeys[HOTKEY_PUSHTOTALK] = ptt;
```

Therefore PTT is always reverted.

CoBC:
Hmmmm, I can't reproduce this one also :(

I tried enabled PTT in general tab, it's enabled correctly, I tried 
disabled in general tab, it's disabled, I tried to uncheck the PTT 
shortcut in shortcuts tab, it appears uncheck and general tab and is 
disabled, and I tried check again in shortcuts tab, it appears check in 
general tab and it's enabled.

So, not sure to understand the issue :(

bear101:
It is the key combination that does not match

PR #2392, opened 2024-06-18
update chinese translation for installer

hejiadong0608:


CoBC:
Unfortunately the language file zh.isl doesn't exists so this will not work.
Maybe we can use the translation available at https://github.com/kira-96/Inno-Setup-Chinese-Simplified-Translation for Chinese and then merge this PR with translated component names

hejiadong0608:
@CoBC  I tried putting this file here, but I'm not sure if it will work.

CoBC:
Yes it works. Can you just use zh_CN instead of chinesesimplified in teamtalk5.iss to be align with other languages?

hejiadong0608:
@CoBC  OK i've finished it.

CoBC:
OK, looks good to me

bear101:
Shouldn't the Chinese translation be submitted to Inno Setup project?
https://github.com/jrsoftware/issrc/blob/main/CONTRIBUTING.md

hejiadong0608:
@bear101 This Inno translation file is not mine, I submit it here just to make the TeamTalk installer support Chinese.

Issue #2393, opened 2024-06-18
Add direct interactions with Orca and Voice Over for TTS Events


CoBC:
### Description

Following implementation of Qt Accessibility Announcement, and after some user tests, it seems notifications send using Qt Accessibility Announcement are real accessibility events, and not really messages without context. So, majority of screen readers ignore those notifications when TT is not in focus.
E.G., with NVDA on Windows or Voice Over on MacOS, enable Qt Accessibility Announcement in preferences, go to your desktop for example and ask someone to login to your server, no message will be read by screen reader. Now, go to TT's window and ask the person to log out, screen reader will read notification.
SO, finally, Qt Accessibility Announcement is not really a good solution for our TTS events.
I'll try to implement direct interactions with Voice Over (via Apple Script) and with Orca (using dBus).
I guess we can keep Qt Accessibility Announcement available as a backup in case it can be useful.

### Application

- [X] qtTeamTalk
- [ ] TeamTalkAndroid
- [ ] iTeamTalk
- [ ] TeamTalkServer

amirmahdifard:
yes please, do not remove qt accessibility announcement, I need it. Ad the thing that you want but don't remove qt accessibility announcement, at least on windows

PR #2396, opened 2024-06-19
Add Apple Script support for TTS Events

CoBC:
#2393


Client/qtTeamTalk/utiltts.cpp
```diff
@@ -134,6 +134,17 @@ void addTextToSpeechMessage(const QString& msg)
         QAccessible::updateAccessibility(&announcementEvent);
 #endif
         break;
+    case TTSENGINE_APPLESCRIPT:
+        QString appleScript = QString(R"(
+            tell application "VoiceOver"
+                output "%1"
+            end tell
+        )").arg(msg);
```
bear101, line 145: what happens if "msg" contains quotes? Should they be escaped with \" ?

PR #2398, opened 2024-06-20
Set TeamTalk Sound Device as AudioDevice for QSoundEffect

### LINKED ISSUE #2397: Windows QT client: TT wav sounds are always played on the default device
### Description

Hello,
The problem started when I noticed that notification sounds such as when voice activation was turned on or when a user joined a channel were not always played in TeamTalk QT Windows.
After investigating the problem, it turned out that the sounds always go through the default playback device.
Example:
I have Creative SoundBlaster as my default playback device in Windows. In teamtalk, in the sound system settings, I set the output device to Realtek speakers. The call goes through Realtek, but sounds are still played by SoundBlaster. When you change Realtek to the default sound card, sounds are played through that sound card.

### Application

- [X] qtTeamTalk
- [ ] TeamTalkAndroid
- [ ] iTeamTalk
- [ ] TeamTalkClassic
- [ ] TeamTalkServer

### Platform

- [X] Windows
- [ ] macOS
- [ ] Android
- [ ] iOS
- [ ] Linux

### Expected behavior

_No response_

### Actual behavior

_No response_

1. Set whatever the default sound card is in your system.
2. In preferences/sound system, set the output device to a different sound card that is not the default one. Teamtalk sounds should play through the default sound card.

CoBC:
Fixes #2397

### DIFF
--- a/Client/qtTeamTalk/preferencesdlg.cpp
+++ b/Client/qtTeamTalk/preferencesdlg.cpp
@@ -629,8 +629,8 @@ void PreferencesDlg::initSoundEventsTab()
     connect(ui.spackBox, static_cast<void (QComboBox::*)(int)>(&QComboBox::currentIndexChanged), this, &PreferencesDlg::slotSPackChange);
     ui.sndVolSpinBox->setValue(ttSettings->value(SETTINGS_SOUNDEVENT_VOLUME, SETTINGS_SOUNDEVENT_VOLUME_DEFAULT).toInt());
     ui.sndeventPlaybackComboBox->addItem(tr("Default"), PLAYBACKMODE_DEFAULT);
-    ui.sndeventPlaybackComboBox->addItem(tr("One by One (TeamTalk Sound Device)"), PLAYBACKMODE_ONEBYONE);
-    ui.sndeventPlaybackComboBox->addItem(tr("Overlapping (TeamTalk Sound Device)"), PLAYBACKMODE_OVERLAPPING);
+    ui.sndeventPlaybackComboBox->addItem(tr("One by One"), PLAYBACKMODE_ONEBYONE);
+    ui.sndeventPlaybackComboBox->addItem(tr("Overlapping"), PLAYBACKMODE_OVERLAPPING);
     setCurrentItemData(ui.sndeventPlaybackComboBox, ttSettings->value(SETTINGS_SOUNDEVENT_PLAYBACKMODE, SETTINGS_SOUNDEVENT_PLAYBACKMODE_DEFAULT));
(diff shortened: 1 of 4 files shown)


Client/qtTeamTalk/preferencesdlg.cpp
```diff
@@ -625,8 +625,8 @@ void PreferencesDlg::slotTabChange(int index)
         connect(ui.spackBox, static_cast<void (QComboBox::*)(int)>(&QComboBox::currentIndexChanged), this, &PreferencesDlg::slotSPackChange);
         ui.sndVolSpinBox->setValue(ttSettings->value(SETTINGS_SOUNDEVENT_VOLUME, SETTINGS_SOUNDEVENT_VOLUME_DEFAULT).toInt());
         ui.sndeventPlaybackComboBox->addItem(tr("Default"), PLAYBACKMODE_DEFAULT);
-        ui.sndeventPlaybackComboBox->addItem(tr("One by One (TeamTalk Sound Device)"), PLAYBACKMODE_ONEBYONE);
-        ui.sndeventPlaybackComboBox->addItem(tr("Overlapping (TeamTalk Sound Device)"), PLAYBACKMODE_OVERLAPPING);
+        ui.sndeventPlaybackComboBox->addItem(tr("One by One"), PLAYBACKMODE_ONEBYONE);
```
bear101, line 628: The reason I've written "(TeamTalk Sound Device)" is because it's played on your selected sound device. Unlike "Default" which is played on your OS selected sound device.

bear101: Maybe the "Default" changed to "OS Selected".

Client/qtTeamTalk/utilsound.cpp
```diff
@@ -630,6 +669,7 @@ void PlaySoundEvent::playDefaultSoundEvent(const QString& filename)
     static QSoundEffect* effect = nullptr;
     delete effect;
     effect = new QSoundEffect(ttSettings);
+    effect->setAudioDevice(getSelectedOutputAudioDevice());
```
bear101, line 672: What happens if getSelectedOutputAudioDevice() returns QAudioDevice() ?

Client/qtTeamTalk/preferencesdlg.cpp
CoBC: The default now uses also TeamTalk output device, so it's why I found this precision a bit useless.

Client/qtTeamTalk/utilsound.cpp
CoBC: I've rollback to default output device in this case

amirmahdifard:
@CoBC now that all of them are using teamtalk sound device, rename default to interupt priviace, as teamtalk classic, So people will know what each playback mode does. Thanks

PR #2400, opened 2024-06-21
SpamBot ban duration/expiration

bear101:


amirmahdifard:
@bear101 don't you think it is better to implement ban duration for teamtalk server to also add a long standing issue, and then in spambot, you only tell that if you are banning someone, set a unban time for it, and then the server will unban that in that time, So the spambot doesn't have to do unban job. Then, In qt client, We can add a new editbox in banned users dialog and ban user dialog, And call it unban time. If any time is specifyed in that box, everytime the banned user tryed to login, the server will compare the unban time with the current time, And if current time is same or grater than the unban time, unban the user and let them login, And if we left the unban time editbox blank, then that's a permanent ban.

PR #2410, opened 2024-07-04
Introduce media file tab for controlling playback

bear101:


bear101:
@CoBC can you take a look and see if there's still many obvious bugs?

CoBC:
The volume slider seams to have no effect, when I change his value, it just returns to 1000
Also some QLabel (E.G. format) has no accessible label asociated

bear101:
@CoBC were you using SpeexDSP when volume had no effect? I've changed volume slider to be disabled if AGC is not enabled in SpeexDSP.

I've also changed the media playback off slider to range from 0-10000 so one can set it more precisely. The best solution would be to change its range depending on the length of the media file.

CoBC:
I use TeamTalk Audio preprocessor. Volume slider still seams to have no effect :(, and the "elapsed" slider don't refresh his value, so, when I click on "Play" button, "Elapsed" slider just became active but with a value of 0

bear101:
If you use the Stream Media file dialog isn't it the same issue with the "Elapsed" slider?

CoBC:
Yes, same issue

bear101:
@CoBC do the labels have to changed in the same as a "user information dialog" in order to be readable for screen readers?

CoBC:
Yes :)

CoBC (approved):

PR #2412, opened 2024-07-10
add doSendPrivateMessage

kirill-jjj:
added this function to make it easier for new developers to write the code


bear101:
I suggest you add this function to your own util-library. The idea of TeamTalk5.py is that it should reflect the C-API (TeamTalk.h file).

PR #2420, opened 2024-08-02
Display IP address in user info if USERRIGHT_BAN_USERS

CoBC:
Fix a bug reported by user after change to QLabel for User Info dialog

### DIFF
--- a/Client/qtTeamTalk/userinfodlg.cpp
+++ b/Client/qtTeamTalk/userinfodlg.cpp
@@ -106,7 +106,7 @@ void UserInfoDlg::updateUser()
     if(ui.usertypeLabel->text() != usertypeLabelText)
         ui.usertypeLabel->setText(usertypeLabelText);
 
-    if (TT_GetMyUserType(ttInst) & USERTYPE_ADMIN || TT_GetMyUserID(ttInst) == user.nUserID)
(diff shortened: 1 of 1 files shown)


amirmahdifard:
@CoBC if you remember, bear101 told you befor, so you only need to check if userright ban users, and not also admin because if you only check that it will also work for admins too, so you don't need to check both

Client/qtTeamTalk/userinfodlg.cpp
```diff
@@ -106,7 +106,7 @@ void UserInfoDlg::updateUser()
     if(ui.usertypeLabel->text() != usertypeLabelText)
         ui.usertypeLabel->setText(usertypeLabelText);
 
-    if (TT_GetMyUserType(ttInst) & USERTYPE_ADMIN || TT_GetMyUserID(ttInst) == user.nUserID)
+    if (TT_GetMyUserID(ttInst) == user.nUserID || TT_GetMyUserRights(ttInst) & USERRIGHT_BAN_USERS)
     {
         QString ipLabelText = QString(tr("IP-address") + ": %1").arg(_Q(user.szIPAddress));
```
bear101, line 111: Isn't it easier to check if `_Q(user.szIPAddress).size() > 0` ?

PR #2424, opened 2024-08-10
Replace radio button by comboboxes in some dialogs

CoBC:
An user request was made to replace radio button by comboboxes.
@bear101 Is it OK? If yes, I can continue

### DIFF
--- a/ChangeLog.txt
+++ b/ChangeLog.txt
@@ -11,6 +11,7 @@ Default Qt Client
 - Fixed state of checkable items in "Preferences", "User Account", "Server properties" dialogs not announce by Voice Over on MacOS
 - Fixed labels of edit fields in "Custom Video Format" dialog
 - Replaced readonly edit fields by labels in "User Info" and "Server Statistics" dialogs
+- Radio buttons for Gender, Sound System and Status mode changed to combo boxes
 - Context menu to move banned/unbanned users to banned/unbanned list in "Banned Users" dialog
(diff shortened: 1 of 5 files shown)


bear101:
I'm fine with this change but maybe we should release what we have now before merging it. What do you think?

One annoying thing at the moment is that Qt has mistakenly specified Qt 6.8beta2 as Qt 6.9 on Windows (if you check About box).

CoBC:
OK to release current changes before merging new ones.

Qt 6.8 Beta 3 should be release tuesday: 
https://wiki.qt.io/Qt_6.8_Release so we can maybe release the new 
release in this new beta which hopefuly will fix this mistake.

amirmahdifard:
my opinion is to also inclood only this change in 5.17 because our focus is on accessibility, and this change makes the accessibility of selections of these dialogs better. my opinion

bear101:
@CoBC if you select DirectSound in Preferences and press Ok and then return to Sound System in Preferences dialog. Then sound devices will appear twice.

CoBC:
@bear101 Qt 6.8 Beta 3 has been released. So I think you can update to 
this beta and than release the V5.17

bear101:
@CoBC I've created the v5.17 release and uploaded it here: https://bearware.dk/teamtalk/v5.17/

There's some documentation that needs to be updated but I think we should release anyway. Then we can update the docs when we have more feedback.

CoBC:
@bear101 Yes, I think we can update docs later. We just have to remember 
this point :)

PR #2428, opened 2024-08-13
Add 3D Positionning Support

CoBC:
This is the last feature we have to retrieve from classic client for interested users with compatible devices.
Unfortunately my devices seams to be incompatible with 3D support, so it's a bit difficult to test it :(

### DIFF
--- a/Client/qtTeamTalk/preferencesdlg.cpp
+++ b/Client/qtTeamTalk/preferencesdlg.cpp
@@ -1180,7 +1180,9 @@ void PreferencesDlg::slotSoundOutputChange(int index)
         devinfo += ". ";
         devinfo += tr("Sample Rates:");
         for(int i=0;dev.outputSampleRates[i]>0 && i < TT_SAMPLERATES_MAX;i++)
-            devinfo += " " + QString::number(dev.outputSampleRates[i]);
+            devinfo += ", " + QString::number(dev.outputSampleRates[i]);
+        devinfo += ". ";
+        devinfo += tr("3D Positionning support: ") + QString(isOutputDevice3DSupported(dev)?tr("Yes"):tr("No"));
     }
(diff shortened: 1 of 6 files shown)


CoBC:
Do you have any idea to support it? Maybe it was not working for a while 
also in classic client when it was used. However, don't know if a lot of 
users use this feature :)

amirmahdifard:
I also suggest to wait and merge this pr when it has been finished, and then releace 5.17, and announce that 5.17 is completely the death of teamtalk classic, and from that point, teamtalk classic's vertion will no longer be updated, and the installer of teamtalk will  no longer update teamtalk classic as it currently does if you already have the older vertion of classic client installed on your system, which is a bug and shouldn't happen!

beqabeqa473:
I don't think this worked on newer windows versions from windows vista. I think 3d positioning stopped working in versions after windows xp. The problem is that microsoft rewrote whole sound system and to make it work again i think there should be additional tools required.

CoBC:
@amirmahdifard Really not sure to understand your point, classic client is already officially abandoned for a while now, and Windows installer no longer include it since 5.14 release. If you still have classic client install, only the library will be updated during update, but, anyway, classic executable is no longer updated.
I only mentioned classic client in my PR description to note that this feature was available previously in this client.
Now, we have to determine if it's really difficult to rewrite code to support this feature again or not, however if it never worked after Windows XP, probably there is a bit of refactor to write to get it working again.

amirmahdifard:
@CoBC trust me, it stil does. Once I saw someone was in a server with teamtalk classic 5.16, and I asked him, and he said if you first install an older vertion of teamtalk classic such as 5.13 or 5.12,, then if you launch the new installer which doesn't support classic, and you update your qt teamtalk with it, it will also update your classic too. Also, I don't know why bear101 stil updates classic's vertion as wel up on his vertion update commits

PR #2434, opened 2024-08-16
Correctly ask for language selection at first start

CoBC:
migrateSettings is call before language selection, if TT has never been started, migrateSettings define a lanauge, so SETTINGS_DISPLAY_LANGUAGE is always set and the language selection isn't trigger.
So, I changed the check to ask for language selection if it is the first startup and not only if SETTINGS_DISPLAY_LANGUAGE isn't present, and I add a check to ensure migrateSettings define a new language only if SETTINGS_DISPLAY_LANGUAGE is previously set.

### DIFF
--- a/Client/qtTeamTalk/preferencesdlg.cpp
+++ b/Client/qtTeamTalk/preferencesdlg.cpp
@@ -497,6 +497,11 @@ void PreferencesDlg::initDisplayTab()
     }
     QString lang = ttSettings->value(SETTINGS_DISPLAY_LANGUAGE, SETTINGS_DISPLAY_LANGUAGE_DEFAULT).toString();
     int index = ui.languageBox->findData(lang);
+    if (index < 0)
+    {
+        QString langPrefix = lang.section('_', 0, 0);
+        index = ui.languageBox->findData(langPrefix);
+    }
     if (index >= 0)
         ui.languageBox->setCurrentIndex(index);
 
--- a/ChangeLog.txt
+++ b/ChangeLog.txt
@@ -11,6 +11,7 @@ Default Qt Client
 - Ability to restore all custom TTS and status bar events to default values
 - Fixed gender changed from female to male when migrating to 5.17 release
 - Fixed sound events disabled when migrating to 5.17 release
+- Fixed language not set at first start
 Android Client
 -
 iOS Client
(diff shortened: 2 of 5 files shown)


CoBC:
When language is initially set, Qt returns language code as something like "fr_FR" for French, so, 9c6970f add a partial check to correctly select languageBox item which is corresponding (E.G. fr for fr_FR)

Client/qtTeamTalk/preferencesdlg.cpp
```diff
@@ -489,6 +489,19 @@ void PreferencesDlg::initDisplayTab()
     }
     QString lang = ttSettings->value(SETTINGS_DISPLAY_LANGUAGE, SETTINGS_DISPLAY_LANGUAGE_DEFAULT).toString();
     int index = ui.languageBox->findData(lang);
+    if (index < 0)
+    {
+        QString langPrefix = lang.section('_', 0, 0);
```
bear101, line 494: What is the purpose of this code? Why is the language split using "_"?

CoBC: E.G. when I start TT on my system with a blank config, SETTINGS_DISPLAY_LANGUAGE is set to fr_FR, so if I open and than close preferences, language in reset to default because fr_FR is not corresponding to an item in the combobox, so we should check for a partial item

bear101: Doesn't this cause problems with pt_BR and pt_PT? Which one does it choose?

CoBC: Partial check is only done if no item is fully equal to settings, so if you put pt_BR in the config file "portuguès" will be selected in the combo, and for pt_PT it will be "Portuguès europeu". Same for Chinese with zh_CN and zh_TW

bear101: Ok, can't you use `ui.languageBox->findData(langPrefix);` instead of the for-loop?

CoBC: Done

bear101:
Hm, I'm not having much luck with this change. When I start TeamTalk on macOS then it tells me my system language is English and I choose to continue in English. Then it tell me that it cannot load en_DK. What does that mean?

bear101:
Hm mainwindow.cpp:893 is not being executed on macOS

bear101:
I guess mainwindow.cpp:893 should be moved to mainwindow.cpp:848

PR #2435, opened 2024-08-16
Add username variable for custom TTS and status events for user logged in/out

CoBC:


CoBC:
Only classroom changes aren't supported

PR #2437, opened 2024-08-16
Only migrate settings if first start=false

CoBC:
When migrateSettings() is call with a blank ini file, a lot of settings aren't incorrectly set to empty values, E.G. sound events. Also some keys aren't incorrectly redefine, E.G. gender define to female instead of neutral.

### DIFF
--- a/ChangeLog.txt
+++ b/ChangeLog.txt
@@ -4,7 +4,7 @@
 
 Version 5.17.1, unreleased
 Default Qt Client
--
+- Fixed a lot of settings keys not defined correctly on new config
 Android Client
 -
 iOS Client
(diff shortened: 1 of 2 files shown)


CoBC:
@bear101 it seams some config keys are also migrate incorrectly in existing config, E.G. all sound events are disabled when running 5.17 on an existing 5.16.1 config.
I don't really understand why, do you have an ideA? I think we should fix this and than release a 5.17.1 because it's a major bug for a lot of users.

bear101:
With this change the following code is not called in migrate settings:
```
    if (ttSettings->value(SETTINGS_GENERAL_VERSION).toString() != SETTINGS_VERSION)
        ttSettings->setValue(SETTINGS_GENERAL_VERSION, SETTINGS_VERSION);
````
Also SETTINGS_GENERAL_FIRSTSTART was first introduced in v5.8 so settings prior to that will not be migrated.

CoBC:
Ah, yes, you're right. also I guess all other fixes are sufficient, so closing this PR

PR #2448, opened 2024-08-24
Option to disable OS information sending

CoBC:


bear101:
I think this configuration option is unnecessary. If we go into such details then it will be impossible to find the useful configurations.
Already you can determine the OS by looking at the client name. At least for mobile vs. desktop.

CoBC:
OK so just keep as it is now.

PR #2450, opened 2024-08-26
Simplify text of user rights

CoBC:


amirmahdifard:
but why change status right is not in the user rights list?

PR #2451, opened 2024-08-27
Fixed VoiceOver not announcing the user's gender and whether a channel is password-protected

Muamalaljanahi:


bear101:
So accessibilityLabel takes over what is specified in Main.strings?

Also why do you remove "Channel. %d users" ?

Muamalaljanahi:
> 
> Also why do you remove "Channel. %d users" ?

I removed it because VoiceOver was announcing "Channel. %d users" instead of saying that the channel is "password protected" or "no password"

PR #2458, opened 2024-09-17
Convert all RX/TX values to local format

CoBC:
In mainwindow status bar and server statistics dialog
Also rename getFormattedFileSize to getFormattedSize in utilUI


bear101:
good idea

PR #2459, opened 2024-10-02
Fix language selection and first start saving

CoBC:


Client/qtTeamTalk/mainwindow.cpp
```diff
@@ -712,8 +712,17 @@ void MainWindow::loadSettings()
             this->ui.retranslateUi(this);
         else
         {
-            QMessageBox::information(this, tr("Translate"),
-                QString("Failed to load language file %1").arg(lang));
+            QString langPrefix = lang.section('_', 0, 0);
+            if (switchLanguage(langPrefix))
```
bear101, line 716: I think I've asked this before. But doesn't this change to "langPrefix" strip "_BR" and "_PT" from Portuguese language file?

CoBC: Ah yes, when you asked previously it was not the case in the previous context but you're right for here. I'll look into it tomorrow

CoBC:
It works as expected, because if the first language detected (E.G. pt_BR or pt_PT) is find in language folder, nothing else is done, but, if the detected language is E.G. en_DK, it will not be find in language folder so en will be search and find.

PR #2495, opened 2024-12-04
Also convert units when speaking client stat

CoBC:


amirmahdifard:
@CoBC sadly, for me when I use this, screen reader only announces the rx and the tx, and doesn't announces the ping. now that I've see this code, I want to ask a question from you. is it possible to remove this if ping greater than blaablaa, so it always announces the ping, even if it's 0 or whatever, then at least it will always announces the ping for everyone, because it doesn't have to first detect that statement, and when we use this option, it wil defenetly announce the current ping

PR #2517, opened 2025-01-25
updated the Korean translations of qtTeamTalk

ajackpot:
I updated the Korean translations of qtTeamTalk to be compatible with the latest version.
Thank you.


bear101:
Thanks a lot!

PR #2533, opened 2025-03-03
Android client. Fixed select for move option

amirmahdifard:
Currently, as much as the select for move option is clicked, it selects the user, which results the app trys to move that user 20 times to the destination channel. But now, like windows clients, once clicking this option will select the user and change the title to deselect for move for that user, and clicking it again will deselect that user.


amirmahdifard:
@bear101 I've openned these, And I'll open new prs again when these are merged. Thanks

bear101:
Thanks, I've merged to master.

amirmahdifard:
@bear101 Thanks! please also merge my other pull request for android client as wel so I can open new ones quickly. Thanks!

PR #2535, opened 2025-03-05
Fix crash issue when toggling echo cancellation off

bear101:
When duplex mode is disabled (due to echo cancellation = off) then WebRTC preprocessor is still active in echo cancellation mode. WebRTC then tries to access an AudioFrame::output_samples which is no longer available.

In other words the WebRTC preprocessor is configured in a way that is incompatible with the sound device configuration. WebRTC can only echo cancel when sound devices are initialized with
TT_InitSoundDuplexDevices().


ChrisDuffley:
Thank you @bear101 for finally fixing this as it's been persisting for a very long time now.

Issue #2536, opened 2025-03-05
android client. UI improvements


amirmahdifard:
### Description

Currently, a remove button exists beside each server, and long pressing a server just brings up the edit dialog for that server. That remove button should be removed from that playse, and long pressing on a server should open a popup dialog contaning 2 options, edit or remove the server. This makes the screen much better for everyone, also for screenreader users. I'll try to do this.

### Application

- [ ] qtTeamTalk
- [x] TeamTalkAndroid
- [ ] iTeamTalk
- [ ] TeamTalkServer

bear101:
Do you plan to fix this? Otherwise I'll move it to Discussions' feature requests.

amirmahdifard:
I already opened a pr and fixed this and you merged it already, so closing as completed.

PR #2539, opened 2025-03-08
Android client. Recoded server list dialog

amirmahdifard:
before, a remove button was existed beside all servers, and long pressing a server would edit that server. that way, screenreader users needed to swipe twice to reach the next server, so for example, to find 20th server, we needed to swipe 40 times, because there was a remove button beside every server. But now, exactly like the desktop qt client's host manager that was recoded in 5.17, I recoded android client's server list to make it use popup menus, modern, and much more cleaner and clear.
This fixes
https://github.com/BearWare/TeamTalk5/issues/2536

### DIFF
--- a/ChangeLog.txt
+++ b/ChangeLog.txt
@@ -20,6 +20,7 @@ Default Qt Client
 - Fixed language not set at first start
 - Fixed OPUS DTX mode sending too big packets when variable bitrate mode (VBR) was disabled
 Android Client
+- Redesigned server list dialog to remove the old remove server button and make it to use popup menus, results in a more clear and modern dialog, and fast navigating for screenreader users
 - IP Address and status message is now shown in User Properties dialog
 - Ability to select/deselect user for move to other channel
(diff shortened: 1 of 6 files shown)


amirmahdifard:
@bear101 hello. please merge this pr so I can do my other contributions, which I want to finish them all befor my hollydays starts, because our new year hollydays is 2 weeks long and will start with in around 10 days or so. So until then I want to contribute some more things, so we have a better teamtalk android by then. thanks so much!

CoBC:
Can you stop to always request to merge yours PR? Can you just wait like 
all other contributors? It's really ennoying, nobody is at your 
disposition here!

amirmahdifard:
@CoBC it would be nice if you respect others. I just do not have access to request a review like you, that's why.

jeremyp3:
+1 with @CoBC 
you don't need to mention anyone in your pr. the people involved already know about your pr. don't worry.

bear101:
Thanks, it's merged now.

PR #2542, opened 2025-03-12
Toast notifications support

CoBC:


Client/qtTeamTalk/CMakeLists.txt
```diff
@@ -34,6 +34,13 @@ endif()
 # TextToSpeech was introduced in Qt 6.4
 find_package (Qt6 COMPONENTS Widgets Xml Network Multimedia TextToSpeech)
 
+if (WIN32)
+  include_directories(${CMAKE_CURRENT_SOURCE_DIR}/3rdparty/WinToast)
+  set(WINTOAST_SOURCES
+    3rdparty/WinToast/wintoastlib.cpp
```
bear101, line 40: wintoastlib.h is missing

Client/qtTeamTalk/settings.h
```diff
@@ -421,6 +418,10 @@
 #define SETTINGS_TTS_ASSERTIVE                         "texttospeech/assertive"
 #define SETTINGS_TTS_ASSERTIVE_DEFAULT                 false
 #endif
+#ifndef Q_OS_DARWIN
```
bear101, line 421: Please stick to `#if defined` instead of `#ifdef`. `#ifdef` cannot be or'ed

Client/qtTeamTalk/utiltts.h
```diff
@@ -96,13 +96,10 @@ enum TextToSpeechEngine
     TTSENGINE_NONE = 0,
     TTSENGINE_QT = 1,
     TTSENGINE_TOLK = 2,
-    TTSENGINE_NOTIFY = 3,
```
bear101, line 99: Better to rename `TTSENGINE_NOTIFY` to `TTSENGINE_NOTIFY_OBSOLETE`, so we don't reduce the value `3`.

Client/qtTeamTalk/preferences.ui
```diff
@@ -1576,7 +1576,7 @@
      </widget>
      <widget class="QWidget" name="ttsTab">
       <attribute name="title">
-       <string>Text To Speech</string>
+       <string>Notification</string>
```
bear101, line 1579: How come you want to call it "Notification" instead of "Text To Speech". "Sound Events" are also sort of "Notification".

PR #2548, opened 2025-03-17
Chat history in a list

CoBC:
@bear101 Can you take a look and remind me of the problems with this change?


valiant8086:
not sure if I grasp why the change to a new issue, but I still think this is a useful change for screen reader users, and I'd like to think any downsides to it can be fixed for the sighted users as well.

bear101:
With this change to QListWidget then it has the following issues:

- Select words for copy/paste. E.g. if someone writes an email or URL
- It is not possible to see cursor indicator for selection
- If a message contains two URLs then you can only open the first one
- Lines are not wrapped so text has to be scrolled

Making this option configurable is probably a bit tricky but it should be possible with inheritance. So a ChatTextEdit based on QListWidget and another based on QPlainTextEdit.

valiant8086:
really, they need to fix the edit field anyway to get it to read properly with screen readers. It's a hot mess right now very close to impossible to actually use it. If yall can get the list view thing an option we can switch to, I would switch to it anyway because I think it's more practical as a screen reader user, but even so, the edit field behaving the way it does might be something qt can fix or is it an issue with implementation here.

Thanks for considering this.

CoBC:
The issue with edit field is related to Qt
I'll try to fix issues with list

beqabeqa473:
I still vote for an option to make it configurable.
I don't see a problem of two implementations here and there is no impossibility to do that.

valiant8086:
Yes, I have a feeling that may be where this ends up. In any case, some sighted users might prefer the list if they don't think they'll have many long messages, and want the new way to select messages. It'll also be that some screen reader users will want the readonly if they ever fix it to read and navigate properly.

While I don't want this pr to go by the wayside, don't we have other applications that use qt edit, and do they have the same bugs with navigation? Is this filed with the qt developers?

amirmahdifard:
onistly i'm tired of qt's read only bugs. I'm actually wondering how some people are using the read only editbox that aren't facing its bugs, So I can use it the same way as them. Currently I don't remember when did I touched teamtalk's messy read only history last time. I'm even using nvda's speech history addon to read people's private messages and to chat with them. but that addon is also annoying because you'll loos your item focuse while holding shift and nvda speecks a new thing, but it's at least better than qt's read only editboxes!  After merge of this pr, chatting in tt will be wunderful and enjoyable! lets just chat constantly 2 days!

bear101:
@CoBC I'll work a bit on this one and do a force push

Issue #2545, opened 2025-03-14
android client. Add the ability to export each server in to .tt file


amirmahdifard:
### Description

Currently, an option exist to export all servers in to a .tt file, but sometimes we have a server and we want to get a tt file from that server. now, i've added this feature, but it's pending. i'm waiting for my current pull request to be merged, then i'll provide the feature in a pull request. I'm doing this to make sure everything continues clean, with out conflicts, etc. it's better in my opinion

### Application

- [ ] qtTeamTalk
- [x] TeamTalkAndroid
- [ ] iTeamTalk
- [ ] TeamTalkServer

beqabeqa473:
please do things without over-explaining things. Base your work on master, everything will be handled by reviewers/mergers. No need for explaining things which are not interesting at all

PR #2549, opened 2025-03-17
android client. Ability to export each server only

amirmahdifard:
Currently, in the more options popup menu we have an option to export all server list in a tt file. But what if we needed to export only one server in to a tt file seperatly and not others? the feature is now hear. fixes https://github.com/BearWare/TeamTalk5/issues/2545 also, a bit of code cleanup that I've forgot to do before. Enjoy!


amirmahdifard:
@bear101 can you please don't releace 5.18 out yet? because there are stil features that I want to develop and after that we can releace. Thanks!

Client/TeamTalkAndroid/src/main/java/dk/bearware/gui/ServerListActivity.java
```diff
@@ -275,7 +275,6 @@ protected void onActivityResult(int requestCode, int resultCode, Intent data) {
                     if (entries != null) {
                         for (ServerEntry entry : entries) {
                             entry.servertype = ServerEntry.ServerType.LOCAL;
-                            entry.rememberLastChannel = true;
```
bear101, line 278: What issue are you fixing here?

amirmahdifard:
never mind, the remember last channel's value is not even saved in the tt file. channel and the channel password is saved, this remember channel is the only thing that is not saved. @bear101 should we also save this value when exporting tt files from servers, so we can keep and respectt user's wishes, raather than forcefully disabling or enabling something regardless of what the user has set up?

Client/TeamTalkAndroid/src/main/res/xml/pref_serverentry.xml
```diff
@@ -103,7 +103,6 @@
             android:disableDependentsState="true"
             android:persistent="false" />
         <EditTextPreference
-            android:defaultValue="/"
```
bear101, line 106: Why this change? Shouldn't users join root channel by default?

amirmahdifard:
@bear101 When the "last joined channel" checkbox is unchecked, the client will join you to the root channel even when the value is empty. I personally tested this. It was very annoying for some users who wanted to disable this feature. We had to uncheck the box, and also, clear the value. But now, the following is happening. When the box is checked, it automaticly updates the value to your last channel, as expected. When the box is not checked, you'll join any channel you specify in the channel value yourself. Empty which is default will join to root channel, /public channels/audio playing/ will join that channel, etc.

amirmahdifard:
I think tomorro i'll be away for our new year vaykation, So I only can do work today

PR #2557, opened 2025-03-26
Ignore text size limitation for screen readers

CoBC:


CoBC:
We will have a conflict with #2556, so one PR will have to be modify before all can be merge

PR #2560, opened 2025-04-02
Configurable Chat History as List view

bear101:
This is a continuation of #2548.

@CoBC there's a bug in the text message popup dialog so there's an edit field appearing. Any idea why this is happening?

### DIFF
--- a/ChangeLog.txt
+++ b/ChangeLog.txt
@@ -14,6 +14,7 @@ Default Qt Client
 - Last login time shown in "User Accounts" dialog
 - Ability to sort by "Last Login Time" in "User Accounts" dialog
 - Ability to send toast notifications on Linux and Windows
+- Ability to change chat history to list view in "Preferences" dialog
 - Improve "Server Properties" dialog accessibility
 - WebRTC updated from r4332 to r6818
 - Qt updated to 6.9.0beta3 on macOS and Windows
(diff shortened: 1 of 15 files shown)


CoBC:
I can't start the app after switch to list view :(, so I can't test the dialog for now :(

bear101:
Hm, strange. I don't see any crashes. I've tested with Qt 6.9.0 on macOS with VoiceOver and Windows 10 with Narrator.

CoBC:
I never know how to debug. Do you have a way to see what's happened? I don't have any UI, simply nothing happened when I launch teamtalk5.exe :(

jeremyp3:
I have the same crash with linux under qt6. I haven't tested with qt5.

as soon as I restart TeamTalk after checking the box, crash segfault

bear101:
can you try to temporarily disable the screen reader while TeamTalk is starting. I wonder if the screen reader crashes the application because I delete some controls.

CoBC:
I just tried and the result is the same :(

CoBC:
OK, it works.
I confirm the issue with readonly edit field even if checkbox to use list is checked...

bear101:
@CoBC I've fixed the edit field issue. I don't understand why my new code is working and the old one doesn't

valiant8086:
I like what I'm hearing so far. Is there a way to test this without source or do we need to wait till it hits main branch for snapshots?

bear101:
@CoBC the removal of the widgets in mainwindow should now work the same as textmessagedlg.

CoBC (approved):

CoBC:
Yes, good idea

CoBC:
@bear101 Hmmm, not seen during test, but the list is inserted at the end of chat tab, so the message edit field is appearing before the history...

PR #2562, opened 2025-04-05
Android. remember last channel value saved on tt files

amirmahdifard:
from now on, when exporting a tt file, doesn't matter for a single server or the server list, the remember last channel checkbox's value will also be saved on tt files as how as the user have set them and will be loaded when importing tt file, so users do not have to turn it on or off again for any server, their settings are restored as how as it was before.
this fixes https://github.com/BearWare/TeamTalk5/issues/2554

PR #2565, opened 2025-04-09
Store and load last-joined-channel in .tt file

bear101:
Same as "rememberLastChannel" in Android client.

@amirmahdifard this is to support your changes in #2562


amirmahdifard:
@bear101 Thanks for sinking with qt client! would you also put nickname please? qt client also doesn't stor nickname. I already did for android in my pr, so please you also do aswel. Thanks!

PR #2568, opened 2025-04-11
Don't translate "Default" sound pack name

CoBC:
Avoid to ask for default sound pack usage at every language change.
@bear101 The migratioon of SETTINGS_SOUNDS_PACK in utilui.cpp isn't apply, do you know why, I don't understand :(

### DIFF
--- a/Client/qtTeamTalk/utilui.cpp
+++ b/Client/qtTeamTalk/utilui.cpp
@@ -177,6 +177,10 @@ void migrateSettings()
             ttSettings->remove("texttospeech/tts-timestamp");
         }
 #endif
+
+        // Default sounds pack name changed in 5.5 format
+        if (ttSettings->value(SETTINGS_SOUNDS_PACK, SETTINGS_SOUNDS_PACK_DEFAULT).toString() == QCoreApplication::translate("UtilUI", "Default"))
+            ttSettings->setValue(SETTINGS_SOUNDS_PACK, SETTINGS_SOUNDS_PACK_DEFAULT);
     }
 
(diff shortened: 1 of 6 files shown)


Client/qtTeamTalk/utilui.cpp
```diff
@@ -177,6 +177,10 @@ void migrateSettings()
             ttSettings->remove("texttospeech/tts-timestamp");
         }
 #endif
+
+        // Default sounds pack name changed in 5.5 format
+        if (ttSettings->value(SETTINGS_SOUNDS_PACK, SETTINGS_SOUNDS_PACK_DEFAULT).toString() == QCoreApplication::translate("UtilUI", "Default"))
```
bear101, line 182: To disable the translation shouldn't it be:
`if (ttSettings->value(SETTINGS_SOUNDS_PACK, QCoreApplication::translate("UtilUI", SETTINGS_SOUNDS_PACK_DEFAULT)).toString() == QCoreApplication::translate("UtilUI", "Default"))`

Otherwise you'll compare "Default" to translated "Default.

CoBC: I tried but the result is same, migration isn't apply :(

bear101: The problem is that the translation file is not loaded when `migrateSettings()` is called. Therefore `QCoreApplication::translate("UtilUI", "Default")` returns "Default".

I suggest to remove `SETTINGS_SOUNDS_PACK` from settings in `migrateSettings()` or leave it untouched.

CoBC: Ah yes, you're right! Thanks, I've removed it from migrateSettings().

PR #2571, opened 2025-04-12
Insert ChatTextList/ChatTextEdit at correct index

bear101:
Fixes issue where edit field is above the history.


CoBC (approved):

CoBC (commented):
Oh, bad radiobutton checked for the review :(, I don't approve for now.
Message edit field still appearing above history here :(

CoBC:
I'm wondering if the issue can be related to the fact that the message edit field and send button are hidden by default...

PR #2573, opened 2025-04-12
Hotfixes for new history list

CoBC:
* Avoid to use same accelerator key multiple times
* Remove accessible name and accessible description from ui file, already defined by cpp
* Use getFormattedSize for disk quota, same as plain text history


valiant8086:
Sorry, now that this stuff is in master, I was searching for a way to get snapshot builds that I presumed would be based on master, but I've found nothing. Does this project not do that?

bear101:
Choose "Actions" in the top menu and then select a build. For each build you'll see a list of artifacts (compiled binaries) for testing.

PR #2577, opened 2025-04-16
Fix tab order for ChatTextEdit/ChatTextList and ChatLineEdit

### LINKED ISSUE #2576: History list is placed after text edit field
### Description

When pressing Tab from the tab list, focus is placed on the message edit field firstly, then on the new history list.

### Application

- [x] qtTeamTalk
- [ ] TeamTalkAndroid
- [ ] iTeamTalk
- [ ] TeamTalkClassic
- [ ] TeamTalkServer

### Platform

- [x] Windows
- [x] macOS
- [ ] Android
- [ ] iOS
- [x] Linux

### Expected behavior

Like the old history area, which was focused before the message edit field

### Actual behavior

Focus is on the text edit field then when pressing tab, it is on the history list

1. place the focus to the tab list, with the chat tab selected
2. Press tab

bear101:
Fixes #2576

### DIFF
--- a/Client/qtTeamTalk/mainwindow.cpp
+++ b/Client/qtTeamTalk/mainwindow.cpp
@@ -2776,6 +2776,7 @@ void MainWindow::setupChatHistory()
         auto index = layout->indexOf(ui.chatEdit);
         layout->removeWidget(ui.chatEdit);
         layout->insertWidget(index, chat);
+        ui.chatTab->setTabOrder(chat, ui.msgEdit);
         m_chathistory[TAB_CHAT] = chat;
         delete ui.chatEdit;
         ui.chatEdit = nullptr;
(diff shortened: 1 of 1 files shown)


Nardol:
Visualy, is the list at the same place than the text area is when list is disabled?
Because in a screen reader perspective, it looks list is below Message edit field when history text area looks to be above. But the review cursor can sometimes give a wrong representation of an interface.

CoBC:
When I write a text in message edit field, tab order is: send button, history list, message edit field.
We can probably fix it in the same way, but we have to confirm order visually. If I use an app to recognize screen content, the message edit field isn't visible at all apparently :(
@bear101 can you confirm if it's only a tab order issue or if we really have a visual issue with this list?

CoBC (approved):

bear101:
@CoBC  is ChatLineEdit now recognized?

CoBC:
@bear101 If you can, maybe release a new beta to include this fix

bear101:
> @bear101 If you can, maybe release a new beta to include this fix

Done!

Issue #2516, opened 2025-01-09
Label Android client's sliders


amirmahdifard:
### Description

Currently, everything such as buttons, switches has lables, but only seeck bars doesn't have lables. those cannot be labled with android:text, so I'll use content discription to lable those for screen readers, because android says that because seeckbars are dinamic, if we want to lable them for screen readers, we should use content discription for them. I'll soon work on these issues. I opened them so that I can remember.

### Application

- [ ] qtTeamTalk
- [X] TeamTalkAndroid
- [ ] iTeamTalk
- [ ] TeamTalkServer

bear101:
Do you plan to fix this? Otherwise I'll move it to Discussions' feature requests.

PR #2580, opened 2025-04-16
Android. Fixed texts, labled sliders, and added userid

amirmahdifard:
Currently, for example, in texts, instead of havving like this
connection:
online
or,
username
amirmahdifard
the texts are like this
connection: online
username: amirmahdifard
to match the desktop client, eas of adding new texts, clean code, more beautiful, less swiping and less annoying for screenreader users, and other goodys.
Note that I made it so that it's up to the translaters how they want to seperate lable and data from each other, for english translation I seperated with colen, But if a translater doesn't seperate them with anything, then only a space will seperate them, like this
connection online
so, it's up to them, may be someone wants to seperate with comma or anything else in their strings, they don't need to touch the code, I added colen to the english translation.
also, added userid to the user prop to be shown, the texts are now much clean, beautiful and better
added content description to sliders As I said i'll do them before
this fixes https://github.com/BearWare/TeamTalk5/issues/2516


Client/TeamTalkAndroid/src/main/res/values/strings.xml
```diff
@@ -175,11 +175,13 @@
     <string name="channel_prop_title_noaudiorecord">No Audio Recording</string>
 
 
-    <string name="user_prop_title_username">Username</string>
-    <string name="user_prop_title_statusmsg">Status Message</string>
-    <string name="user_prop_title_clientname">Client Name</string>
+    <string name="user_prop_title_nickname">Nickname:</string>
```
bear101, line 178: Instead of appending strings, isn't it better to use "Nickname: %1$s" ?

amirmahdifard:
@bear101 no, that's defined by the code, just the lable is translatable, the data after it will just be desplayed automaticly. This is the best way for implementing this. Thanks!

amirmahdifard:
@bear101 I also fixed a long standing bug that exists since long ago that for some random users such as tt media player bots, and other bots/custom clients, the user volume slider and media volume slider, and the default buttons was not shown in the user prop.
This pr is now finished, so please merge it. Thank you!

bear101:
Another Android beta release is on the way.

amirmahdifard:
@bear101 Wow! thanks! your comment really gave me energy! I'm really glad to know that my work is good! and stil I'll do! So, please don't yet releace the stable version

PR #2587, opened 2025-04-18
Custom Chat Templates

CoBC:
Add the ability to custom template for Channel, Broadcast and Private message.
Option "Timestamp messages" has been removed and replaced by this new template system.
Discussed in #2582 
@bear101 When I custom template for "Channel message" it doesn't take effect, I have to custom "Private message", so it seams channel messages are associated to user type, do you know why?


Client/qtTeamTalk/chattextedit.cpp
```diff
@@ -246,6 +244,18 @@ QString ChatTextEdit::addTextMessage(const MyTextMessage& msg)
 
     line += QString("%1\r\n%2").arg(getTextMessagePrefix(msg, user)).arg(msg.moreMessage);
 
+    switch (msg.nMsgType)
+    {
+    case MSGTYPE_CHANNEL :
+        line = UtilUI::getChatTemplate(SETTINGS_CHATTEMPLATES_CHANNELMSG, {{"{date}", dt}, {"{user}", getDisplayName(user)}, {"{content}", msg.moreMessage}});
```
bear101, line 250: Isn't there a "break" missing? Same in chateditlist.cpp

bear101:
It looks like it's not possible to avoid time stamps on "log messages". See e.g. ChatTextEdit::addLogMessage(..)

CoBC:
Yes, log messages aren't concern for now. Should I also add log message template?

bear101:
> Yes, log messages aren't concern for now. Should I also add log message template?

Please do. The log messages become quite long and maybe some people would rather not use timestamp.

valiant8086:
If not already, someone should probably take the check box out that was there to turn timestamps on and off. Doing that with the template instead. If template will be too complicated for people, could use a list that has the two default behaviors, yes with timetamps yes without and a third item that's what's selected if it's a custom one. I'll probably be able to test it soon, if the master branch can build in github actions now.

CoBC:
Yes, I already removed this checkbox. You can test using this artifact: https://github.com/BearWare/TeamTalk5/actions/runs/14548540064/artifacts/2973457492

valiant8086:
Ok. I've been able to test. I like it very much.

1. When I try to press enter to make a new line in a template, the way they already are in there, it hits ok and puts me back on the display panel of settings. shift+enter and ctrl+enter don't do anything. Workaround was to make the template in notepad and copy and paste that to the edit area for the template.
2. I think this one might be new, when I push shift+f10 on a message, then down arrow to focus the context menu, and then close it, if I shift+f10 again, the resulting context menu is totally not accessible even by knowing I need to arrow first to find it. I'm pretty sure I had it open multiple times in a row when I was testing 5155. Workaround is to escape out of the context menu that's not reading anything, escape out of the conversation and open it again.
3. back on the templates discussion, what is this checkbox here that says custom template setting. Wouldn't we assume I want to customize templates if I press the button to open that dialogue? I think you can remove that checkbox and get by with a reset to default button that appears next to the multiline edit for each available template type. Pressing it resets the currently selected template to default values. Could also put a reset all templates button there. If I don't want custom templates but I've messed with them, I hit the message template settings button and then I hit the reset all message templates button, close out of that dialogue and I'm all done. If you want to be able to toggle custom templates on and off, well I guess that check box makes more sense then. You could put it inside the templates dialogue to declutter the display panel a bit, but that's just busywork.
4. at some point, it might be worth removing the checkbox to show display name instead of username and integrate that into templates. have {displayname} and {username} it facilitates something bazaar like {displayname} ({username}) {content} {date} which is most likely not something I want, but...

I'm really pleased with this. I've removed the < and > from the name, put the date at the end and I stopped it showing the year for all of my templates. Thanks for your work. Again though, the qt edit rears its ugly head because it is very difficult to even edit the templates in the edit field there due to accessibility issues. I can't wait till qt fixes that or screen readers catch up. For now, even if you figure out how to let me put newlines in there, for blind users it is going to make the most sense to edit templates in a document editor and then copy them over to the edit field after we have it the way we need.

CoBC:
1. Fixed by #2598, Shift+Enter will insert a new line
2. Hmm, yes, I can reproduce, I have to investigate this one...
3. There is already a button to reset all templates to default value. The checkbox is just autochecked when at least one template has been changed. The goal of this checkbox is just to align UI with all other custom options, to have a checkbox and a button.
4. No, we will probably not remove this checkbox because it have an effect on other piece of UI, E.G. channel tree, not only chat history.

PR #2588, opened 2025-04-18
android: added user logged in and user logged out to the chat list in chat tab

amirmahdifard:
Finally, as so many requests from users, and also like desktop clients wich has every events on history, the same has been done for android client. Now, user logged in and user logged out events are also shown on the chat history list.


Client/TeamTalkAndroid/src/main/java/dk/bearware/backend/TeamTalkService.java
```diff
@@ -936,21 +936,23 @@ public void onCmdUserLoggedIn(User user) {
                 activecmds.put(cmdid, CmdComplete.CMD_COMPLETE_UNSUBSCRIBE);
         }
 
+        String name = Utils.getDisplayName(getBaseContext(), user);
```
bear101, line 939: If you have a server with 50+ users then this will generate a lot of entries in the chat-log. Are you sure you want this?

amirmahdifard:
@bear101 yes,  it's the expected behave. I do test all of my changes in all ways when i'm developing something and before I push the changes, so no worrys. So yes, it's expected to generate all loggins and outs in the chat list, actually the system for it is just generally like that, and it's also good for example we can track who are currently logged in to the hole server whenever we are connected to the server in total, just by looking at the chat list. Thanks!

bear101:
@beqabeqa473 do you think it's a good idea to get a notification for every user that is logged in?

CoBC:
I would suggest to have the ability to enable/disable log events, as we have in Qt client, so each user can individually check or uncheck logins, logouts, channel join/left and other events

amirmahdifard:
the settings are already there, we can enable/disable. Also there is a checkbox that enables or disable the logs

amirmahdifard:
do you think it's a good idea to get a notification for every user that is logged in?
@bear101 Can I know the exact problem of this implementation please, it's the same as qt client, every loggins and outs are in the history, wich is already the case for qt client. So, what's the problem and what do you exactly want to happen? let me know. Thanks.

bear101:
> "do you think it's a good idea to get a notification for every user that is logged in?" @bear101 Can I know the exact problem of this implementation please, it's the same as qt client, every loggins and outs are in the history, wich is already the case for qt client. So, what's the problem and what do you exactly want to happen? let me know, So I can implement. Thanks.

When you log into a server with 50 users then you get 50 entries in the chat tab. It is possible to filter the TTS events but you cannot filter the chat tab entries.

amirmahdifard:
@bear101 yeah, I understand this you are talking about. But as I said before, the system in android client is just like that and nothing can be done to fix it. For example, the same also happens with tts events when you loos your connection to the server and then your connection is back, your tts events will announce logins of all users that are currently in the server, so it's not falt of this pr. There is only one place for handling logged ins and outs in teamtalk service on android, and the only thing I did is that I modifyed it to also generate these events in the chat list
But even with this, it's a good thing. We can just see all users in the server, with out looking at all channels and online users dialog wich currently we don't have in android client
also later, I'll open a pr to increace the amount of entry items the chat list can holde, because I think I've seen in the code that currently it only can holde 100 items and the old ones will be cleared. i'll make sure performence and stability of the app when I do everything.
So can this now be merged so I can work on replacing buttons in main window with seeckbars and give it so users can have it in 5.18? Thanks a lot.

bear101:
Ok, I'll squash the commit, so we can revert it if users start to complain.

amirmahdifard:
@bear101 Don't worry, they won't complane. A lot of users exactly wanted this that I did now.

amirmahdifard:
@bear101 Always I have a full backup of the source code before I start to implement features, so if a revert is needed at any time, I can just open a pr to revert. So, don't worry. Even if users complane wich i'm sure they won't, I'll remove the generation of user logged in messages, and only keep log out messages. So, don't worry

PR #2594, opened 2025-04-18
Use checkbox and ... QToolButton to edit custom chat messages

bear101:
This makes the toggling similar to other sub-configuration options.


Client/qtTeamTalk/preferences.ui
```diff
@@ -442,18 +442,45 @@
...
+              </item>
+             </layout>
             </item>
             <item>
              <layout class="QHBoxLayout" name="horizontalLayout_30">
               <item>
                <widget class="QLabel" name="timestampformatLabel">
                 <property name="text">
-                 <string>Timestamp format</string>
+                 <string>Text message timestamp format</string>
```
CoBC, line 483: Why this change? This doesn't affect only text message

bear101: What else is time stamped?

CoBC: The format defined here affects all date and time, E.G. on files list, user accounts, banned users...

bear101: Ah, ok. I'll change it back.

CoBC (approved):

PR #2606, opened 2025-04-20
Replace radiobuttons by Combobox in video capture tab

CoBC:
Also use suffix instead of different label to display bitrate unit

### DIFF
--- a/Client/qtTeamTalk/preferencesdlg.cpp
+++ b/Client/qtTeamTalk/preferencesdlg.cpp
@@ -1656,7 +1656,7 @@ void PreferencesDlg::slotVideoResolutionChange(int index)
     Q_UNUSED(index);
 }
 
-void PreferencesDlg::slotImageFormatChange(bool /*checked*/)
+void PreferencesDlg::slotImageFormatChange(int /*index*/)
 {
     m_vidfmt.picFourCC = FourCC(getCurrentItemData(ui.vidImgFmtBox).toInt());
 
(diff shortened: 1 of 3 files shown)


Client/qtTeamTalk/preferencesdlg.cpp
```diff
@@ -209,12 +209,7 @@ PreferencesDlg::PreferencesDlg(SoundDevice& devin, SoundDevice& devout, QWidget
     connect(ui.captureformatsBox, QOverload<int>::of(&QComboBox::currentIndexChanged),
             this, &PreferencesDlg::slotVideoResolutionChange);
     connect(ui.vidtestButton, &QAbstractButton::clicked, this, &PreferencesDlg::slotTestVideoFormat);
-    connect(ui.vidrgb32RadioButton, &QAbstractButton::clicked,
-            this, &PreferencesDlg::slotImageFormatChange);
-    connect(ui.vidi420RadioButton, &QAbstractButton::clicked,
-            this, &PreferencesDlg::slotImageFormatChange);
-    connect(ui.vidyuy2RadioButton, &QAbstractButton::clicked,
-            this, &PreferencesDlg::slotImageFormatChange);
+    connect(ui.vidImgFmtBox, QOverload<int>::of(&QComboBox::currentIndexChanged), this, &PreferencesDlg::slotImageFormatChange);
```
bear101, line 212: PreferencesDlg::slotImageFormatChange takes a boolean as argument. Shouldn't it be an `int` ?

I experience a crash here:
9  PreferencesDlg::slotVideoCaptureDevChange(int) preferencesdlg.cpp 1564 0x10012f0ea 
10 PreferencesDlg::slotImageFormatChange(bool)    preferencesdlg.cpp 1665 0x100130696

`index` in PreferencesDlg::slotVideoCaptureDevChange(int index) is -1.

bear101 (changes requested):

CoBC:
I didn't see the crash here, but yes, I changed to int argument

Issue #2529, opened 2025-02-28
Android client. Increace/dcreace volume/ microphone gane buttons in main screen of the app should be replaced by sliders to be normal like any other app


amirmahdifard:
### Description

Currently, in main screen of the app, we have a fue buttons, to turn up/down the speecker volume and a button to mute/unmute it, and same for microphone gane. This method is not good and for modifying volumes and stuf, we have sliders and should use those. sliders will remove 4 buttons from the screen, and with the slider it self you can understand what is the current value, and modifying it is much more simpler and modern, exactly like the sliders used in user profiles and channel creation dialogs. I will do these as soon as possible and bear101 merge my commits. Thanks

### Application

- [ ] qtTeamTalk
- [x] TeamTalkAndroid
- [ ] iTeamTalk
- [ ] TeamTalkServer

amirmahdifard:
@CoBC hello. will you please assighn me to this issue? Thanks

bear101:
Do you plan to fix this? Otherwise I'll move it to Discussions' feature requests.

amirmahdifard:
@bear101 yes i plan to work on this and all my other issues, but as I said before, i'm waiting for my open prs to be merged, so I can open prs one by one in turm

PR #2607, opened 2025-04-20
android. Changes to replace buttons with seekbars

amirmahdifard:
Done!

### DIFF
--- a/ChangeLog.txt
+++ b/ChangeLog.txt
@@ -28,6 +28,7 @@ Default Qt Client
 - Fixed OPUS DTX mode sending too big packets when variable bitrate mode (VBR) was disabled
 - Fixed ask to use default sounds pack at every language change (will be asked one last time)
 Android Client
+- Replaced the old turn up/down buttons for speaker volume and microphone gain with new brand sliders that has a lot of advantages over the buttons, and looks more beautiful!
 - Added popup menu actions in Server List dialog
 - Added "Message" and "User properties" user menu actions in channel list tab
 - Ability to export each server in to a .tt file separately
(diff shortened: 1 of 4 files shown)


amirmahdifard:
@poretsky hi, I hope you are doing good. Can I count on your help for this implementation a bit please? thank you for answering!

amirmahdifard:
my problem is that this crashes. @bear101 if you have idea I would be a lot thankful to you, so I don't loos this important feature

amirmahdifard:
@bear101 If you have a simple review to this, you can see that i'm really handling everything same as buttons, even the tt get sound and set sound, adding all needed checks. But I have no idea why the app crashes when pressing a server to connect. I've done everything I just don't know why is the crash happening, wich if I did I'd fix the crash quickly, please give me a little bit of help on this one, really thank you.
Advantages of seekbars over the buttons
1. Fast for both sited and blind users.
2. Blind users has a seekbar controler feature in their screenreader like talkback, wich they can enter a number and the slider will be set automaticly to that number, raather than interacting with buttons.
3. Seekbars are desighned to controle volumes even like windows teamtalk, raather than buttons.
A small help will be greately good. Thanks!

amirmahdifard:
never mind, finally after destroying my self for 2 days constantly, I found the crash. As I said before, I was sure about my code, the problem was caused by incorrect leyowt desighning. i'll fix this until end of today. Thanks!

amirmahdifard:
@bear101 I have a question would you please answer quickly, The app crashes only by these lines wich is being fixed when I comment them out. Do you know why
        masterSeekBar.setProgress(Utils.refVolumeToPercent(ttclient.getSoundOutputVolume()));
                micSeekBar.setProgress(Utils.refGainToPercent(ttclient.getSoundInputGainLevel()));

amirmahdifard:
@bear101 are you ok with this, they are now working. Can you please have a look and let me know your think? thanks.

amirmahdifard:
@bear101 This is now completed!

amirmahdifard:
this fixes https://github.com/BearWare/TeamTalk5/issues/2529

PR #2608, opened 2025-04-20
Update documentation for TeamTalk v5.18

bear101:
@CoBC could you maybe write a few notes regarding the text-to-speech engines in preferences.md. Look for:

`# Text to Speech {#texttospeechtab}`


amirmahdifard:
@bear101 can you please don't releace 5.18 yet? my changes won't take long. Thank you

CoBC:
What type of notes do you want? Should I explain each option as you did for all tabs, or do you only want some short explanations for each engine (E.G. precise something like: "Tolk, available on Windows, use running screen-reader to transmit TTS events")?

bear101:
> What type of notes do you want? Should I explain each option as you did for all tabs, or do you only want some short explanations for each engine (E.G. precise something like: "Tolk, available on Windows, use running screen-reader to transmit TTS events")?

E.g. on Windows there's TTS engines None, Default, Tolk and Qt Announcement. What are the advantages of each of these and which one is recommended.

CoBC:
@bear101 I wrote some notes about TTS engines, let me know if you want more or something else ;)

PR #2611, opened 2025-04-21
Fix file deletion

CoBC:
Already fixed in #2521, don't know when it was broke again :(


bear101:
It was removed in commit 1ef145dd8aa1e19681bab4156a7fadc47151b96d. I don't understand why. I think it's because FilesView is a subclassed control and therefore Qt Designer removes the feature because it does not know it's a QListView.

PR #2612, opened 2025-04-21
Fix MOTD not displayed in chat list

CoBC:
MOTD wasn't displayed if dialog box to display MOTD was enabled and chat control was configured as a list.
Also fix some missing spaces in updateServer() and joinedChannel()


bear101:
Shouldn't the MOTD dialog box be moved to mainwindow.cpp instead of having it two places?

CoBC:
In this case, maybe move all server info display in mainwindow?

bear101:
> In this case, maybe move all server info display in mainwindow?

Only the display of the MOTD dialog box. The MOTD in the chat-text/chat-list should remain.

PR #2615, opened 2025-04-22
Android. Old 4 buttons has been replaced with 2 new brand nice seekbars

amirmahdifard:
So, I've replaced these buttons finally, and made them exactly act as the buttons, the text is not removed so sited users also can have an idea of what the sliders are for, and also inclooded content description for blind users.
When activating or deactivating the voice activation, exactly like the buttons did, the content description and the text's labels will be changed to reflect the new value and change, also the slider's value will be updated to reflect the value of the gain or voice activation, So there is no mismatch, every changes are being tested, and made sure there is no crash, no unexpected bug, and sliders are completely normally, kicked out the buttons. Now people can enjoy using these. I'm really happy now when looking at teamtalk's screen now, both the server list and main screens. In this version I redesighned both screen and made them look more beautiful!, And removed everything related to those turn up/down buttons.
this fixes https://github.com/BearWare/TeamTalk5/issues/2529

### DIFF
--- a/ChangeLog.txt
+++ b/ChangeLog.txt
@@ -28,6 +28,7 @@ Default Qt Client
 - Fixed OPUS DTX mode sending too big packets when variable bitrate mode (VBR) was disabled
 - Fixed ask to use default sounds pack at every language change (will be asked one last time)
 Android Client
+- Replaced the old turn up/down buttons for speaker volume and microphone gain with new brand sliders that has a lot of advantages over the buttons, and looks more beautiful!
 - Added popup menu actions in Server List dialog
 - Added "Message" and "User properties" user menu actions in channel list tab
 - Ability to export each server in to a .tt file separately
--- a/ChangeLog.txt
+++ b/ChangeLog.txt
@@ -28,7 +28,7 @@ Default Qt Client
 - Fixed OPUS DTX mode sending too big packets when variable bitrate mode (VBR) was disabled
 - Fixed ask to use default sounds pack at every language change (will be asked one last time)
 Android Client
-- Replaced the old turn up/down buttons for speaker volume and microphone gain with new brand sliders that has a lot of advantages over the buttons, and looks more beautiful!
+- Replaced turn up/down buttons for speaker volume, microphone gain and voice activation level with sliders
 - Added popup menu actions in Server List dialog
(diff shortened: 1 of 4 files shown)


bear101:
The seekbars are so small that it is almost impossible to change them. Either they go to zero or 100 %

amirmahdifard:
@bear101 Even tho i'm blind, I tryed mooving them like sited users, And yes I also think they are small, But I tryed to make them exactly same as sliders in user prop dialog. But, I cannot see the screen, So would you be able to fix this yourself from activity main.xml leyowt, or can you tell me what should be changed? hight, columb, wro? etc? or what?

bear101:
I've been trying to make the seekbars bigger but I cannot find a way. Maybe it's possible to only display the seekbars in horizontal mode and then they can have a minimum size?

amirmahdifard:
@bear101 just make them like use prop. remove the grid leyowt vol controles, and don't explisitly set android:columb etc, and just let system define it per user's mobile and tablets screens. In these kind of cases, it's always to leave sises up to the system it self to deturmin, because you don't know what other's phone's screen sise, so making it up to the system makes it compatible with all devices

amirmahdifard:
@bear101 when buttons were existed before the sliders, you didn't have this being small problem. The only thing I did Is that I put the seekbarsinside a lini leyowt. So if you want, I can take them out, and then you can see if the issue fixes or not. Do you want me to?

amirmahdifard:
@bear101 see if this makes any difrence? if not, i'll move the mute speekers button and turn voice activation button a little bit more to the right side, so seekbars can dance around more easily in the senter of the screen, and they aren't pushed off by the buttons. I'm a waiting on your answer to know what to do

amirmahdifard:
@bear101 I wanted to say, Can you please merge my prs so I can also continue my work and don't loos my hope for implementing new features and fixing bugs? You can merge my pull requests, and making seekbars bigger and other things can be done on other new pull requests, This pull request is implementing this feature and my other pull request is fixing an open bug, and they are completely unrelated to each other. So, Can you please merge? Thanks so much!

amirmahdifard:
@bear101 hi. I found the problem and fixed the seekbars. can you please have a look and see if thats ok for you so finally this can be agreed? thanks

PR #2617, opened 2025-04-23
Wrap "Connection status" to make room for controls

bear101:
This makes more room for other controls.

### DIFF
--- a/Client/TeamTalkAndroid/src/main/java/dk/bearware/gui/MainActivity.java
+++ b/Client/TeamTalkAndroid/src/main/java/dk/bearware/gui/MainActivity.java
@@ -1339,7 +1339,7 @@ private EditText newmsg;
                         con = getString(R.string.stat_connecting);
                     }
                     
-                    connection.setText(getString(R.string.label_connection) + " " + con);
+                    connection.setText(getString(R.string.label_connection) + "\n" + con);
                     connection.setTextColor(con_color);
 
                     ClientStatistics stats = new ClientStatistics();


amirmahdifard:
@bear101 I don't think this is a good idea, because stats and volcontroles are completely in difrent grids, so if you want to mess around with them and move them around or making them small or big, should be done from activity main.xml, not the actual code. The actual code doesn't controle the way items are audered in the screen. In my pr I told you a way we can fix that. Or just close this pr, and increase seekbar's waits a little

bear101:
@amirmahdifard Does a line break change something for the screen reader? Visually this looks better.

amirmahdifard:
@bear101 As I said before, I just tested it, and it's not really good as before this pr. For example when navigating line by line, the others wich are not have line break are announced like ping: 3, but this one, first is connection, and another swipe is needed to reach the online, and sometimes the focus jumps around when swiping. And that Was one of the biggest reason why I created the pr to remove duplicate texts, because That was annoying. When I did that, the app was made a lot better for blind people with any tipe of other screenreader compatibility, and Sited users could see the texts as wel.
Even by this pull request, there is no more space for other controles and this doesn't have anything to do with making seekbars bigger.
So, if possible, please don't merge this pr and close it, so there is a balance between sited users and blind users, Because i'm sure before this pr sited users didn't have any problem seeing this text.
Thanks a lot!

amirmahdifard:
I reverted it back for my self.

PR #2618, opened 2025-04-23
Only indicate "Online" status if authenticated

bear101:
When kicked from server then "Connection Status" should not be online. Instead indicate "Unauthorized".


amirmahdifard:
@bear101 nice work! I haven't tested this yet, but befor when connection to server was lost, it would become connection offline, but I think in that case it should've became connecting because client is trying to connect to the server every fue seconds

PR #2621, opened 2025-04-23
Add "-v" command line argument to retrieve TT version

CoBC:
Requested by @Nardol for automatic update detection, download and installation. Can probably be useful to others.
@bear101 I see a strange behaviour on Windows, version is print directly in the prompt, not on another line as it is habitually the case. Do you know a better way to handle info on Windows?

### DIFF
--- a/Client/qtTeamTalk/main.cpp
+++ b/Client/qtTeamTalk/main.cpp
@@ -15,6 +15,9 @@
  * along with this program.  If not, see <https://www.gnu.org/licenses/>.
  */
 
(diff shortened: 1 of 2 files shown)


bear101:
I just check the About menu item to see version information.

Nardol:
I use a Bash script to update TeamTalk and I would like to check if the update is really necessary before downloading, extracting and replacing updated files.
To do so, having a command line parameter which print the actual version as this PR implements would allow to compare actual and latest version directly in a Bash script.

Issue #361, opened 2018-05-30
Transmit control for mobile clients


bear101:
Class-room control and block streams currently isn't supported in mobile clients.

muamal-aljanahi:
We often need this feature
 To prevent noise
👍

beqabeqa473:
Hi all.

what do you think about design of these functions?

add functions in a contextt menu, or make them as now subscriptions are but allign them on the right?

bear101:
I don't know what gives quickest access. It's a bit slow if you have to first press a user, start a new activity and then return back to the channel list. But maybe it's the only option?

beqabeqa473:
it is possible to add it to context menu, but does it give any benefit
tha's a question.

i am now starting to work on that feature, and if no other ways will
be found, possibly i will do it in userprop, on the right side.
i am also working on issue 67.

when you have a deadline for releasing 5.8 version?
i'm asking to understand if i can finish it before release.

thanks.

On 8/30/21, Bjørn Damstedt Rasmussen ***@***.***> wrote:
> I don't know what gives quickest access. It's a bit slow if you have to
> first press a user, start a new activity and then return back to the channel
> list. But maybe it's the only option?
>
> --
> You are receiving this because you commented.
> Reply to this email directly or view it on GitHub:
> https://github.com/BearWare/TeamTalk5/issues/361#issuecomment-908558950


-- 
with best regards Beqa Gozalishvili
Tell: +995593454005
Email: ***@***.***
Web: https://gozaltech.org
Skype: beqabeqa473
Telegram: https://t.me/gozaltech
facebook: https://facebook.com/gozaltech
twitter: https://twitter.com/beqabeqa473
Instagram: https://instagram.com/beqa.gozalishvili

bear101:
If you can finish this week then I don't mind postponing the release. Alternatively we'll make a v5.8.2 for Android that is released shortly afterwards.

beqabeqa473:
Ok, thanks, i will try to finish these two tasks this week.

On 8/31/21, Bjørn Damstedt Rasmussen ***@***.***> wrote:
> If you can finish this week then I don't mind postponing the release.
> Alternatively we'll make a v5.8.2 for Android that is released shortly
> afterwards.
>
> --
> You are receiving this because you commented.
> Reply to this email directly or view it on GitHub:
> https://github.com/BearWare/TeamTalk5/issues/361#issuecomment-909486943


-- 
with best regards Beqa Gozalishvili
[...]

PR #2630, opened 2025-04-28
android client: Added transmit controles to user prop

amirmahdifard:
In user prop, as everyone requested in the related issue, now subscriptions are on the left side, and transmit controles are on the right side.
This fixes https://github.com/BearWare/TeamTalk5/issues/361


amirmahdifard:
@bear101 please merge this at least. Thanks.

bear101:
When I test on Pixel 5 then only the left half of the checkbox is visible. Do you know if we can add scrolling to the user properties activity?

amirmahdifard:
@bear101 yeah i'll fix that, The transmission checkboxes are too much at the edg of the right of the screen. i'll fix that

amirmahdifard:
@bear101 done. Now all columb spans are being removed, and to say simply, there are only 2 columbs, and columb 0 wich is actually columb 1 is for holding subscriptions, and columb 1 is holding transmit controles. Sweet as apple, ready to merge!

bear101:
The controls in the second column are still only half visible. E.g. only "Allow voice transmi" is show on Pixel 5 (I use emulator).

amirmahdifard:
@bear101 in the code, these are 2 seperet columbs, that should devide the screen 50% by 50% from senter of the screen, so why subscriptions are so big and come until senter of the screen, but transmitions are going out of the screen because they are too small? but in the code they are both same, I even tryed to put width of the transmit controles to match parent instead of rap content, didn't fix. I tryed setting their columb span to 2 while subscriptions spans were 1, stil didn't resolved. What's causing this?

amirmahdifard:
there is too much space between the left and right options. If the spaces could be removed so that the options are exactly beside each other, this would be solved and that's exactly what I want as wel.

amirmahdifard:
@bear101 there are too much requesting announcements and sounds of subscriptions and intercept subscriptions for android, But because yet we don't have none of them, I don't have cloo about wich event I should call, what to call in main activity case, and what in teamtalk service case. Can you please ad one of them only as an example, eg, subscription voice, so then I Can follow and add the rest. Also one of the transmit announcements as wel. Thanks!

amirmahdifard:
I don't know what's the problem of columb 1 completly, in prop of some random bots for example, completely even default button is not shown. and the screen already has so much space, so I don't know why they don't get feet. @bear101 what I do finally?

amirmahdifard:
@bear101 first, I don't know how to do that, so if you want, you can merge this pr and do that in another pr. second I don't think that's a good idea, because currently there is not so much options on it that cannot be feet in the screen, we should just fix the spaces that are between columbs. scrolable activity is for times you have so many options, and you want to scrole between options. But stil, you can do whatever you want.

amirmahdifard:
@bear101 I will try to fix thix, but a question. When user prop becomes a scrolable activity, do you stil want the transmit controles to be in right side, or do you want them below all subscriptions and intercepts so you can scrole the screen up and down? answer this so I will know what i'm doing.

amirmahdifard:
@bear101 i'm sure now this is exactly what you want. Just have a look and see how it's ok now!

Client/TeamTalkAndroid/src/main/java/dk/bearware/gui/UserPropActivity.java
```diff
@@ -298,6 +309,76 @@ else if(btn == subscribeInterceptmedia)
                 } else {
                     ttclient.doUnsubscribe(user.nUserID, Subscription.SUBSCRIBE_INTERCEPT_MEDIAFILE);
                 }
+            else if(btn == transmitVoice)
+                if (checked) {
+                    Channel chan = new Channel();
+                    ttclient.getChannel(user.nChannelID, chan);
+                    chan.transmitUsers[0][0] = user.nUserID;
```
bear101, line 316: You need to assign the user-ID to the first value in `transmitUsers[0]` that is 0. Otherwise you end up erasing the first already in the list.

bear101:
Nice job with the scrollview!

bear101 (changes requested):

Client/TeamTalkAndroid/src/main/java/dk/bearware/gui/UserPropActivity.java
bear101: I would suggest you create a separate method in Utils class for handling this toggling. Then input arguments to the method can be user-ID and StreamType.

amirmahdifard:
@bear101 about handling in utils, no, I'd prefare to keep it like this, It's nice this way. Thanks.

amirmahdifard:
@bear101 about your first comment regarding assighning userid to 0, would you please show one of them as an example in a new comment that how you want it? because I don't really understand what you mean. Thank you!

amirmahdifard:
@bear101 I think may be you want something like this,
for (int i = 0; i < chan.transmitUsers.length; i++) {
    if (chan.transmitUsers[i][0] == user.nUserID || chan.transmitUsers[i][0] == 0) {
        chan.transmitUsers[i][0] = user.nUserID;
        chan.transmitUsers[i][1] = StreamType.STREAMTYPE_VOICE; // or NONE
        break;
    }
}
If yes, I onnistly don't know why because the current works completely fine, it does not have any problem. I tested this feature before pushing to you, so what's the problem with it?

amirmahdifard:
@bear101 I have to implement so many features for teamtalk android with in 5.18.1, So I need a very active development. Can you please answer my comments more quickly and every process to be more quicker please, so I can be a bit faster. Thanks!

amirmahdifard:
ok I've found the overrighting bug, i'll fix it soon

amirmahdifard:
i'll fix this, And will make it ready for review after

amirmahdifard:
@bear101 hi! This is now, ready to merge, I've tested and confurmed the following.
Disabled voice transmission of my self, and then disabled voice transmission or video transmission or whatever transmission of another user, and confurmed that voice transmission of my self is stil disabled and not wiped out
Disabled a user's voice transmission, and then, again disabled that user's video transmission, And made sure that voice transmission of the user is not wiped out during setting the video transmission.
Looked and made sure that the value of the switches are properly retreaved, the switch is on when user really have the streamtype, and it is off when user doesn't have it.
So, this is now as perfect as possible, i'm waiting for the merge. Thanks!

Client/TeamTalkAndroid/src/main/java/dk/bearware/gui/UserPropActivity.java
```diff
@@ -172,6 +178,54 @@ void showUser() {
         subscribeInterceptdesk.setChecked((user.uLocalSubscriptions & Subscription.SUBSCRIBE_INTERCEPT_DESKTOP) != 0);
         subscribeInterceptmedia.setChecked((user.uLocalSubscriptions & Subscription.SUBSCRIBE_INTERCEPT_MEDIAFILE) != 0);
 
+        Channel chan = new Channel();
+        ttclient.getChannel(user.nChannelID, chan);
+
+        boolean hasVoice = true;
+        for (int i = 0; i < chan.transmitUsers.length; i++) {
```
bear101, line 185: Put these loops into the Utils class so you're not repeating the same code. The method should have as input `userid`, `transmitUsers` and `StreamType`.

amirmahdifard:
@bear101 But I like it this way, This way will have less bugs and is more manageable for my self as a developer, As you see, I putted the announcements of transmits function managements in utils class because that was ok, But this one is more good if used direct like this.
I could also use a single for loop for all of these, But I  put difrent for loop for each of them, because that's more better. So can you please merge it so we can have it like this? be sure this is better and is better manageable, So if later users has a request, I can easyly do their requests. Thanks!

bear101:
I'm cleaning up this merge request here #2670

amirmahdifard:
@bear101 ok, you can take it.

bear101:
replaced by #2670

PR #2620, opened 2025-04-23
Android: Speacker is reactivated when microphone is disabled vire voice activation button same as transmit button

amirmahdifard:
this fixes https://github.com/BearWare/TeamTalk5/issues/1500

PR #2631, opened 2025-04-28
Fix speaker not being reenabled when disabling voice activation

bear101:
Continues #2620


amirmahdifard:
@bear101 you forgot to close issue 1500 as completed

PR #2644, opened 2025-05-10
Feat: Auto-switch to new default sound device

BlindMaster24:
@bear101 Please review this. I'm unable to test it currently. If it works, please merge, as many users would benefit from this change.

### DIFF
--- a/Client/qtTeamTalk/mainwindow.cpp
+++ b/Client/qtTeamTalk/mainwindow.cpp
@@ -1863,16 +1863,32 @@ void MainWindow::processTTMessage(const TTMessage& msg)
         qDebug() << "Unplugged sound device: " << _Q(msg.sounddevice.szDeviceName);
         break;
     case CLIENTEVENT_SOUNDDEVICE_NEW_DEFAULT_INPUT:
+        Q_ASSERT(msg.ttType == __SOUNDDEVICE);
         qDebug() << "New default sound input device: " << _Q(msg.sounddevice.szDeviceName);
+        if (ttSettings->value(SETTINGS_SOUND_INPUTDEVICE, SOUNDDEVICEID_DEFAULT).toInt() == SOUNDDEVICEID_DEFAULT) {
+            initSound();
+        }
         break;
     case CLIENTEVENT_SOUNDDEVICE_NEW_DEFAULT_OUTPUT:
+        Q_ASSERT(msg.ttType == __SOUNDDEVICE);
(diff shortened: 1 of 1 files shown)


amirmahdifard:
@BlindMaster24 explane what does this pull request do in the discription, so everyone can understand.

beqabeqa473:
Making pr without any testing is not a good idea. Please test and continue with that after that.

kirill-jjj:
@beqabeqa473, @bear101, can confirm this pull request works great!

BlindMaster24:
@kirill-jjj ?

BlindMaster24:
@bear101 ?

bear101:
Why should it switch on both new sound device and new communication devices? Shouldn't it only switch on communication device?

bear101:
But should it switch if it's not their communication device?

BlindMaster24:
> But should it switch if it's not their communication device?

Yes, it absolutely should. The main point is what the user has configured for audio output in TeamTalk's settings.
Users often have different devices for general system sound ('Default Playback Device,' like speakers) and for calls ('Default Communication Device,' like a USB headset).
If TeamTalk's output is set to 'Default Device,' it must follow the system's current 'Default Playback Device.' So, if someone unplugs their main speakers (which were the Default Playback Device), and the system audio switches to built-in laptop speakers, TeamTalk should also output sound there. That’s the expected behavior for a 'Default Device' setting.
My Bluetooth headphones are a prime example of why this is needed. If I turned them off for over 10 seconds, then back on, they'd reconnect and correctly become the system's 'Default Playback Device.' However, for some reason, I'd then have no audio output at all from my PC through the headphones—not just TeamTalk, but all system sounds seemed affected until I specifically went into TeamTalk and re-initialized devices. This change ensures TeamTalk automatically switches its output to the new system default (my reconnected headphones) as soon as they are ready, resolving this frustrating issue and making it work seamlessly, regardless of what the 'Default Communication Device' is.

BlindMaster24:
@bear101 I can def say these changes will make it way easier for people.
Lots of ppl tested it in the test build and they liked it.

BlindMaster24:
@bear101 ?

BlindMaster24:
> But should it switch if it's not their communication device?

?

bear101:
@CoBC do you agree with this change? Looks ok to me

BlindMaster24:
@CoBC Thanks!

Issue #2623, opened 2025-04-24
Volume Indicator Displays Default Value After Restart Until Manual Adjustment


lbk2907:
### Description

In TeamTalkDroid 5.18.0.5155, when setting a custom volume level (e.g., 35%), the UI displays the correct value initially. However after restarting the client:
1. The volume indicator incorrectly shows the default value (50%).
2. The actual volume remains at the previously set level (35% in this case)
3. The UI only updates to display the correct volume level after manually adjusting it.
This is a bug that creates user confusion about the actual volume state.

### Application

- [ ] qtTeamTalk
- [x] TeamTalkAndroid
- [ ] iTeamTalk
- [ ] TeamTalkClassic
- [ ] TeamTalkServer

### Platform

- [ ] Windows
- [ ] macOS
- [x] Android
- [ ] iOS
- [ ] Linux

### Expected behavior

The volume indicator should:
1. Persist and display the last-set volume level after app restart.
2. The UI immediately display the correct volume level without requiring manual adjustment.

### Actual behavior

1. On restart, the volume indicator shows default volume (50%).
2. Maintains previous volume level in background.
3. Requires manually adjust the volume for the volume indicator to display the correct volume level.

### Steps to reproduce problem

1. Launch TeamTalkDroid 5.18.0.5155 and connect to any server.
2. Adjust volume to any non-default level (e.g., 35%).
3. Restart TeamTalk and connect to any server.
4. Observe volume indicator shows default value (50%).
5. Manually adjust the volume (increase/decrease).
6. Observe volume indicator shows current volume level (in this case is 34% if you decrease or 36% if you increase).

### Additional Context
Device: Google Pixel 7 Pro
Android Version: 16Beta4
Build Number: BP22.250325.007
Note: From what I know, my friend who uses Android 12 is also experiencing this issue.

amirmahdifard:
@lbk2907 hi. actually, befor anyone reports this issue, I have a good news for you. while I was convurting volume and microphone gain's turn up/down buttons to seekbars, I've fixed all these issues in the new system. Do you want a test apk file to try the new system?

lbk2907:
Sure if you can provide. I not have enough space on my computer for me to build it by myself.
@bear101 You can take a look as well.

lbk2907:
Actually I forgot to mention that this issue occurs only with the microphone gain indicator.

amirmahdifard:
[...]

PR #2656, opened 2025-05-20
Android. Fixed bug where microphone gain value was not saved

amirmahdifard:
this fixes https://github.com/BearWare/TeamTalk5/issues/2623


Client/TeamTalkAndroid/src/main/java/dk/bearware/gui/MainActivity.java
```diff
@@ -549,6 +549,7 @@ protected void onStop() {
         if (mConnection.isBound()) {
             SharedPreferences.Editor editor = prefs.edit();
             editor.putInt(Preferences.PREF_SOUNDSYSTEM_MASTERVOLUME, ttclient.getSoundOutputVolume());
+            editor.putInt(Preferences.PREF_SOUNDSYSTEM_MICROPHONEGAIN, ttclient.getSoundInputGainLevel());
```
bear101, line 552: If channel is configured with `AudioPreprocessor` (in TeamTalkService line 799) then you'll end up saving the `AudioPreprocessor` gain value. `Preferences.PREF_SOUNDSYSTEM_MICROPHONEGAIN` should only be saved if the user changes the value.

amirmahdifard:
@bear101 ok, So we can just close this pr, But I don't know Why in new tt sdk 5.18, we have this problem of microphone gain not saving. In 5.15 it worked fine

PR #2683, opened 2025-05-28
Add hotkey to speak ping status

BlindMaster24:
Adds a configurable hotkey to announce the current ping status via Text-To-Speech.

This feature provides a convenient way to check ping status, and crucially, it improves accessibility as the ping information in the toolbar is not readily available to screen reader users.

### DIFF
--- a/ChangeLog.txt
+++ b/ChangeLog.txt
@@ -4,7 +4,7 @@
 
 Version 5.18.1, unreleased
 Default Qt Client
--
+- Added a hotkey to announce the current ping status
 Android Client
 - Option to automatically join root channel in preferences
 - Added the ability to control user's transmissions
--- a/Client/qtTeamTalk/mainwindow.cpp
+++ b/Client/qtTeamTalk/mainwindow.cpp
@@ -2423,6 +2423,56 @@ void MainWindow::hotkeyToggle(HotKeyID id, bool active)
             else
                 showMinimized();
         }
+            case HOTKEY_SPEAK_PING:
+        if (active)
+        {
+            ClientStatistics stats = {};
+            if (TT_GetClientStatistics(ttInst, &stats))
+            {
+                int ping = stats.nUdpPingTimeMs;
+                QString pingMsg;
+                if (ping >= 0)
+                {
+                    pingMsg = QObject::tr("Current ping is %1 milliseconds.").arg(ping);
+                }
+                else
+                {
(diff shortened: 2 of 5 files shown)


kirill-jjj:
> Not sure to understand the goal of this change. Why don't just use option to speak client stats (ctrl+t), or use the screen-reader shortcut to read status bar?

"Because that way we can only find out rx, tx, and this hotkey is for ping, latency to the server. we can check it on android, but not windows."

CoBC:
The ping is not available all the time, when it is available, it's returns in client stats, so CTRL+T or status bar reading will announce it

kirill-jjj:
> The ping is not available all the time, when it is available, it's returns in client stats, so CTRL+T or status bar reading will announce it
tried it my self write now
RX: 13.8 kB, TX: 14.0 kB
i'am online on my server, talking with other person, so I guess ping should appear, as all things for it to appear is done. But here is no ping.
what should I do for it to appear by ctrl+t?

kirill-jjj:
> > The ping is not available all the time, when it is available, it's returns in client stats, so CTRL+T or status bar reading will announce it> > 
tried it my self write now
> > RX: 13.8 kB, TX: 14.0 kB
> > i'am online on my server, talking with other person, so I guess ping should appear, as all things for it to appear is done. But here is no ping.
> > what should I do for it to appear by ctrl+t?

amirmahdifard:
@BlindMaster24 really thanks for your feature of announcing ping, I also never see the ping by ctrl+t. But, A thing. Please, instead of adding another shortcut for only announcing ping, make ctrl+t also announce ping always, like, rx: blaablaa. Tx: blaablaa. Ping: 23. Before I asked this, But it was ignoard.

BlindMaster24:
> @BlindMaster24 really thanks for your feature of announcing ping, I also never see the ping by ctrl+t. But, A thing. Please, instead of adding another shortcut for only announcing ping, make ctrl+t also announce ping always, like, rx: blaablaa. Tx: blaablaa. Ping: 23. Before I asked this, But it was ignoard.

Ok

BlindMaster24:
@CoBC @bear101 Looks like the method isn't returning ping info. Maybe needs to wait for the ping calculation to finish before creating the response string?

BlindMaster24:
@bear101 I think you should check out this function in the SDK.

PR #2686, opened 2025-05-30
Fixed some TTS events not available if only push notification are enabled

CoBC:
If TTS engine is defined to "None" and push notification are enabled, following events are unavailable:
- Speak Client Statistics
- Speak User or Channel Info
- Speak Channel Statistics
- Enable/Disable Text-to-Speech Events

This affects especially Linux users which use push notification to received TTS events and no TTS Engine


bear101:
The MacOS build fails

PR #2687, opened 2025-05-31
Qt client. Max profiles limit increaced

amirmahdifard:
yesterday, someone asked for maximum profiles limitation should've been removed, because they have over 100 150 servers, and they don't like to use server list to add their new servers, they want to create a new profile for each server, and connect to all of them at once. So, I've increaced the maximum profiles entrys to 1000. But if you want, you can just remove the limitation completely.

### DIFF
--- a/Client/qtTeamTalk/mainwindow.cpp
+++ b/Client/qtTeamTalk/mainwindow.cpp
@@ -4088,7 +4088,7 @@ void MainWindow::slotClientNewInstance(bool /*checked=false*/)
     // load existing profiles
     QMap<QString, QString> profiles;
     QStringList profilenames;
-    const int MAX_PROFILES = 16;
+    const int MAX_PROFILES = 1000;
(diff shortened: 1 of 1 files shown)


bear101:
If you have 1000 servers then you should use .tt files and not profiles.

amirmahdifard:
@bear101 yes you are true, but one of my friends asked me a strange request from teamtalk. He said he has over 50 60 servers hosted, and he don't want to connect to one of them at a time and switch between them, he wants to create a profile per each server, and open all of them at once, and be able to connect to all of his servers at the same time. So, he only adds one server in each profile. So, he asked me to increace the max profiles limit to 1000, and I did.

bear101:
@amirmahdifard please stop posting every day. There's too much spam.

amirmahdifard:
@bear101 ok, i'm not spamming. You asked me a question and I answered your question.

PR #2688, opened 2025-06-01
fix: Correct ping announcement

BlindMaster24:


BlindMaster24:
@bear101 If you have questions, happy to answer.

Client/qtTeamTalk/mainwindow.h
```diff
@@ -125,6 +125,7 @@ class MainWindow : public QMainWindow
     QLabel* m_pinglabel;
     QLabel* m_dtxlabel;
     QProgressBar* m_dtxprogress;
+    int m_lastPing;
```
bear101, line 128: Why do you need this member variable when there already a `m_clientstats`?

BlindMaster24: > Why do you need this member variable when there already a `m_clientstats`?

@bear101 Hi! I've made this change to address an accessibility issue with how ping was announced via Text-To-Speech.
In the previous version, the ping was only audible via TTS if the slotSpeakClientStats function was triggered at the precise moment m_clientstats.nUdpPingTimeMs held a valid ping from a network update. This window is very brief. Sighted users might see the ping value update on the label (m_pinglabel) for a short period, but for visually impaired users relying on TTS, it often meant the ping announcement was missed entirely because the timing had to be perfect.
To make this more reliable and ensure visually impaired users can consistently get the ping information, I introduced m_lastPing. This variable now caches the last known valid ping value. This way, even if slotSpeakClientStats is called between the brief network ping updates, it uses this stored valid ping for the TTS announcement. This is somewhat similar to how an Android app might persistently display the latest known ping. The visual m_pinglabel still shows the immediate (and possibly transient -1) ping, but m_lastPing ensures the TTS provides a consistently useful value.

BlindMaster24:
@bear101 RX: 0 bytes, TX: 0 bytes, PING: 38
RX: 0 bytes, TX: 0 bytes, PING: 46
RX: 0 bytes, TX: 0 bytes, PING: 59

CoBC:
Still not in favor of this change. Why we should have a value which sighted users don't have? More, the ping can change everytime, so a ping of 20 stored doesn't reflect the real current value when I check the ping 10 seconds later.

kirill-jjj:
> Still not in favor of this change. Why we should have a value which sighted users don't have? More, the ping can change everytime, so a ping of 20 stored doesn't reflect the real current value when I check the ping 10 seconds later.

The core issue is that for screen reader users, getting any ping reading is incredibly difficult right now. It only briefly appears when a keepalive packet arrives, you need to spam ctrl+t to catch it. If @blindmaster24 could implement a truly real-time ping, that would've been the way, but TeamTalk updates ping, when keepalive packetr is sent.
So, while a stored value isn't perfectly live, having a recently updated ping is a massive improvement over struggling to get one at all. The Android client is handling ping similarly for years already, and noone saw the problem in that.

kirill-jjj:
> Still not in favor of this change. Why we should have a value which sighted users don't have? More, the ping can change everytime, so a ping of 20 stored doesn't reflect the real current value when I check the ping 10 seconds later.
 
Oh, and it updates faster than you think; if you check the ping 10 seconds later, it will actually change. It is not that slow.

beqabeqa473:
@CoBC i also think it's better than nothing. First of all we should do your life easier. This is the same if you ask why we need dedicate commands for screenreader to get time and battery info? sighted users can seevthem in tray, so go ahead with that approach.
Untill new ping is available user should have a possibility to see current one at least. First of this increases productivity. And if you yourself are AT user, this is very strange from you especially.

amirmahdifard:
yes, exactly. I also agree that current ping value should be available to tts. before, I actually never managed to hear the ping at all, And I thort it's a bug. rx/tx is always available, so ping should also be available when pressed it, we want to hear the current ping whenever we press this key.

BlindMaster24:
@CoBC ?

amirmahdifard:
@BlindMaster24 don't spam. Everyone has already saw this, so just wait.

BlindMaster24:
@bear101 Sorry to bother you, but what's the plan? It seems like this might not make it into 5.18.1

BlindMaster24:
@bear101 @CoBC
Any news?

Client/qtTeamTalk/mainwindow.h
bear101: Instead of introducing `m_lastPing` then check `m_clientstats.nUdpPingTimeMs  > 0` and announce that instead. If it's not there then don't announce it.

@CoBC with this change then ping time is announced with TTS even if it's not new. Do we really want that?

BlindMaster24: > Instead of introducing `m_lastPing` then check `m_clientstats.nUdpPingTimeMs > 0` and announce that instead. If it's not there then don't announce it.
> 
> @CoBC with this change then ping time is announced with TTS even if it's not new. Do we really want that?

@bear101 We have no other choice because blind users cannot reliably capture the real ping value. Otherwise, as discussed, they would have to spam Ctrl+T to get the ping.

BlindMaster24: > Instead of introducing `m_lastPing` then check `m_clientstats.nUdpPingTimeMs > 0` and announce that instead. If it's not there then don't announce it.
> 
> @CoBC with this change then ping time is announced with TTS even if it's not new. Do we really want that?

@bear101
As was discussed again, this has been done this way in the Android client for a very long time now. We have no other choice. Everyone is fine with it on Android.

CoBC:
No, I'm still not in favor of this change. Ping should remain announced only if it's new. We should have access with TTS as the same info of other people.

bear101:
We keep the existing functionality

kirill-jjj:
@CoBC @bear101 So, let me get this straight: you're suggesting that blind users have to get the ping by repeatedly pressing a key? Is that correct? You don't cair?
do something with android version then, ping works there just like that.

BlindMaster24:
@bear101 If you're tired after your 20 years of fixing small things and so on, here I am, I'm ready! I have the motivation to fix this. What's the problem?

bear101:
As a visible user you only see the ping when it's updated. With your change the ping is always announced but the user does not know if it's valid. Or when it was valid. Also it's not reset after disconnect so you'll be announcing an old/invalid ping value. This is a feature change you haven't thought through.

BlindMaster24:
@bear101
Your points about the ping value's validity and the lack of a reset on disconnect are correct. I did not consider those failure modes.
However, the core accessibility problem remains: the current implementation makes it nearly impossible for screen reader users to get a ping reading. My solution, while imperfect, is the only one on the table that addresses this.
How would you solve this accessibility issue correctly?

BlindMaster24:
@bear101 ?

BlindMaster24:
@bear101 I am prepared to implement these changes correctly.
Could you please re-open this pull request so I can push the updated, correct implementation?

PR #2691, opened 2025-06-06
Default Qt Client: Limit OPUS bitrate range to 6-510

hwangsihu:


bear101:
Why this change?

`Rates from 500 to 512000 bits per second are meaningful, as well as the
  special values #OPUS_AUTO and #OPUS_BITRATE_MAX.`
https://github.com/xiph/opus/blob/main/include/opus_defines.h#L278

hwangsihu:
Opus officially supports average bit-rates only in the range 6 kb/s – 510 kb/s.
Values outside this window are ignored or clamped during SDP negotiation, and if the encoder is asked for less than 6 kb/s it usually returns an error, so the channels never get updated.

bear101:
TeamTalk.h defines  [OPUS_MIN_BITRATE](https://github.com/BearWare/TeamTalk5/blob/master/Library/TeamTalk_DLL/TeamTalk.h#L1198C8-L1198C26). I suggest to use this value instead.

bear101:
And `OPUS_MAX_BITRATE` for max bitrate.

Client/qtTeamTalk/channeldlg.cpp
```diff
@@ -20,6 +20,7 @@
 #include "utilui.h"
 #include "settings.h"
 
+#include <TeamTalk.h>
```
bear101: You don't need this include statement.

Client/qtTeamTalk/channeldlg.cpp
```diff
@@ -115,7 +116,7 @@ ChannelDlg::ChannelDlg(ChannelDlgType type, const Channel& chan, QWidget * paren
     ui.opus_channelsBox->addItem(tr("Stereo"), 2);
     ui.opus_appBox->addItem(tr("VoIP"), OPUS_APPLICATION_VOIP);
     ui.opus_appBox->addItem(tr("Music"), OPUS_APPLICATION_AUDIO);
-    ui.opus_bpsSpinBox->setRange(1, 512);
+ui.opus_bpsSpinBox->setRange(OPUS_MIN_BITRATE / 1000, OPUS_MAX_BITRATE / 1000);
```
bear101: Please indent 4 spaces.

bear101 (changes requested):
Looks good. The division later of 1000 in ChannelDlg::getChannel() ensures the correct result.

bear101:
Thanks!

PR #2696, opened 2025-06-08
ServerList refactor

beqabeqa473:
Made several changes in Server list dialog to modernize code and ui:
* Extract remove server dialog into its own method
* Refactor server export, reduce code repetition
* Refactor load servers from the network using executor instead of deprecated AsyncTask
* Refactore versionCheck to use executorService instead of now deprecated AsyncTask
* Migrate to Recycler view for displaying list of servers instead of using ListFragment
* Add possibility to filter servers using a search field on top of list to quickly find needed server, especially if user has unofficial servers checked or lots of local servers added.
* Add accessibility actions for edit, export and remove server. Now it is possible to use vertical swipes with recent Talkback to have the similar UX s in iOS
* General code cleanups and refactorings

### DIFF
--- a/Client/TeamTalkAndroid/src/main/java/dk/bearware/gui/ServerListActivity.java
+++ b/Client/TeamTalkAndroid/src/main/java/dk/bearware/gui/ServerListActivity.java
@@ -67,6 +67,8 @@ import java.io.StringReader;
 import java.util.Collections;
 import java.util.Comparator;
 import java.util.Vector;
+import java.util.concurrent.ExecutorService;
+import java.util.concurrent.Executors;
 
 import javax.xml.parsers.DocumentBuilder;
 import javax.xml.parsers.DocumentBuilderFactory;
@@ -97,6 +99,7 @@ extends AppCompatActivity
     ServerEntry serverentry;
 
     private ServerListAdapter adapter;
+    private ExecutorService executorService;
 
     public static final String TAG = "bearware";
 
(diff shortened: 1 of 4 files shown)


beqabeqa473:
@bear101 Please look at visuals to make sure all is displayed correctly.

amirmahdifard:
@beqabeqa473 nice! I stil haven't wrote the user accounts activity, But I was planning to write it like this, Do you want to do it So I can work on others like server prop activity etc?

amirmahdifard:
talkback actions should also be available for users and channels items.

beqabeqa473:
> talkback actions should also be available for users and channels items.

Yes, not part of this pr though.

beqabeqa473:
> @beqabeqa473 nice! I stil haven't wrote the user accounts activity, But I was planning to write it like this, Do you want to do it So I can work on others like server prop activity etc?

I don't know how time will permit. This is relatively old work i've just decided to make it available.

amirmahdifard:
@beqabeqa473 oh, sad then. I asked it because I'm currently working alone for android client, so implementing those new activitys would take so much time for a alone person. So I was looking for someone who could help out a little, e.g making one of the activitys, banned users, user accounts, etc, While I do the rest.

beqabeqa473:
Frankly speaking, to make process to move forward TeamTalk client for android should be rewritten from the ground up. We have a very large technical debt, and we need to change somethingы to fix this.

This is not a topic of this pr though.

amirmahdifard:
@beqabeqa473 no, I don't see any reason and any problem to rewrite the teamtalk android, it's already working fine, Just the new activitys should be added, Which I'll just do them alone if others don't have time to do it.

bear101:
Nice improvement! Looks good.

PR #2698, opened 2025-06-11
Use TTS to announce subscription changes

bear101:
@amirmahdifard  can you give this a try and see if it works as expected?


amirmahdifard:
@bear101 I haven't tryed it on my device yet, but looking at the code, I have a question. Why do you want to add all these announcements in to the utils activity? and also, I think the one that I added was more simpler. It simply checks throo all subscriptions and used a turnery operator to announce subscription, + it's off or on. But I think in this implementation the code has became more complicated. Did you have a reason to implement it like this?

amirmahdifard:
@bear101 I've just tryed it on my phone. It doesn't work at all. Completely doesn't work, and doesn't announce anything. Also, about togling transmission controles on user prop something is broken as well. When opening the profile and togling the transmit switch on or off, it works and changes the transmission actually, but the tts is not announced. Tts will only be announced when you toggle it more than one time. For example, if I open my user prop and toggle my voice transmission off, it does it but doesn't speek anything in my android client, but When I press the switch once more to toggle my voice back on with out closing the profile, it says amir can now transmit voice. But in windows when I toggle transmission for a user, the android client announces it fine.

bear101:
@amirmahdifard please try again

amirmahdifard:
@bear101 after @beqabeqa473 's merged pull requests, I can no longer build the app. I tryed this pr and also the laytist master, but it gives this error.
Gradle Sync issues  selected  20 of 20  level 3
I tryed resinking gradle, but It didn't solve the problem. Any thorts or ideas?

amirmahdifard:
the only change in gradle file that has happened is that adds view binding, but I don't know why I get that error.

amirmahdifard:
@beqabeqa473 hi. I'm getting this error, And I've looked at all gradle files to find where is using this line that makes android studio to look for it, But I didnn't find anything to remove. hear is the error. Please help out. Thanks.
Could not find androidx.databinding:viewbinding:8.1.2.
Required by:
    project :
Search in build.gradle files

amirmahdifard:
I'm also wundering why we used view binding for this, why we didn't use find view by id like others, like channel prop activity?

bear101:
@amirmahdifard you need to clean your build configuration and build again. That worked for me.

amirmahdifard:
@bear101 That was the first thing I tryed. I tryed the following.
1. Completely deleted my build configurations/cashes, invalidated and restarted, also cleared the gradle cash files, didn't work.
2. I tryed messing up with gradle version, gradle files, didn't work, Apparently It's looking for a doesn't exist library, Which I pasted the error above.

amirmahdifard:
I just pulled laytist version, and openned the directory in android studio, and tried to build, but getting "Could not find androidx.databinding:viewbinding:8.1.2 error.".

amirmahdifard:
@bear101 Thanks! It's been fixed. Do you know what was the problem? the problem was that I live in iran, and recent fights between iran and israel made internets in iran limited and restricted, thus android studio couldn't download dependencys properly with out a good vpn! haha

beqabeqa473:
the aim is to avoid findViewById all together. Some more bunch of changes will be soon.

amirmahdifard:
@bear101 I've checked this now again, this is now working. they are being announced now. Thanks, but the problem with the transmissions controles stil exist.

amirmahdifard:
@beqabeqa473 nice working, keep up the good work! but to things. The first one very important. While you replace find by ids in channel prop activity, please don't forget to reflect my changes on it, I coded 2 weeks of my time on it. I got rid of that stupid audio codec activity, and added all newly added options, so remember to get it from my open pull request and work on it instead of that old one that's in the master, Thanks! second, Is binding better than find by id? I'd like to know.

amirmahdifard:
ok, I've red the difrences between view binding and find view by id. It's grate and better in all cases, but only that find by id can be used with out any enabling a extra step in build.gradle and works more simpler. view binding just uses les code and faster righting etc

beqabeqa473:
Please, don't flood prs with unrelated comments. Couple is ok, more is bad.

amirmahdifard:
ok, i'm sorry

PR #2635, opened 2025-04-30
Android client: Introduced new options to settings with their abilitys

amirmahdifard:
1. Added the ability to play sound when voice activation button is pressed like tx button, with the checkbox in sound events settings to be able to toggle, respectivly.
2. As so many user requests that it's risky to have private conversations with teamtalk android as it does not announce and notify intercept and subscription events, added the ability to play sound and announce intercept events same as desktop client, and announce subscription events again same as desktop clients, and added options for all of them in to settings respectivly, so users can turn them on and off.
3. Added announcements of transmit events in channels with their checkboxes in tts events settings, respectivly.
4. Another bug has been fixed where media file volume slider for people in their user props are not set to the media file volume set in settings.
This pull request is now finished, please merge as soon as possible, So I can do A clean uptodate of my fork with teamtalk's master, So I can open my next pull requests much easy.

PR #2699, opened 2025-06-12
Play sound event when intercept subscription is toggled

bear101:
Applies to Android client.

Only last commit applies to this merge request since it's based on #2698.


bear101:
@amirmahdifard this should conclude #2635.

PR #2705, opened 2025-06-18
Android client: Added the ability to set status messages.

amirmahdifard:
As promiced before, i've added the ability to set status message for android client, and more surprise, is that I've desighned it like nickname, Which means you'll be able to set difrent status messages for each server seperatly if you want, and if you don't, your general status message will be used, exactly like nickname. Also, status messages for servers are stored on .tt files as well, so no worrys about losing them! enjoy!

### DIFF
--- a/ChangeLog.txt
+++ b/ChangeLog.txt
@@ -15,6 +15,7 @@ Default Qt Client
 - Fixed crash issue in "Connect to a Server" dialog on macOS 15.1
 - macOS updated to Qt 6.10.0 beta1
 Android Client
+- Added the ability to set status messages
 - Option to automatically join root channel in preferences
 - Added the ability to control user's transmissions
 - Ability to play sound when voice activation is toggled
--- a/Client/TeamTalkAndroid/src/main/java/dk/bearware/backend/TeamTalkService.java
+++ b/Client/TeamTalkAndroid/src/main/java/dk/bearware/backend/TeamTalkService.java
@@ -876,17 +876,21 @@ public class TeamTalkService extends Service
             //update status bar widget
             displayNotification(true);
 
-            // check whether to switch to female icon
+            // check whether to switch to female icon and put status message per server
             SharedPreferences prefs = PreferenceManager.getDefaultSharedPreferences(getBaseContext());
             int statusmode = TeamTalkConstants.STATUSMODE_AVAILABLE;
-            String msg = "";
+            String statusmsg = "";
             User myself = users.get(ttclient.getMyUserID());
             if (myself != null) {
                 statusmode = myself.nStatusMode;
-                msg = myself.szStatusMsg;
+                statusmsg = ttserver.statusmsg;
+            }
+            if (TextUtils.isEmpty(statusmsg)) {
+                statusmsg = PreferenceManager.getDefaultSharedPreferences(getApplicationContext()).getString(Preferences.PREF_GENERAL_STATUSMSG, "");
             }
             if (prefs.getBoolean(Preferences.PREF_GENERAL_GENDER, false))
-                ttclient.doChangeStatus(statusmode | TeamTalkConstants.STATUSMODE_FEMALE, msg);
+                statusmode |= TeamTalkConstants.STATUSMODE_FEMALE;
+            ttclient.doChangeStatus(statusmode, statusmsg);
         }
     }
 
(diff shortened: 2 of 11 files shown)


bear101:
It seems this change introduces a server-specific status message. There is not such a thing. There's only a global status message. Please only stick to the general (global) status message.

kirill-jjj:
> As promiced before, i've added the ability to set status message for android client, and more surprise, is that I've desighned it like nickname, Which means you'll be able to set difrent status messages for each server seperatly if you want, and if you don't, your general status message will be used, exactly like nickname. Also, status messages for servers are stored on .tt files as well, so no worrys about losing them! enjoy!

how the other clients will read .tt files with specific status messages which will work only on android?

MahdiDabaghi:
hello. I honestly, don't know why you disagree with such an implementation. As I know, so many people was requesting this feature, and it would be good if it would be implemented in desktop clients as well.
moreover, As the description of the pr suggests and we can see clearly, the feature specific status messages per servers is an optional feature. It means, who don't like this feature, they can simply not use it. As I see, when you set a global status message in general settings, and then don't set any status message for any server and leave the edit box blank, your general global status message will be used for all servers with out any problem or interruption. So it's clear that who likes, will use this feature, and who doesn't like, will be able to simply skip it. As long as something is optional and useful at the same time, there should be nothing preventing it from being accepted.
actually, I was also one of the people who requested him to implement this feature because,
1. It's possible, like it's possible for nickname to be like this.
2. While you disagree with this for an unknown reason, some people find it useful. Sometimes I want to put a different status message for my class server, while I don't want that status in my other servers like when I'm around my friends in another server.
So, I kindly request you to please accept this feature as it is, because it's optional and won't bother anyone, people who don't want it, can simply not use it at all. Thanks. Hope I've explained it enough and fairly.
please always provide a valid reason when you disagree with something.

BlindMaster24:
@MahdiDabaghi The problem is that this isn't even in the qt client. And since it's not there, I am in favor of the global status.

BlindMaster24:
@bear101
Perhaps I could add individual status messages to the Qt client?
Firstly, I can propose use cases, plus it's just a cool feature, and secondly, we would help push this PR forward.

MahdiDabaghi:
exactly, I also think this feature should be added to qt client as well, because this feature is full of good points, not even a single bad point.
But Currently, @amirmahdifard is facing internet limitations in iran due to recent fights, but in our teamtalk server, he told me to tell you guys to accept this pr, and he will add it to qt client as well if no one else is willing to do it, but if you want to do it, then glad!

amirmahdifard:
@bear101 Can I know the reason why you disagree with this? the feature is absolutely cool, both for android and windows clients.
@kirill-jjj Once this is merged for android, I'll add the same feature exactly for qt client and aline it between clients, so tt files extracted in android client will have the same usage for qt client as well.
@bear101 please merge this, as soon as the ristrictions from iran's internet is lifted or my vpn's status became better, i'll add this to qt client because it's very cool and useful, + many people has requested me to implement it like this while i'm at it.

bear101:
@CoBC do you think it's useful to also have a specific status message for each server? (Like there's currently nickname)

amirmahdifard:
be sure that it's useful! so many people has requested me to add this feature, since you login to different servers, you can have different status messages in each one. And you can not use it simply if you don't want, like you do with nickname. I'm sure this feature won't annoy anyone, so anyone won't be disagree with it. And It can be one of the new features of the new version of teamtalk!

CoBC:
Yes, I think it could be a good idea. Do you want me to work on it for Qt client?

amirmahdifard:
horay! thanks corentin! please work on it for qt client while I'm trying to fix my limited internet, and this pull request is ready and can be merged for android, the implementation needs to be done only on qt client now.

abolfazlghasemi0033:
I hope all is well in your world. This idea is quite interesting and, importantly, entirely optional. In my view, Amir Mahdi Fard is absolutely right. The fact that participation is optional ensures that no issues will arise, and it also introduces a feature that many people have long been anticipating.
I do not believe that unjustified opposition to such an initiative serves any purpose. What truly matters, in my opinion, is that this feature adds a positive dimension to the program. It can also enhance the organization of classes, discussion groups, and other such categories, thereby fostering a sense of greater functionality and efficiency.

bear101:
> Yes, I think it could be a good idea. Do you want me to work on it for Qt client?

Yes, can you put it in the Qt client. Both for server entry dialog and generate .tt file dialog.

PR #2710, opened 2025-06-25
Android client. Added a new option to prevent talking users to be moved to the top of users list

amirmahdifard:
So many screenreader users are annoyed by this fact that talking users are moved to the top of the list of users, because in a channel that has so many users and users are constantly opening and closing their microphones, the blind user will end up messaging, moving, kicking etc, taking action on a wrong user or they cannot do that because users are constantly moving up and down.
      Added a new option, checked by default to not break current way for sited users, which if unchecked, users will no longer move up and down, which makes it stable for blind users. Sited users or whoever wants, they can check this box, and who don't wish, can uncheck it.

### DIFF
--- a/ChangeLog.txt
+++ b/ChangeLog.txt
@@ -15,6 +15,7 @@ Default Qt Client
 - Fixed crash issue in "Connect to a Server" dialog on macOS 15.1
 - macOS updated to Qt 6.10.0 beta1
 Android Client
+- Added a new option to prevent talking users from moving to the top of users list
 - Added the ability to set status messages
 - Option to automatically join root channel in preferences
 - Added the ability to control user's transmissions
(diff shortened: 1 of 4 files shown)


amirmahdifard:
@bear101 Can you look please?

bear101:
Code says default value is false:
`PreferenceManager.getDefaultSharedPreferences(getBaseContext()).getBoolean("movetalk_checkbox", false)`
UI says default value is true:
`android:defaultValue="true"`

So which one is it? Backwards compatible is `false`.

amirmahdifard:
@bear101 I've changed it to true in ui in code to make it true by default for the user. But ok, i'll fix it

amirmahdifard:
Done! Now they can read each other properly, Sorry I was forgot that before lol.

amirmahdifard:
Please merge this so I can work on other features before israle starts attacking iran again lol

bear101:
Lets hope the Israelis are taking a break 🙈

PR #2716, opened 2025-06-27
Use blank value for nickname and status message is same as general values

CoBC:


Client/qtTeamTalk/mainwindow.cpp
```diff
@@ -4591,7 +4592,7 @@ void MainWindow::slotMeChangeStatus(bool /*checked =false */)
     QString statusmsg = ttSettings->value(SETTINGS_GENERAL_STATUSMESSAGE).toString();
     if ((TT_GetFlags(ttInst) & CLIENT_AUTHORIZED) && m_host.statusmsg.size())
         statusmsg = m_host.statusmsg;
-    ChangeStatusDlg dlg(statusmsg, this);
+    ChangeStatusDlg dlg((statusmsg != ttSettings->value(SETTINGS_GENERAL_STATUSMESSAGE)?statusmsg:""), this);
```
bear101: Isn't there a `.toString()` missing:
`ttSettings->value(SETTINGS_GENERAL_STATUSMESSAGE).toString()`

bear101: Also why is the `else` empty string? Shouldn't it be `ttSettings->value(SETTINGS_GENERAL_STATUSMESSAGE).toString()` ?

CoBC: My idea is to leave blank value if the current value is the same as global value, so user knows that no specific value is set for this host.

PR #2717, opened 2025-06-27
Android Client: Update Dependencies

hwangsihu:


amirmahdifard:
this doesn't build
The project is using an incompatible version (AGP 8.10.1) of the Android Gradle plugin. Latest supported version is AGP 8.3.0
See Android Studio & AGP compatibility options.

amirmahdifard:
I was usimg android studio 2023. I'm upgrading to 2025 to see that works or not

bear101:
@beqabeqa473 and @poretsky, are you ok with this upgrade? I have it up and running with Android Studio 2025.1.1

hwangsihu:
> I was usimg android studio 2023. I'm upgrading to 2025 to see that works or not

Yeah, it works fine in version 2025 and it seems to be functionally fine.

hwangsihu:
Hmm, by the way, I made a commit a few hours ago to remove unused resources, but it failed because of unused translation strings. If someone could take a look at it, I think the file size could be reduced a bit.

beqabeqa473:
It is ok for me, i am not using android studio, but even for studio users, it should be ok.

amirmahdifard:
yeah apparently, I've tryed it with android studio 2025.1.1 and it seemze to work

hwangsihu:
Functionally it seems 99% verified... LOL

amirmahdifard:
But I have a question. Why did you touch other places also? like, added network timeout 10000? can you please just update the dependencys and don't touch other things? I'm afrade It'll break some things.

hwangsihu:
It seems like this happens when you update Gradle Wrapper to the latest version, but it doesn't seem to be a problem.

PR #2722, opened 2025-07-01
Android client. Fixed away status and added Channel operator

amirmahdifard:
Fixed the case if a user's status was away and they also had a status message, screenreader just said away and didn't read the status message.
2 Added indication of Channel operator if user is channel operator.
3. Cleaned up the lojic, instead of hundrids of if and else blocks, used clean and simple turnery operators to handle these much more cleanly, + the code is now much faster, and more readable.

### DIFF
--- a/Client/TeamTalkAndroid/src/main/java/dk/bearware/gui/MainActivity.java
+++ b/Client/TeamTalkAndroid/src/main/java/dk/bearware/gui/MainActivity.java
@@ -1299,6 +1299,7 @@ private EditText newmsg;
                 nickname.setText(name);
                 status.setText(user.szStatusMsg);
                 
+                boolean isOperator = ttclient.isChannelOperator(user.nUserID, user.nChannelID);
                 boolean talking = (user.uUserState & UserState.USERSTATE_VOICE) != 0;
                 boolean female = (user.nStatusMode & TeamTalkConstants.STATUSMODE_FEMALE) != 0;
                 boolean neutral = (user.nStatusMode & TeamTalkConstants.STATUSMODE_NEUTRAL) != 0;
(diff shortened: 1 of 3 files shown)


BlindMaster24:
@amirmahdifard I'll say it before the others. Split everything into separate PRs, it will be better for everyone.

amirmahdifard:
@BlindMaster24 why do you comment before even knowing anything about the pr? this pr is a small pr that adds a single feature, not more.

Client/TeamTalkAndroid/src/main/java/dk/bearware/gui/MainActivity.java
```diff
@@ -1309,35 +1310,9 @@ else if(item instanceof User) {
...
-                        icon_resource = away? R.drawable.man_orange : R.drawable.man_blue;
-                        nickname.setContentDescription(name + " 👨");
-                    }
-                    else {
-                        icon_resource = away? R.drawable.man_orange : R.drawable.man_blue;
-                        nickname.setContentDescription(name);
-                    }
-                }
-                status.setContentDescription(away ? getString(R.string.user_state_away) : null);
+                nickname.setContentDescription((talking ? getString(R.string.user_state_now_speaking, name) : name) + (female ? " 👩 " : neutral ? " 🧑 " : male ? " 👨 " : "") + (isOperator ? getString(R.string.user_state_operator) : ""));
```
bear101: This is unreadable...

bear101:
Only add/change on feature per PR.

bear101 (changes requested):

Client/TeamTalkAndroid/src/main/java/dk/bearware/gui/MainActivity.java
amirmahdifard: it does all the works of those if and else blocks. this way adding new indications will be much easy and more simple, 29 lines have been removed and only 4 lines added. first it handles talking state, then gender baced on femail or  newtral, or mail and nothing. Then, it handles operator state. Is it clear now?

beqabeqa473:
This is indeed not readable. Please refactor it other way.

Also, useless commits like indentation inside prs are not needed. Just do interactive rebase, squash commits, and do a force-push. that is ok for prs.

amirmahdifard:
@beqabeqa473 hi. But how Can I refaqter this any other way, with this way, I wanted to avoid those so many confusing if and else blocks. if femail blaablaa if mail talking away blaablaa if talking femail blaablaa. But this way with turnery operator, things are just appended to each other and adding new ones/modifying current ones will be easy. Ok I agree this may be is not readable, but Can it be accepted please? I promice it will work even better than the passed. About readability, I'll exactly explane what it does. first, if talking, it puts now speeking on the name, else just puts the name. Then, first it compares femail gender to newtral gender, and then in another turnery operator, compares mail gender to nothing, empty. Then, it compares that if is operator, call channel operator, and else, do nothing. Done. I've done something that before was done in 29 lines, I've done it in 2 lines. Isn't it better really?

beqabeqa473:
Please refactor it to collect info in a message and set content description once at the end.

This will not be accepted probably, i am against this change myself, because it is missing code clarity.

amirmahdifard:
ok. so you mean make a veriable called message, and then aply these turnery operators on it, and then call setcontentDiscription(message); right?

beqabeqa473:
collect information in a readable manner and set content description afterwards

amirmahdifard:
done! this is very clean and completely readable, let me explane.
I left icon rezorces lojic untouched, just removed a redundent statement because the too statement were exactly doing the same.
But about content discription, there are now a fue strings. speeking, gender, op, and instead of all mixed together. each string collects something seperately, then they have been used in the content discription afterwuds. This is now better as much as possible and I personally like it!

beqabeqa473:
Ok, seems better. Please rebase ontop current master and make a force-push

amirmahdifard:
@beqabeqa473 I'm not using git bash at the moment and using github desktop, and github desktop doesn't have force push feature, But As I know it's ok to merge with current commits, but if I want to do that I must delete this branch and creat a new one lol, i'll do it.

PR #2723, opened 2025-07-02
Fix file transfer speed display

CoBC:


Client/qtTeamTalk/filetransferdlg.cpp
```diff
@@ -101,7 +101,7 @@ void FileTransferDlg::updateFileTransfer(const FileTransfer& transfer)
         throughput = transfer.nTransferred / elapsed;
         throughput /= 1024.0;
     }
-    QString lt = QString(tr("%1/second, last second %2").arg(getFormattedSize(throughput)).arg(getFormattedSize(diff/1024.0)));
+    QString lt = QString(tr("%1/second, last second %2").arg(getFormattedSize(transfer.nTransferred / elapsed)).arg(getFormattedSize(diff)));
```
bear101: If `elapsed` is 0 what happens?

PR #2739, opened 2025-07-06
android. Added popup to online users activity!

amirmahdifard:
@bear101 like you requested, now it is possible to kick, ban, move, message, grant/revoke operator of a user from online users activity, full controle over users, exactly like windows! enjoy!

### DIFF
--- a/Client/TeamTalkAndroid/src/main/java/dk/bearware/gui/OnlineUsersActivity.java
+++ b/Client/TeamTalkAndroid/src/main/java/dk/bearware/gui/OnlineUsersActivity.java
@@ -287,7 +287,6 @@ public class OnlineUsersActivity extends AppCompatActivity implements
     public void onCmdUserLeftChannel(int nChannelID, User user) {
         Log.d(TAG, "User " + user.szNickname + " left channel " + nChannelID);
         updateUser(user);
-        sortAndNotifyDataSetChanged();
     }
 
     @Override


Client/TeamTalkAndroid/src/main/java/dk/bearware/gui/MainActivity.java
```diff
@@ -173,6 +173,7 @@ public class MainActivity
     TabLayout mTabLayout;
 
     public static final String TAG = "bearware";
+    public static List<Integer> userIDS = new ArrayList<>();
```
bear101: Do not make this variable `static`. In `OnlineUsersActivity` you should call the `TeamTalkService.getUsers()` instead

bear101 (changes requested):

Client/TeamTalkAndroid/src/main/java/dk/bearware/gui/MainActivity.java
amirmahdifard: But I needed to create this static list and use it in online users activity so selected for move users are the same in both activitys. So you can long press a user in online users activity, select it for move, and then go back to main activity and long press a channel and move the users. So please tell me what's your problem with this.

amirmahdifard:
I didn't create this static list in online users activity, I just created in main activity and used it in online users activity like this.
                        case R.id.action_select:
                            if (MainActivity.userIDS.contains(selectedUser.nUserID)) {
                                MainActivity.userIDS.remove((Integer) selectedUser.nUserID);
                            } else {
                                MainActivity.userIDS.add(selectedUser.nUserID);
                            }
                            return true;
                            @beqabeqa473 do you understand what is he asking me to do? his comment is not clear to me. I needed to have this public list so selected for move users is always same.

beqabeqa473:
is it the same in qt client?

If yes, then create utility helpers for buffer instead of this array in main activity.

amirmahdifard:
I prefare to remove the move action completely and you can add it yourself as how as you whish, because we have the aray list in main activity already, I just made it public static So I can use it in other activitys.

beqabeqa473:
@amirmahdifard I've done some changes to your branch, rebased ontop of master and force-pushed to cleanup history. you can reset to latest master commit and pull changes if you want some adjustments.

amirmahdifard:
@beqabeqa473 thanks, i'll just wait for the merge

bear101:
@amirmahdifard isn't it better to change `onlineUsers` to a `Map` instead of an `ArrayList`? It looks overly complicated to access a user in the array

amirmahdifard:
@bear101 no, it's not better at all, trust me. In main activity, for storing selected for move users, we have an aray list, named userids. The only thing I did, is that I made that aray list userids public static list in main activity, So I could use and modify it from other activitys, for example in online users activity, by simply calling. Mainactivity.userids.contanes, add, remove, etc. this way, selected for move users is same and sink between online users activity and main activity, so user can select some users for move in online users activity, and then go back to main activity and move the selected users in destination channel. So please trust me that this is the cleanist, best, and correct way and bugless way to implimenting this. So do you allow me to close this pr, open a new clean one, and also add move action as I explaned now and you merge it?

beqabeqa473:
@bear101 i did some changes to this pr, so i can finish it. what would be your suggestions?

amirmahdifard:
@beqabeqa473 Thanks for taking this.

PR #2745, opened 2025-07-12
Android Client: Update Gradle Wrapper and Enabling Configuration Cache

hwangsihu:


bear101:
What does `.github/workflows/update-gradle-wrapper.yml` do?

Also shouldn't we use stable gradle releases instead of release candidates? We don't need to be so "bleeding edge" with gradle.

hwangsihu:
This file automatically generates a PR for upgrading Gradle Wrapper. It is similar to Dependabot.
Also, once 9.0.0 is officially released, how about switching to stable? We are already using the RC version, and until then, it seems best to keep the RC up to date.

bear101:
What is the advantage of Gradle v9 compared to Gradle v8?

I'm not interested in becoming a Gradle beta-tester :)

hwangsihu:
Hmm...
As I mentioned earlier, we are already using the RC version.
Would you like to switch to the stable release?
However, since we have migrated all the code, I'm not sure if we can go back.

bear101:
Is it possible to make `release-channel` point to the stable channel? Or will this break the current builds?

hwangsihu:
I don't know. I'm trying it out.

hwangsihu:
Fortunately, Gradle Script is well compatible.

PR #2747, opened 2025-07-12
Include TeamTalk5 Test files as artifacts in Android Build

hwangsihu:


amirmahdifard:
@hwangsihu please don't disable the exe files for qt teamtalk from artifaqt. I need that. Can I know what you've exactly disabled?

PR #2758, opened 2025-07-19
Fix errors

hwangsihu:


hwangsihu:
https://github.com/hwangsihu/TeamTalk5/actions/runs/16394659634/job/46325249505#step:9:1

@bear101, how do you think this problem can be solved?
Actually, I'm not sure if modifying the code like this is the right thing to do.

bear101:
Why do you remove the log-statements (MYTRACE)? It would be a good indicator of what is going wrong

hwangsihu:
Hmm, is that a problem? GPT seems to have deleted it.

bear101:
What is GPT?

hwangsihu:
It means ChatGPT.
I asked them to modify the script to be compatible with FFmpeg 7.1.1, but they messed it up.
(Probably) FFmpegResampler.cpp seems to be successful, but FFmpegStreamer.cpp is the problem.

beqabeqa473:
If you don't know how to fix, don't ask llm to do it on behave of you. it could create security flaws. Stop doing that.

PR #2759, opened 2025-07-19
Update to ACE v8.0.4

bear101:


hwangsihu:
You may need to delete all patches and recreate them.

bear101:
@beqabeqa473 The musl patch has to be created again. The current one cannot be applied to ACE v8. Since there is no musl based workflow then I cannot test if it works properly.

PR #2760, opened 2025-07-19
Update TeamTalkLib Repository Download Links

hwangsihu:


bear101:
What is the advantage of this change?

hwangsihu:
Advantages? None. However, since WebRTC has .git, I wanted to unify it.

PR #2762, opened 2025-07-20
Build Optimization

hwangsihu:


beqabeqa473:
Hello.

What does this pr improve?
don't use latest, we should leave what is working. If os will be upgraded later, latest tag will be changed, and build may fail because of configuration issues.
latest java apis used in android are from java17, what is reason to use java21?

It would be great for everyone to leave a description to understand purpose of patricular prs.

hwangsihu:
Wouldn't it be better to use the latest LTS?
If not, could you explain why?
I have fixed Ubuntu to 24.04.

PR #2763, opened 2025-07-20
Update depot_tools to latest commit and Changed to use Git for cloning on Windows

hwangsihu:


bear101:
Depot Tools has a feature to automatically update itself but I've disabled the feature [here](https://github.com/BearWare/TeamTalk5/blob/master/Library/TeamTalkLib/build/webrtc/CMakeLists.txt#L224)

We don't know if future versions of WebRTC can build branch branch-heads/6818.

hwangsihu:
It is currently working properly.

hwangsihu:
I did not add an automatic upgrade step, but only added a bootstrapping step. This is because otherwise an error will occur.

bear101:
Do you get a build error with DepotTools `39b2e4ef`?

hwangsihu:
No, but since the previous depot_tools is too old, I think it's okay to keep this commit if there are no errors in the latest version.

hwangsihu:
Also, I discovered a     days     that there's a version mismatch for      _      between Windows and Linux, which    also not ideal.

Issue #2753, opened 2025-07-15
Android client cannot join to typed channel in server info screen


OzancanKaratas:
### Description

Client cannot join any specified channel. No error.

### Application

- [ ] qtTeamTalk
- [x] TeamTalkAndroid
- [ ] iTeamTalk
- [ ] TeamTalkClassic
- [ ] TeamTalkServer

### Platform

- [ ] Windows
- [ ] macOS
- [x] Android
- [ ] iOS
- [ ] Linux

### Expected behavior

Client must join to specified channel as before.

### Actual behavior

After save and join, client cannot join any channel. No error.

### Steps to reproduce problem

1. Create a server on Android client.
2. Go to Join Channel section.
3. Type any channel name, and password if exist on your server.
4. Save, and join the new created server.

amirmahdifard:
i'll fix this

PR #2765, opened 2025-07-22
Android. fixed client not joining the channel specifyed in server prop dialog

amirmahdifard:
@bear101 when you were implementing login complete functions in ttservice, you touched this function and due to some incorrect checks, it was became broken, whitch i've now fixed it.
this fixes https://github.com/BearWare/TeamTalk5/issues/2753


bear101:
Nicely spotted!

PR #2768, opened 2025-07-22
android. Added action to leave channels

amirmahdifard:
like qt client, there is now a seperet leave action that allows you to leave your current channel from everywhere


bear101:
Nice and simple pull-request! Thanks @amirmahdifard

PR #2771, opened 2025-07-24
Android. made user interface better and more sinked with qt client

amirmahdifard:
in qt client, pressing enter on users names opens text message dialog, and you can open user prop dialog and text message dialog, and all other dialogs from the context menu.
So, for the following reasons.
1. Text message dialog is used more than the user prop dialog,
2. For eas of access, faster swipe and usage of screenreaders for blind users,
3. To sink with qt client,
By request, removed the message button from beside of the users, and made it so that klicking on users opens the text message dialog, and opening all of the dialogs are possible from the popup menu. This also makes the dialog modernised and beautiful, i've done the same for server list screen before.

### DIFF
--- a/ChangeLog.txt
+++ b/ChangeLog.txt
@@ -32,6 +32,7 @@ Android Client
 - Ability to play sound for intercept subscription toggled
 - Ability to use Text-To-Speech to announce transmission changes
 - Ability to use Text-To-Speech to announce subscription changes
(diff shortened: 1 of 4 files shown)


bear101:
@beqabeqa473 do you agree with this change?

beqabeqa473:
I would like to see custom accessibility actions for some frequently used features, like message, mute/unmute, kick/ban, move.

This type of change may make some confusions for people i think.
But final decision is up to you @bear101

amirmahdifard:
actually no, people are not used to this way at all.

PR #2772, opened 2025-07-26
Android. Fixed a bug where channel operator is not announced after we make a user channel operator, until a user update

amirmahdifard:


Client/TeamTalkAndroid/src/main/java/dk/bearware/gui/MainActivity.java
```diff
@@ -1569,6 +1569,9 @@ public boolean onMenuItemClick(MenuItem item) {
                 ttclient.getMyUserAccount(myuseraccount);
                 if ((myuseraccount.uUserRights & UserRight.USERRIGHT_OPERATOR_ENABLE) != UserRight.USERRIGHT_NONE) {
                     ttclient.doChannelOp(selectedUser.nUserID, selectedUser.nChannelID, ttclient.isChannelOperator(selectedUser.nUserID, selectedUser.nChannelID)? false: true);
+                    accessibilityAssistant.lockEvents();
```
bear101, line 1572: There are no changes occurring here. The update happens in `onCmdChannelUpdate` (or what the callback is called).

amirmahdifard: But in channel update we do have notify data set changed too, so why it doesn't happen when user becomes channel operator?
    public void onCmdChannelUpdate(Channel channel) {
        if (curchannel != null && curchannel.nChannelID == channel.nParentID) {
            accessibilityAssistant.lockEvents();
            channelsAdapter.notifyDataSetChanged();
            accessibilityAssistant.unlockEvents();
        }

PR #2776, opened 2025-07-28
Android. Added ability to pause and resume media file streams!

amirmahdifard:
@bear101 This is ok, But I have a problem that also corentin had it, and you fixed it. The problem is that when pausing the stream it works, but when resuming it doesn't resume, I think the file position is lost
Can you please fix it so this feature is also added to android? Thanks!


amirmahdifard:
@bear101 are you ok to push this forward so we can have it in 5.19 final update? there are a lot requests for this feature.

amirmahdifard:
@bear101 please also remember to take care of this, so it'll be one of the features of 5.19. In 5.20, we'll have user accounts dialog and server prop dialog hopefully.

amirmahdifard:
@bear101 when do you plan to work on this?

bear101:
No plans so far.

amirmahdifard:
ok, so later prepare this for 5.20. everything else has been done by me, just please fix it so that it'll properly pause and resume, like the problem co bc had, and you managed to fix it.

bear101:
Fix it yourself and stop complaining.

PR #2785, opened 2025-08-03
Update OpenSSL to 3.5.1

hwangsihu:


bear101:
Thanks for this update! We're currently finishing the v5.19 release, so we'll schedule it for v5.20

Library/TeamTalkLib/teamtalk/PacketLayout.inl
```diff
@@ -68,7 +68,6 @@ CryptPacket< PACKETTYPE, PACKET_KIND_CRYPT, PACKET_KIND_DECRYPTED >::CryptPacket
     int encrypt_len = 0, tmpLen = 0;
     EVP_CIPHER_CTX* aesEncCtx = EVP_CIPHER_CTX_new();
     cipher_guard g(aesEncCtx);
-    EVP_CIPHER_CTX_init(aesEncCtx);
```
bear101, line 71: How come you remove this call?

The OpenSSL documentation states:

```
The EVP cipher routines are a high level interface to certain symmetric ciphers.

EVP_CIPHER_CTX_init() initializes cipher contex ctx.

EVP_EncryptInit_ex() sets up cipher context ctx for encryption with cipher type from ENGINE impl. ctx must be initialized before calling this function.
```

PR #2796, opened 2025-08-05
Call slotClientExit() on close event

### LINKED ISSUE #2795: Quit from menu does not exit TeamTalk
### Description

TeamTalk for macOS does not quit after commit 3a07d673c. I suggest reverting the commit.

### Application

- [x] qtTeamTalk
- [ ] TeamTalkAndroid
- [ ] iTeamTalk
- [ ] TeamTalkClassic
- [ ] TeamTalkServer

### Platform

- [ ] Windows
- [x] macOS
- [ ] Android
- [ ] iOS
- [ ] Linux

### Expected behavior

Cmd+Q should quit TeamTalk

### Actual behavior

Pressing Cmd+Q does not exit TeamTalk

Press Cmd+Q

CoBC:
Should fixes #2795
@bear101 Can you test please? I'm not able to test on macOS for now :(

### DIFF
--- a/ChangeLog.txt
+++ b/ChangeLog.txt
@@ -8,6 +8,7 @@ Default Qt Client
 - Reintroduce option to speak tree item using TTS on MacOS due to Qt issues
 - Fixed check state of server log events on macOS not being visible
 - Fixed global status message set to: "general_/statusmsg"
+- Fixed answering "No" to quit application causing server disconnect
 Android Client
 - Minimum supported Android version is now 5.0 (Lollipop)
 iOS Client
--- a/Client/qtTeamTalk/mainwindow.cpp
+++ b/Client/qtTeamTalk/mainwindow.cpp
@@ -8028,11 +8028,11 @@ void MainWindow::keyPressEvent(QKeyEvent* e)
 
 void MainWindow::closeEvent(QCloseEvent *event)
 {
+    QMainWindow::closeEvent(event);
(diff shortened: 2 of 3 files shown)


bear101:
I don't macOS will quit without `QApplication::quit();`

Also this is really bad timing for introducing this. It has not been tested at all.

CoBC:
In this case maybe just replace this->close(); by QApplication::quit(); on slotClientExit()
Otherwise if you prefer just revert my previous PR but the confirmation dialog will appears after disconnecting from server in this case.

bear101:
On macOS there's a difference between close and quit. Close means to hide the window whereas quit means to exit the application.

bear101:
I've reverted the commit and reopened the issue.

bear101:
@CoBC I've move the confirmation dialog to slotClientExit(). Can you test if this works on Windows and Linux?

bear101:
@CoBC can you test if the latest commit works on Linux? I've tested on Windows and macOS

CoBC:
Can't test on Linux now, but I guess @jeremyp3 can help :)

jeremyp3:
After testing the PR, I can confirm that a warning is displayed (when the option is checked) for all ways of exiting TeamTalk on Linux:
* Alt + F4
* Ctrl + Q
* Client menu then Exit

So for me, everything is fine.

thanks !

PR #2799, opened 2025-08-06
Reintroduce call to EVP_CIPHER_CTX_init()

bear101:


hwangsihu:
There is no immediate problem, so go ahead and merge it. I heard that it is a deprecated function, so I think we need to find an alternative.

hwangsihu:
Please continue. I heard that it is a deprecated function, but let's wait and see for now.

hwangsihu:
@bear101, if possible, could you upgrade OpenSSL to 3.5.2 and merge it? It was upgraded to 3.5.2 two days ago.

bear101:
Hm, the functions are deprecated:
`[EVP_CIPHER_CTX_init](https://man.openbsd.org/EVP_CIPHER_CTX_init.3#EVP_CIPHER_CTX_init)() is a deprecated function that could be used to clear a cipher context on the stack before EVP_CIPHER_CTX was made opaque. Calling it on a cipher context just returned from [EVP_CIPHER_CTX_new(3)](https://man.openbsd.org/EVP_CIPHER_CTX_new.3) has no effect. Calling it on a cipher context that was already used may leak memory with older versions of the library. Instead, use [EVP_CIPHER_CTX_reset(3)](https://man.openbsd.org/EVP_CIPHER_CTX_reset.3) or [EVP_CIPHER_CTX_free(3)](https://man.openbsd.org/EVP_CIPHER_CTX_free.3).`

PR #2801, opened 2025-08-08
Minimum supported Android version is now 5.0 (Lollipop)

hwangsihu:


amirmahdifard:
Thanks! there are stil a lot of users that are using android 5 and 6 and likes to use teamtalk in there phone.

hwangsihu:
Yeah, there are still many people around me who use old Android devices, and I hear a lot of complaints. Last time, due to an error, I could only downgrade to 7.0, but not anymore!
Additionally, it has been confirmed that almost all functions are working normally.

amirmahdifard:
isn't it possible to downgrade to android 4? I have a device that has android 4 in it.

hwangsihu:
Unfortunately... This is not possible. Certain dependencies require Android 5.0 or higher.

amirmahdifard:
ok, sad

hwangsihu:
Thanks

beqabeqa473:
jetpack libraries are going to increase minimum sdk version to 23, so this is unnecessary change.

PR #2806, opened 2025-08-11
Optimizing workflows

hwangsihu:
Consolidated Ubuntu workflows from separate 22/24 versions into single ubuntu.yml. Streamlined workflow configurations by adding pull_request triggers, combining shell commands, removing redundant steps, and simplifying build processes across Android, iOS, macOS, and Windows workflows. Removed obsolete smoketest workflow to reduce CI complexity.


amirmahdifard:
take care about artifaqts of windows builds and stuf. I need it.

.github/workflows/macos.yml
```diff
@@ -29,12 +29,12 @@ jobs:
       with:
         version: '6.10.*'
         modules: 'qtmultimedia qtspeech'
-        cache-key-prefix: 'install-qt-action'
```
bear101, line 32: The Qt mirrors are very unstable so I think it would be a mistake to not use cached versions.

.github/workflows/android.yml
```diff
@@ -72,7 +66,7 @@ jobs:
         ln -s ${{runner.workspace}}/TeamTalk5/Library/TeamTalkJNI/libs/TeamTalk5.jar ${{runner.workspace}}/TeamTalk5/Client/TeamTalkAndroid/libs/TeamTalk5.jar
         ln -s ${{runner.workspace}}/TeamTalk5/Library/TeamTalkJNI/libs/TeamTalk5Test.jar ${{runner.workspace}}/TeamTalk5/Client/TeamTalkAndroid/libs/TeamTalk5Test.jar
         ln -s ${{runner.workspace}}/TeamTalk5/Library/TeamTalkJNI/libs/${{ matrix.arch }}/libTeamTalk5-jni.so ${{runner.workspace}}/TeamTalk5/Client/TeamTalkAndroid/src/main/jniLibs/${{ matrix.arch }}/libTeamTalk5-jni.so
-        ./gradlew build -x packageRelease
```
bear101: Why do you remove `-x packageRelease` ?

.github/workflows/windows.yml
```diff
@@ -12,32 +12,8 @@ env:
   SKIPKNOWNBUGS: 1
 
 jobs:
-  msvc:
```
bear101: I would rather keep the matrix because it should be extended to support Windows for ARM64

Build/Makefile
```diff
@@ -21,10 +21,18 @@ buildlib: genlib
 	cmake --build $(BUILDDIR) $(CMAKE_BUILD_EXTRA)
 
 android-build:
-	@echo "Testing if ANDROID_NDK_HOME environment variable is set: $(ANDROID_NDK_HOME)"
-	test -n "$(ANDROID_NDK_HOME)"
+	@if [ -n "$(ANDROID_NDK_LATEST_HOME)" ]; then \
```
bear101: No one knows if `ANDROID_NDK_LATEST_HOME` is able to build TeamTalk. This would lead to build errors that are not the fault of whoever made the merge-request.

bear101:
The `on: [push, pull_request]` builds pull-requests on the TeamTalk5 repo? It's a bit annoying that I cannot see if pull-requests are actually building correctly.

.github/workflows/macos.yml
hwangsihu: > The Qt mirrors are very unstable so I think it would be a mistake to not use cached versions.

https://github.com/jurplel/install-qt-action/blob/master/action.yml
This value is already defined, so there is no need to specify the default value again.

hwangsihu:
> The `on: [push, pull_request]` builds pull-requests on the TeamTalk5 repo? It's a bit annoying that I cannot see if pull-requests are actually building correctly.

Yes, it looks like the build is currently running in this PR and seems to be working properly.

bear101 (changes requested):
Please keep smoketest.yml. It builds much faster because it does not need to build dependencies. Then you get faster feedback.

Also keep ubuntu22.yml and ubuntu24.yml since they may diverge.

hwangsihu:
> Also keep ubuntu22.yml and ubuntu24.yml since they may diverge.

There are almost no other parts, so I added only the other parts with an if statement, but should I still separate them? If so, I'll gladly do so, but it will be inconvenient to manage.

Build/Makefile
```diff
@@ -21,10 +21,15 @@ buildlib: genlib
 	cmake --build $(BUILDDIR) $(CMAKE_BUILD_EXTRA)
 
 android-build:
-	@echo "Testing if ANDROID_NDK_HOME environment variable is set: $(ANDROID_NDK_HOME)"
-	test -n "$(ANDROID_NDK_HOME)"
+	if [ -n "$(ANDROID_NDK_HOME)" ]; then \
+		echo "Using ANDROID_NDK_HOME: $(ANDROID_NDK_HOME)"; \
+		NDK_HOME="$(ANDROID_NDK_HOME)"; \
```
bear101: Isn't it easier to just keep `ANDROID_NDK_HOME` instead of introducing `NDK_HOME`

bear101 (commented):
Then ok to merge ubuntu22.yml and ubuntu24.yml.

hwangsihu:
> Then ok to merge ubuntu22.yml and ubuntu24.yml.

Okay, I'll restore the Smoke Test workflow later.

Build/Makefile
hwangsihu: > Isn't it easier to just keep `ANDROID_NDK_HOME` instead of introducing `NDK_HOME`

Yes, I forgeten changing this.

PR #2817, opened 2025-08-16
Fix build errors

hwangsihu:
Only some compatibility issues have been resolved, and there is still a long way to go. It will probably fail the test.

### DIFF
--- a/.github/workflows/smoketest.yml
+++ b/.github/workflows/smoketest.yml
@@ -56,6 +56,7 @@ jobs:
            libtinyxml-dev
            libvpx-dev
            libxss-dev
+           nasm
            ninja-build
            portaudio19-dev
            python3-minimal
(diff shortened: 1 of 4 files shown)


bear101:
Should I close this?

hwangsihu:
No, let's merge it first and fix the remaining errors later.

hwangsihu:
Before that, we need to modify the title of #2477. This update includes changes to upgrade FFmpeg to version 8.0.

bear101:
But it does not work, so why merge it?

hwangsihu:
Because the build issue has been fixed. I think it's better to merge this PR now and resolve the issue where media streams aren't working later.

bear101:
I do not see any reason to merge something that does not work. I'd rather keep the compile error, since it gives a hint what the issue is.

PR #2825, opened 2025-08-22
perf(network): Optimize file transfer speed

BlindMaster24:
I've made a few tweaks to the networking code to speed things up, mostly around buffer sizes and enabling the `TCP_NODELAY` option.

Why these changes?
*   **Buffers:** More data moved at once, no pauses.
*   **`TCP_NODELAY`:** Data sent instantly, smooth flow.

The results are looking good:

*   My friend with a 300mb/s connection went from ~6 MB/s to ~16 MB/s on uploads.
*   Another friend on a 1gbit/s connection is now seeing speeds around 100 MB/s.

### DIFF
--- a/ChangeLog.txt
+++ b/ChangeLog.txt
@@ -4,6 +4,7 @@
 
 Version 5.20, unreleased
 Default Qt Client
+- Optimized file transfer performance by increasing buffer sizes and enabling TCP_NODELAY
 -
 Android Client
 - Minimum supported Android version is now 5.0 (Lollipop)
--- a/Library/TeamTalkLib/teamtalk/client/FileNode.cpp
+++ b/Library/TeamTalkLib/teamtalk/client/FileNode.cpp
@@ -29,7 +29,7 @@
 using namespace std;
 using namespace teamtalk;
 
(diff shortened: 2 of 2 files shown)


amirmahdifard:
@BlindMaster24 nice, but what happens for people that have slow internet like iran?
your changes points to teamtalk server, so fix the changelog, and add the entry under server, instead of default qt client.

hwangsihu:
This is not so much a server-side change as it is a change affecting all clients. It can be viewed as a change to the TeamTalk SDK, or the same log will need to be added to the Android client as well

BlindMaster24:
> This is not so much a server-side change as it is a change affecting all clients. It can be viewed as a change to the TeamTalk SDK, or the same log will need to be added to the Android client as well

Yes.

BlindMaster24:
Hey @bear101, can you tell me if my change is gonna be applied? All I got was a closed PR and no other answer.

bear101:
`TCP_NODELAY` is for terminals where data shouldn't be buffered. `TCP_NODELAY` does not enable faster transmission.

bear101:
Also your PR didn't even build.

PR #2830, opened 2025-08-25
Remove key from ini-file if same as default

bear101:


CoBC:
@bear101 So we need to replace all calls to ttSettings->setValue() by ttSettings->setValueOrClear()? Also, we probably have to make a loop to call this method for all settings keys at first startup with SETTINGS_VERSION=5.6?

bear101:
Yes, we need to replace all setValue() that have a default value. We cannot loop in the beginning because to clear old values because we don't have a list of key and default values. Or at least it will be difficult to make this list.

CoBC:
OK, I can make the replacement if you want

bear101:
I've just rebased and force pushed to fix a Qt 5.15 build issue. You're welcome to do the change. Can you think of any issues with this approach?

CoBC:
I don't think of any issue for now. I've started to make the change. Should I also call ttSettings->remove instead of setValue when we force default value (E.G. statusbardlg.cpp:l182)?

bear101:
For statusbardlg.cpp:182 I would call `remove()`

CoBC:
@bear101 Done! I've marked as ready for review

bear101:
Doh, I made a mistake in `setValueOrClear`. It should be:
```
    void setValueOrClear(const QString& key, const QVariant& newvalue, const QVariant& defvalue)
    {
        if (newvalue == defvalue)
        {
            remove(key);
        }
        else
        {
            setValue(key, newvalue);
        }
    }
```

bear101:
With the above correction then you didn't have to use "remove()" in statusbardlg.cpp.

CoBC:
I find it more logic to directly call remove() because we know that we force default value, so we can remove it clearly. For me, setValueOrClear() is when we didn't know if the new value is same as default or not.

PR #2835, opened 2025-08-29
Reintroduce option to speak tree item using TTS on MacOS

### LINKED ISSUE #2421: VoiceOver does not read the tree views properly in newer versions of TeamTalk
### Description

When I'm focused on a tree view, such as channel list or server list on teamTalk and try to read its content voiceOver doesn't give any feedback, sometimes it reads some items from the tree view, but not the ones that are focused.

### Application

- [X] qtTeamTalk
- [ ] TeamTalkAndroid
- [ ] iTeamTalk
- [ ] TeamTalkClassic
- [ ] TeamTalkServer

### Platform

- [ ] Windows
- [X] macOS
- [ ] Android
- [ ] iOS
- [ ] Linux

### Expected behavior

VoiceOver should read the focused item in the tree views when moving through them.

### Actual behavior

When moving through the tree view items, voiceOver doesn't speak anything or, sometimes, it speaks the previous focused items in the tree view, but not the actual focused ones.

Try to log in to a server with more than one channel or users logged on, focus the channel list tree view and move through each items in it. You should see how VoiceOver behaves.

CoBC:
Fixed #2421

### DIFF
--- a/ChangeLog.txt
+++ b/ChangeLog.txt
@@ -5,6 +5,7 @@
 Version 5.20, unreleased
 Default Qt Client
 - TeamTalk5.ini no longer stores values that have the same value as default
+- Reintroduce option to speak tree item using TTS on MacOS due to Qt issues
 - Fixed check state of server log events on macOS not being visible
 - Fixed global status message set to: "general_/statusmsg"
 Android Client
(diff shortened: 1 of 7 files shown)


bear101:
Merci!

PR #2857, opened 2025-09-04
Init/deinit methods for TeamTalkService in MainActivity

bear101:
Emphasize access to 'ttservice' to avoid null-pointer issues.


amirmahdifard:
@bear101 after merge of this pull request, please revert the block teamtalk service until teamtalk is connected pr that you did before, because doing that is not a good idea. Thanks.

bear101:
@amirmahdifard we're all looking forward to you finding a better solution. Until then: Shut up!

PR #2858, opened 2025-09-04
complete fix ui overlap for android 15

amirmahdifard:
instead of ap compat, this needed to be acsept normal activitys, so it can be used in all activitys, for example, settings activity.
this pull request fixes the issue completely forever.
tested on android 12 myself, android 15 my friend's phone, and android 11 my mother's phone.


amirmahdifard:
@hwangsihu I wanted to open this pull request an hour ago before beka's merged pr. Now, it's been aplyed in all activitys, and edg to edg helper uses normal activity instead of only app compat. @bear101 please merge this and then quickly push an update to google play as a hotfix, so people will stop complaning.

amirmahdifard:
@bear101 thanks! please push an update to google play, i'm tired of people complaning about teamtalk on android 15

hwangsihu:
No, to do that, TT 5.20 would need to be released on all platforms, but there's still an issue with Windows using OpenSSL 1.1 that needs to be fixed, and several other unresolved issues must also be addressed.

hwangsihu:
And repeatedly bothering developers with requests doesn't seem like a good idea either.

PR #2862, opened 2025-09-05
Refactor prefs handling to be easier

beqabeqa473:
Add PrefsHelper class to streamline access to preferences and easier usage
Fix microphone reset to 50% when restarting application
Fix setting input gain to 0 which slider is a 0%
fix typo in micLevel beeing mikeLevel.


bear101:
Thanks Beqa

PR #2871, opened 2025-09-11
Polish translation for qt client

zstanecic:
This is updated from a long time ago: This needs testing by users in beta. Translated with latest Poedit. New poedit 3.7 understands qt linguist .ts format.
Thanks to @vslavik


bear101:
Phew, must have taken a while :) Thanks a lot!

PR #2879, opened 2025-09-19
New translations for Brazilian Portuguese

DalisonCPU:
The missing translations for iPhone and Android have been added. I have also translated many lines of Teamtalk for Windows.
This work was very tiring, as the Teamtalk translation editor is no longer accessible, and, moreover, many translations, despite being in different places, are exactly the same. Shortcut keys were also being displayed as translations, and I believe this should be reviewed in the future.


bear101:
@DalisonCPU should I take a look at it?

DalisonCPU:
I made some mistakes that I need to correct, and I'll create another one when it's finished.

PR #2883, opened 2025-09-20
Install Cygwin instead of MSYS2

hwangsihu:
I also simplified the name of the Yasm installation section and fixed triplet issues.


hwangsihu:
The build is successful, but the unit tests are failing. However, it does not appear to be an issue caused by this PR, so please proceed as is.

bear101:
Thanks a lot @hwangsihu! It is the libvpx test case that is failing but at least it's able to run it now. I'll try and figure out why it isn't working.

PR #2885, opened 2025-09-21
Build Windows 32-bit

bear101:


bear101:
The 32-bit Windows only has one remain build error:

```
(Link target) -> 
         MSVCRT.lib(chandler4gs.obj) : error LNK2001: unresolved external symbol __except_handler4_common [D:\a\TeamTalk5\output\Library\TeamTalkLib\TeamTalk5DLL.vcxproj]
```

@hwangsihu do you have any idea why we get this error?

hwangsihu:
I'll look into it later.
By the way, it seems the log shows that the 32-bit OpenSSL couldn't be found.

bear101:
> I'll look into it later. By the way, it seems the log shows that the 32-bit OpenSSL couldn't be found.

I don't think we can use Qt 5.15 with VS2022 so another thing we need to do is to also disable the Qt build. Basically we need to skip `jurplel/install-qt-action@v4` for the 32-bit build matrix

hwangsihu:
If you're disabling Qt, why bother adding 32-bit?

bear101:
Some are still on 32-bit Windows and can build TeamTalk Classic.

kirill-jjj:
Yeah, a real smart idea. Every app, including Steam, and even NVDA, which has been 32-bit only since 2006, is going to be 64-bit in version 2026.1. Steam is going to be 64-bit only, I saw some news, but TeamTalk? You removed 64-bit in 5.11 or some version close to it, and now when every program is going to be 64-bit, this is the exact time for TeamTalk to re-add 32-bit! I can't even see what the problem is! Maybe you should... I don't know, fix bugs, maybe, fix accessibility which is partly broken, or I don't know, the desktop sharing crash! About which many people have told you, but no! We removed Windows 32-bit support from some old version? Let's re-add it! We removed Android 5 to 7 support in some old version? Let's re-add it! You want an argument for why TeamTalk is using old components? "Well, at least you can use it on Android 5 and Windows 10 32-bit in 2029! What other program can do that? And who cares if we are using really old components? It still works on 32-bit in 2029!" I'm just disappointed that instead of spending time and effort on real problems, you are doing a backward evolution and re-adding what you removed before. I know I'm a nobody, I'm not giving you money, I can't sit with your kids, etc., etc. So I'm just voicing my opinion.

BlindMaster24:
> Yeah, a real smart idea. Every app, including Steam, and even NVDA, which has been 32-bit only since 2006, is going to be 64-bit in version 2026.1. Steam is going to be 64-bit only, I saw some news, but TeamTalk? You removed 64-bit in 5.11 or some version close to it, and now when every program is going to be 64-bit, this is the exact time for TeamTalk to re-add 32-bit! I can't even see what the problem is! Maybe you should... I don't know, fix bugs, maybe, fix accessibility which is partly broken, or I don't know, the desktop sharing crash! About which many people have told you, but no! We removed Windows 32-bit support from some old version? Let's re-add it! We removed Android 5 to 7 support in some old version? Let's re-add it! You want an argument for why TeamTalk is using old components? "Well, at least you can use it on Android 5 and Windows 10 32-bit in 2029! What other program can do that? And who cares if we are using really old components? It still works on 32-bit in 2029!" I'm just disappointed that instead of spending time and effort on real problems, you are doing a backward evolution and re-adding what you removed before. I know I'm a nobody, I'm not giving you money, I can't sit with your kids, etc., etc. So I'm just voicing my opinion.

Totally agree. Tbh, I didn't expect anything less.
Time to send this thing to the internet graveyard, lol.

beqabeqa473:
@kirill-jjj @BlindMaster24 Toxic behavior doesn’t belong here.


@bear101 windows 10 is going to be discontinued soon. Personally i don't think adding 32 bit support back is an effort here. It would be better for all of us to invest to qt accessibility support. There should not be a way for going back.

hwangsihu:
> > I'll look into it later. By the way, it seems the log shows that the 32-bit OpenSSL couldn't be found.
> 
> 
> 
> I don't think we can use Qt 5.15 with VS2022 so another thing we need to do is to also disable the Qt build. Basically we need to skip `jurplel/install-qt-action@v4` for the 32-bit build matrix

It doesn't seem to be a Qt build issue.

hwangsihu:
The build appears to have succeeded. It seems we can now proceed with the merge.

hwangsihu (approved):

PR #2888, opened 2025-09-23
Force BearWare.dk WebLogin in TeamTalk Pro server

bear101:


amirmahdifard:
does this mean people using pro server are now forced to use a bearware.dk login account in there server and is it no longer possible to create it optionally throo the config?

amirmahdifard:
personally, I don't agree. In my opinion everything should be optional

bear101:
> does this mean people using pro server are now forced to use a bearware.dk login account in there server and is it no longer possible to create it optionally throo the config?

I've rephrased the changelog.

PR #2902, opened 2025-09-29
Support high quality Bluetooth recording

bear101:


hwangsihu:
Why did you downgrade the submodule?

bear101:
Well, spotted! I've fixed it. Windows workflow should also have spotted it if it wasn't because GtiHub Actions are failing :(

bear101:
@hwangsihu do you know if using GitHub Action's caching feature could solve the Qt download issue?
https://github.com/BearWare/TeamTalk5/actions/runs/18109682808

hwangsihu:
Qt 6.9.3 does not appear to be fully released yet.
For the time being, you will need to use 6.9.2 instead of 6.9.* or wait.

PR #2906, opened 2025-10-02
Join code for joining server with a "secret" key

bear101:
Ability to publish a server (in BearWare.dk's database) that will have a "secret" key (8 characters) that other can enter to join quickly (without using .tt files or entering IP-address etc.)


amirmahdifard:
I think there are some more important features that users are waiting fore. Such as duration ban in server, etc.

hwangsihu:
Limiting it to 8 characters doesn't seem like a good idea.

amirmahdifard:
@bear101 I want to add input and output devices selection to teamtalk android, but as beka said before, can you please expose the devices in teamtalk lib jni, so I can call and use in teamtalk android, or tell me if it's already available, or how I can list them in teamtalk android client? I want to add new features to teamtalk android, but some features are depend on you, and you are not very active. Can you please responceable for this? Thank you.

PR #2912, opened 2025-10-06
Fix Windows TeamTalk5Server script

hwangsihu:
* Fixed an issue where some characters would display abnormally on non-English computers
* Now after installing the server, it will ask you whether you want to start it
* When running the console version on a Standard server, it reads the configuration files from the current folder, just like the Pro server


bear101:
When tt5srv prints to console and read from console then the function `UnicodeToLocal()` and `LocalToUnicode()` are used: https://github.com/BearWare/TeamTalk5/blob/master/Library/TeamTalkLib/bin/ttsrv/ServerConfig.cpp#L233

I was hoping this was enough to handle the character conversions. It would be best if we can solve the issue in the executable (tt5srv.exe), so users do not have to do this.

hwangsihu:
The text is printed properly in the wizard, but it doesn't seem to work properly when running -verbose to display the server log.

PR #2916, opened 2025-10-08
Update zlib to 1.3.1

hwangsihu:


bear101:
Thanks for getting rid of the "homemade" zlib. Future updates will be much easier now!

PR #2924, opened 2025-10-11
Allow developers to have their own .gitattributes

bear101:
@hwangsihu now you can create your own .gitattributes with whatever line endings you want.


hwangsihu:
```
remote: Permission to BearWare/TeamTalk5.git denied to hwangsihu.
fatal: unable to access 'https://github.com/BearWare/TeamTalk5.git/': The requested URL returned error: 403
```

bear101:
isn't it because of the slash at the end?

hwangsihu:
No, Git always displays errors like that.

bear101:
At least now you can have your own settings for line endings. I don't want to ruin the Git history due to line endings.

hwangsihu:
I think I've found a way to avoid breaking the commit history. I'll resubmit the PR with all the changes later, so please merge this one first and then merge the additional PR. This is because the commit hash will change once the PR is merged.

However, while GitHub handles it well, Git requires you to manually enable it.

PR #2927, opened 2025-10-14
Force line endings to LF

hwangsihu:
After merging that PR, open the **.git-blame-ignore-revs** file and replace it with the new commit hash.


bear101:
This is an unnecessary complexity. The reason I made the PR #2924 was so you can make your own line endings depending on your platform.

The editors I use don't care about line endings, so I don't see this as a big issue.

hwangsihu:
Yes, it doesn't seem that critical due to the autocrlf setting. We'll address it if any issues arise later.

hwangsihu:
@bear101, I think I've discovered an issue where line endings can become jumbled even in the same file. It seems like ignoring git-blame and then renormalizing would be a better solution.

bear101:
Not sure I understand what you mean

PR #2932, opened 2025-10-19
Spambot account

bear101:


hwangsihu:
@bear101, How about embedding all SpamBot's features directly into the server?

bear101:
Having the spam features in the TeamTalk server would complicate it a lot. Spam is only an issue if you run a public server. If you don't then there's no need for all these features. The spambot downloads a bunch of listed VPN servers but the TeamTalk server should not do this. It is not relevant for the server to know all this. It's only purpose is to serve the users the simplest way.

hwangsihu:
If that's the problem, why not make the spam bot itself a standalone program? If the server where the spam bot is installed goes down, it'll likely take several days for it to be reactivated, which is problematic.

hwangsihu:
And if you don't accept my suggestion, it would be nice to have a public server and control the server name by prefixing it with something like **#addspambot#** and **#removespambot#**.

bear101:
You are welcome to change the jSpamBot to run as stand alone. I think you can already do this using the existing Java properties.

amirmahdifard:
yes. I agree as well for it to be standalone program

hwangsihu:
By the way, @bear101, how have you been running SpamBot so far?

bear101:
I run it something like this:

```
java -Ddk.bearware.iplogincount=6 \
     -Ddk.bearware.ipjoinscount=6 \
     -Ddk.bearware.ipkickscount=3 \
     -Ddk.bearware.ipcmdduration=60 \
     -Ddk.bearware.ipv4banprefix=24 \
     -Ddk.bearware.ipv6banprefix=96 \
     -Ddk.bearware.regname="Your TeamTalk SDK regname" \
     -Ddk.bearware.regkey="Your TeamTalk SDK regkey" \
     -Ddk.bearware.abuseipdbkey="Your AbuseDB Key" \
     -Ddk.bearware.abuseipdbtotalreports=2 \
     -Ddk.bearware.abuseipdbdistinctusers=2 \
     -Ddk.bearware.abuseipdbconfidencescore=2 \
     -Ddk.bearware.bandurationsecs=36000 \
     -cp TeamTalk5.jar:jTeamTalkSpamBot.jar:org-json.jar \
     Main
```

hwangsihu:
The problem is vpnips.txt. Where did you get it? I might just be missing it. And it looks like this program also requires a TeamTalkSDK license.

bear101:
Ah, forgot the following files:
```
wget https://raw.githubusercontent.com/X4BNet/lists_vpn/main/output/vpn/ipv4.txt -O vpnips.txt
wget https://raw.githubusercontent.com/BearWare/TeamTalk5/master/Client/jSpamBot/src/badwords.txt -O badwords.txt
wget https://raw.githubusercontent.com/BearWare/TeamTalk5/master/Client/jSpamBot/src/badwords_french.txt -O badwords_french.txt
```
You can run the SDK in trial mode.

bear101:
@hwangsihu don't make changes to other people's branches. When you force-push you basically erase the branch and start a new one.

hwangsihu:
I see, I'm sorry.

hwangsihu:
> Having the spam features in the TeamTalk server would complicate it a lot. Spam is only an issue if you run a public server. If you don't then there's no need for all these features. The spambot downloads a bunch of listed VPN servers but the TeamTalk server should not do this. It is not relevant for the server to know all this. It's only purpose is to serve the users the simplest way.

So, why not document the file download process in a README or similar location and then embed it? That would make implementation much easier and prevent it from appearing in the online users list (Actually it doesn't really matter). And you could optionally disable the VPN blocking feature.

hwangsihu:
And most importantly, since TeamTalk Pro server requires a web login, it will also prevent unauthorized use. Using a separate client would complicate things, as it would (probably) require server-side intervention as well.

hwangsihu:
@bear101 Hmm, I don't know why, but it's not easy. Instead of including SpamBot directly in the server executable, what do you think about giving you direct control over all the features jSpamBot supports, like automatically downloading vpnips and badwords files, or updating them every few minutes?

hwangsihu:
That way, we could also create a feature to hide them from the online user list.

PR #2933, opened 2025-10-19
Refactor wizard

bear101:


amirmahdifard:
it would be a good idea to be able to create channels in wizard as well.

PR #2946, opened 2025-10-22
Fix UTF-8 Encoding Errors on TeamTalkClassic

hwangsihu:


bear101 (changes requested):
This is used by C[TeamTalkDlg::OnCopyData](https://github.com/BearWare/TeamTalk5/blob/utf-8/Client/TeamTalkClassic/TeamTalkDlg.cpp#L4955)

hwangsihu:
Is there any way to replace this?

amirmahdifard:
@hwangsihu the tt5srv.log file doesn't support utf 8 as well. if system locall is set to english, and in the teamtalk server, people send persion or arabik messages to each other for example, it replaces by question marks. Please fix this as wel. I've reported this before, but anyone didn't pay any attention.

hwangsihu:
> This is used by CTeamTalkDlg::OnCopyData

@bear101, do you have any good ideas on how to fix this? The build seems to work fine in GitHub Actions, but I'm having trouble on my computer. Could it be because my computer's language is Korean?

bear101:
@hwangsihu In `Library/TeamTalkLib/teamtalk/Log.h` try adding `UnicodeToUtf8(TXT).c_str()` on Windows.

```
#include "myace/MyACE.h"
#if defined(UNICODE)
#define TT_LOG(TXT)  ACE_DEBUG((LM_INFO, ACE_TEXT("%D %s\n"), UnicodeToUtf8(TXT).c_str()))
#else
#define TT_LOG(TXT)  ACE_DEBUG((LM_INFO, ACE_TEXT("%D %s\n"), TXT))
#endif
```

amirmahdifard:
@hwangsihu i'm sorry, did you see my comment? can the utf issue be fixed in windows server log file as well, So we don't have to change system local in auder to support persian and arabik langs when people send messages like this?

hwangsihu:
Hmm... it doesn't work. Instead, I've encountered an absurd situation where English characters are broken.
For now, I think it would be better to create a separate PR to address this issue later.

hwangsihu:
@bear101 Sorry to bother you, but when can you approve my review? I converted the same symbols to UTF-8, so it should work fine.

Client/TeamTalkClassic/TeamTalkApp.cpp
```diff
@@ -186,7 +186,7 @@ BOOL CTeamTalkApp::InitInstance()
         {
             szArg = info.m_args.GetNext(pos);
             _tcsncat(msg.szPath, szArg.GetBuffer(), MAX_PATH);
-            _tcsncat(msg.szPath, _T("�"), MAX_PATH);
+            _tcsncat(msg.szPath, _T("¤"), MAX_PATH);
```
bear101, line 189: Is this the same character as [here](https://github.com/BearWare/TeamTalk5/blob/master/Client/TeamTalkClassic/TeamTalkDlg.cpp#L4931)?

hwangsihu:
Yes, just the encoding is different.

bear101:
> Yes, just the encoding is different.

So you have tested that the tokenize still works when you drag a .tt file onto the main window?

hwangsihu:
Are you talking about the Connect to Server dialog?
I've seen it work fine there.

bear101:
You drag and drop a .tt-file from Windows Explorer onto TeamTalk's main window using the mouse.

hwangsihu:
> You drag and drop a .tt-file from Windows Explorer onto TeamTalk's main window using the mouse.

@bear101, It seems to be working fine. Can you please check?

Client/TeamTalkClassic/TeamTalkApp.cpp
bear101: AFAIK you need to change this line as well. I cannot test it because I no longer has an installer for Classic that associates .tt files.

hwangsihu: That file is fine. Only the file I modified was corrupted due to an encoding issue.

PR #2948, opened 2025-10-22
Fix Windows TeamTalk5Server script

hwangsihu:
* Now after installing the server, it will ask you whether you want to start it
* When running the console version on a Standard server, it reads the configuration files from the current folder, just like the Pro server

### DIFF
--- a/Setup/Server/Windows/tt5srv_console.bat
+++ b/Setup/Server/Windows/tt5srv_console.bat
@@ -1,10 +1,11 @@
 @echo off
-@set TTSRVCFG=%APPDATA%\BearWare.dk\tt5srv.xml
+set TTSRVCFG=%APPDATA%\BearWare.dk\tt5srv.xml
 echo   --------------------------------------------------------------------------
 echo                       TeamTalk 5 Console Server
 echo   --------------------------------------------------------------------------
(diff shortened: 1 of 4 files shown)


Setup/Server/Windows/tt5srv_console.bat
```diff
@@ -1,10 +1,8 @@
 @echo off
-@set TTSRVCFG=%APPDATA%\BearWare.dk\tt5srv.xml
 echo   --------------------------------------------------------------------------
 echo                       TeamTalk 5 Console Server
 echo   --------------------------------------------------------------------------
 echo.
-@echo on
-tt5srv.exe -wizard -c %TTSRVCFG%
-tt5srv.exe -nd -verbose -c %TTSRVCFG%
+tt5srv.exe -wizard
```
bear101: tt5svc.exe runs as Adminstrator and can write to C:\Program Files.
tt5srv.exe runs as %USERNAME% and cannot write to C:\Program Files.

bear101 (commented):
What should tt5srv.exe require administrator access? It is not needed and maybe the user does not even have administrator access.

bear101 (approved):

PR #2953, opened 2025-10-25
Update .gitignore files

hwangsihu:
Based on Official templets


bear101:
Some of these .gitignore files seem to include a bunch of stuff we don't use. E.g. the python .gitignore mentions a lot of things we are not using.

PR #2954, opened 2025-10-26
Fix Proximity Sensor bug

### LINKED ISSUE #2949: Proximity sensor is activated even if option is disabled
### Description

Proximity sensor is actually turns on when logining to the server, even if "Proximity sensor" setting is disabled.
This has two side-effects:
1. Potential battery drain.
2. Upper speaker turns off on phones with virtual proximity sensor, like poco f6.

@beqabeqa473 

### Application

- [ ] qtTeamTalk
- [x] TeamTalkAndroid
- [ ] iTeamTalk
- [ ] TeamTalkClassic
- [ ] TeamTalkServer

### Platform

- [ ] Windows
- [ ] macOS
- [x] Android
- [ ] iOS
- [ ] Linux

### Expected behavior

Proximity sensor should not be used when appropriate checkbox unticked in the settings.

### Actual behavior

Proximity sensor probably always used.

1. Use phone with virtual proximity sensor.
2. Disable proximity sensor setting in the teamtalk settings.
3. Login to the teamtalk server.

hwangsihu:
Closes #2949

### DIFF
--- a/Client/TeamTalkAndroid/src/main/java/dk/bearware/gui/MainActivity.java
+++ b/Client/TeamTalkAndroid/src/main/java/dk/bearware/gui/MainActivity.java
@@ -203,6 +203,7 @@ extends AppCompatActivity
     boolean restarting;
     SensorManager mSensorManager;
     Sensor mSensor;
+    boolean isProximitySensorRegistered = false;
     Map<Integer, User> users = new HashMap<>();
 
     static final String MESSAGE_NOTIFICATION_TAG = "incoming_message";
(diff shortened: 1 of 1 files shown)


beqabeqa473:
Please make variable cammelcased. snake-case is not allowed by java development convention.
Also check for unregister also, unregister if it is registered.

beqabeqa473:
what is the purpose of unregisterring it when activity is beeing paused?
In this case proximity sensor will not work when app is in background and/or screen is off.

beqabeqa473:
Seems to be ok.

beqabeqa473 (approved):

PR #2957, opened 2025-10-26
Fix dir of OpenSSL 3.x dll and Do not find 1.1.x

hwangsihu:


bear101:
It looks like the Qt action can install OpenSSL:
https://github.com/jurplel/install-qt-action?tab=readme-ov-file#tools
https://aqtinstall.readthedocs.io/en/latest/getting_started.html#installing-tools

```
    - uses: jurplel/install-qt-action@v4
      with:
        arch: '${{ matrix.qtarch }}'
        version: '${{ matrix.qtversion }}'
        modules: '${{ matrix.qtmodules }}'
        setup-python: false
        tools: 'tools_openssl_x86,tools_openssl_x64'
```

hwangsihu:
Good idea, I'll fix later.

hwangsihu:
It looks like there is no 32-bit version of OpenSSL.

Client/qtTeamTalk/CMakeLists.txt
```diff
@@ -208,14 +208,14 @@ endif()
           # Qt Maintenance Tool is able to download and install Qt's OpenSSL binaries. These files
           # will be located in <Qt install dir>/Tools/OpenSSL
           if (${CMAKE_SIZEOF_VOID_P} EQUAL 8)
-            find_file (QT_OPENSSL_LIBCRYPTODLL_PATH NAMES libcrypto-3-x64.dll libcrypto-1_1-x64.dll)
-            find_file (QT_OPENSSL_LIBSSLDLL_PATH NAMES libssl-3-x64.dll libssl-1_1-x64.dll)
+            find_file (QT_OPENSSL_LIBCRYPTODLL_PATH NAMES libcrypto-3-x64.dll PATHS "C:/Program Files/OpenSSL/bin")
```
bear101, line 211: If you add: ` tools: 'tools_openssl_x64'` to the qt step then this path shouldn't be necessary.

hwangsihu: Qt's OpenSSL is outdated but also suffers from version mismatches with 32-bit systems. Wouldn't it be better to just proceed as is?

bear101 (approved):

PR #2959, opened 2025-10-26
Update PortAudio to latest commit on Linux

hwangsihu:


hwangsihu:
@bear101, do you have any ideas on how to make the features you added to the PortAudio fork available in the latest commit? I've tried several times, but it seems like the latest commit doesn't support these features.

bear101:
It's a long process to merge my changes to PortAudio every time. PortAudio does not support unicode therefore I added it myself.

bear101:
Also I added sound device identification to PortAudio, so it's possible to find the last used sound device.

bear101:
Actually unicode issue is fixed in PortAudio. They put sound device name as UTF-8.

PR #2962, opened 2025-10-29
Update opus-tools to 0.2

hwangsihu:


bear101:
Do we need opustools after this change? It seems the only function we use from opustools is the one you added in the anonymous namespace.

hwangsihu:
> Do we need opustools after this change? It seems the only function we use from opustools is the one you added in the anonymous namespace.

I've only added functions that were removed in OpusTools 0.2. It appears that some features still depend on OpusTools.

PR #2974, opened 2025-11-02
Fix jSpamBot Gradle build issue

hwangsihu:


Client/jSpamBot/build.gradle
```diff
@@ -7,8 +7,8 @@ repositories {
 }
 
 compileJava {
-    sourceCompatibility = '11'
-    targetCompatibility = '11'
+    sourceCompatibility = '21'
+    targetCompatibility = '21'
```
bear101: We should not force users to update their Java installation if it's not needed. Java 11 will work just fine as targetCompatibility .

bear101 (approved):

PR #2975, opened 2025-11-02
Optimize workflow steps

hwangsihu:
* Updated to Python 3.14
* Removed steps related to apt autoremove
* No longer reinstall OpenJDK 17 and already installed packages (Only Android and Smoke Test) with apt on Android

### DIFF
--- a/.github/workflows/windows.yml
+++ b/.github/workflows/windows.yml
@@ -12,38 +12,38 @@ env:
   SKIPKNOWNBUGS: 1
 
 jobs:
-  msvc:
+  build:
     if: |
       (github.event_name == 'push' && (!startsWith(github.ref, 'refs/heads/debug') || startsWith(github.ref, 'refs/heads/debug-windows'))) ||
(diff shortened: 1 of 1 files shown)


.github/workflows/android.yml
```diff
@@ -46,8 +46,7 @@ jobs:
       run: printenv | sort
 
     - name: Install dependencies
-      working-directory: ${{runner.workspace}}/TeamTalk5/Build
-      run: sudo make depend-ubuntu24-android
+      run: sudo apt-get install cmake doxygen junit4 ninja-build
```
bear101: We don't know if the dependency list in Build/Makefile is correct if we don't test it.

bear101 (changes requested):

hwangsihu:
By the way, are Ubuntu 18 and Ubuntu 20 still usable? Can they be removed?

bear101:
I guess they can be removed

PR #2981, opened 2025-11-06
Android client. Fixed bug where join and leave events were not announced when focused on another channel

amirmahdifard:
there was a bad bug that when you were focused on other channels, the join and leave events wouldn't be announced to you in android client. It has now finally been fixed.


amirmahdifard:
I'm using github desktop, So I don't think I have that option. Can you please Squash?

PR #2989, opened 2025-11-11
Android: Change client name display format

hwangsihu:
Changed from "**<ClientName> version <Version>**" to "**<ClientName> <Version>**".


bear101:
@amirmahdifard you originally made this line. Do you think this change is necessary?

amirmahdifard:
@bear101 yes, I agree with this change.
but @hwangsihu please also remove this line from values\strings.xml in this pr before it's being merged
    <string name="user_prop_clientversion">version</string>
    Otherwise it looks fine to me. Thanks

bear101 (approved):

PR #2991, opened 2025-11-12
Replace DataToObject() by union in TTMessage

bear101:


bear101:
In TTMessage struct .NET interop reports "contains an object field at offset 4 that is incorrectly aligned or overlapped by a non-object field" on string field due to union.
https://www.developerfusion.com/article/84519/mastering-structs-in-c/

PR #2947, opened 2025-10-22
Optimize WebRTC build: Other Platforms

hwangsihu:
Build successes:

- [x] Windows
- [x] MacOS
- [x] Android
- [x] iOS
- [x] Ubuntu 22
- [x] Ubuntu 24

PR #2995, opened 2025-11-13
Upgrade to VS2026

hwangsihu:


amirmahdifard:
please fix a bug also. github artifaqts builds, which I really use, doesn't support web rtc when building qt teamtalk

hwangsihu:
> please fix a bug also. github artifaqts builds, which I really use, doesn't support web rtc when building qt teamtalk

We can't do that due to disk space issue.

hwangsihu:
But I'm wonder how much it would cost to upgrade to an SSD.

hwangsihu:
And importantly, there is currently an issue with WebRTC auto-checkout, so #2947 needs to be merged before work can continue.

hwangsihu:
By the way, @bear101, could you please enable the **Always suggest updating pull request branches** option in repository settings? Typing `git rebase origin/master` is a bit of a hassle.

PR #2999, opened 2025-11-16
Formatting all Python scripts using Black

hwangsihu:
Line length is 120.


hwangsihu:
I think it could be automated with pre-commit, but it would be better to work on that in a separate PR.

bear101:
I think it would be better to put it in as a step in a workflow that checks formatting.

PR #3005, opened 2025-11-21
Update catch2 to 3.11.0

hwangsihu:


Library/TeamTalkLib/test/catchqt/TeamTalkCatchQT/main.cpp
```diff
@@ -1,5 +1,5 @@
 #define CATCH_CONFIG_RUNNER //E.g. own main is supplied. In contrasts to CATCH_CONFIG_MAIN
-#include <catch.hpp>
+#include <catch_amalgamated.hpp>
```
bear101: In Catch2 documentation they include "catch2/catch_test_macros.hpp". Why are you including this one?

The documentation states:
```
 * Generally the Catch2 users should use specific includes they need,
 * but this header can be used instead for ease-of-experimentation, or
 * just plain convenience, at the cost of (significantly) increased
 * compilation times.
```

Library/TeamTalkLib/test/CatchWin.cpp
```diff
@@ -1,402 +1,402 @@
-#include "catch2/catch.hpp"
```
bear101, line 1: Somethings wrong with line endings here.

Library/TeamTalkLib/test/catchqt/TeamTalkCatchQT/main.cpp
hwangsihu: Since this can't be done by simply changing the header, wouldn't it be better to create a separate PR later?

bear101:
Catch2 actually supports CMake so we should build Catch2 as part of the toolchain (same way as OPUS, Ogg, etc). Here's an example: #3042

PR #3008, opened 2025-11-22
Disable fail-fast in matrix strategies

hwangsihu:


bear101:
Isn't it fine to cancel builds if there's an error. It will free up the build queue, so other builds can run.

hwangsihu:
I don't think so. Let me give you an example.

When updates to Qt for Windows and macOS are released, the workflow temporarily fails.
However, the Windows 32-bit build uses Qt5, which is no longer released, so it doesn't affect it.
Should the 32-bit build also be canceled because of that?
Of course, there are ways to lock the version instead of 6.10.*, but let's think about that later.

PR #3012, opened 2025-11-22
Skip reinstalling existing packages in Smoke Test

hwangsihu:


hwangsihu:
However, there's a minor issue: the logs get a bit long. If a problem occurs, you can just reinstall them, and removing existing packages from the list will make maintenance a bit easier.

bear101:
I don't see how installing libssl-dev and zlib1g-dev would make the logs bigger.

hwangsihu:
So what about the second thing I mentioned, easier maintenance?

bear101:
Maintenance is easier if you state your requirements...

PR #3013, opened 2025-11-22
Restructure Windows workflow and optimize builds

hwangsihu:
- Rename job from 'msvc' to 'build'
- Simplify matrix configuration with arch as primary dimension
- Change runs-on from matrix variable to windows-latest
- Add opensslplatform and vcvars to matrix for better OpenSSL handling
- Update matrix variable references from dotnetarch to arch
- Move MSYS2 setup earlier in workflow
- Remove setup-MSBuild action (not needed)
- Enable parallel builds using all CPU cores (--parallel instead of --parallel 8)
- Add continue-on-error to CHM compilation step
- Update conditional checks to use matrix.arch instead of matrix.name

### DIFF
--- a/.github/workflows/windows.yml
+++ b/.github/workflows/windows.yml
@@ -22,34 +22,34 @@ env:
   SKIPKNOWNBUGS: 1
 
 jobs:
-  msvc:
+  build:
     strategy:
       matrix:
+        arch: [x86, x64]
         include:
-          - name: VS2019Release32
-            vmimage: windows-2025
+          - arch: x86
             cmakeplatform: Win32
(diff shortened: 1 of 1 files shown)


.github/workflows/windows.yml
```diff
@@ -22,34 +22,37 @@ env:
   SKIPKNOWNBUGS: 1
 
 jobs:
-  msvc:
+  build:
+    if: |
```
bear101: Please move the debug branch name to a separate pull request

bear101 (changes requested):

bear101 (approved):

PR #3015, opened 2025-11-22
Optimize Docker build files

hwangsihu:


bear101:
Didn't you change the docker files to use Build/Makefile depend-* targets in a previous PR? Then we wouldn't need duplicate code.

hwangsihu:
I think it's going to be difficult because of a problem with volume mounting. I think adding --mount to the Dockerfile might cause a conflict.

bear101 (commented):
Is the only change that spaces are replaced by tabs? I don't see why we need this change. Also why is "apt-get" preferred over "apt"?

hwangsihu:
I think I saw some tabs and spaces mixed up, so I changed it, but I'll check again.

And apt-get is great for scripting. In fact, if you run depend-* in make, you'll get a warning telling you to use apt-get instead.

bear101:
Please don't change the tab/space issue. It will ruin "git blame". It's more important that "git blame" is correct than tabs vs spaces.

hwangsihu:
Consider creating a **.git-blame-ignore-revs** file, as I suggested previously when normalizing all line endings.

bear101 (changes requested):
Please don't change the tab/space issue. It will ruin "git blame". It's more important that "git blame" is correct than tabs vs spaces.

PR #3017, opened 2025-11-22
Fix iOS build issues

hwangsihu:


hwangsihu:
Oh, I forgot to mark it as a draft. I'm trying to fix an issue where the build/ios-install path is incorrectly pointed to. Could this be my problem?

bear101:
You can just delete the line. It's not used for anything

PR #3026, opened 2025-11-29
Lint fix in TeamTalkAndroid

hwangsihu:
Some APIs that are only compatible with the latest Android have been replaced with older APIs so that they can be used stably on older Androids as well.

### DIFF
--- a/Client/TeamTalkAndroid/src/main/java/dk/bearware/data/AppInfo.java
+++ b/Client/TeamTalkAndroid/src/main/java/dk/bearware/data/AppInfo.java
@@ -29,6 +29,7 @@ import android.util.Log;
 
 import java.io.UnsupportedEncodingException;
 import java.net.URLEncoder;
+import java.nio.charset.StandardCharsets;
 
 import dk.bearware.TeamTalkBase;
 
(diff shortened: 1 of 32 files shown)


Client/TeamTalkAndroid/src/main/java/dk/bearware/data/AppInfo.java
```diff
@@ -77,26 +78,24 @@ public static String getUpdateURL(Context context) {
...
     public static String getBearWareTokenUrl(Context context, String username, String password) {
-
         try {
-            username = URLEncoder.encode(username, "utf-8");
-            password = URLEncoder.encode(password, "utf-8");
+            username = URLEncoder.encode(username, "UTF-8");
+            password = URLEncoder.encode(password, "UTF-8");
         } catch (UnsupportedEncodingException e) {
-            Log.e(TAG, "Unable to encode username/password: " + e);
+            throw new RuntimeException(e);
```
bear101: Why do we want to crash the application?

bear101 (changes requested):

Client/TeamTalkAndroid/src/main/java/dk/bearware/data/AppInfo.java
hwangsihu: It seems like lint will issue a warning if I don't add a catch.

hwangsihu: Otherwise, an error will occur.

bear101: Crashing the application is more user-friendly?

Client/TeamTalkAndroid/src/main/res/values-pt/strings.xml
```diff
@@ -1,343 +1,343 @@
-<?xml version="1.0" encoding="utf-8"?>
```
bear101, line 1: Something is wrong with line endings. It's not possible to see what has changed.

Client/TeamTalkAndroid/src/main/res/values-fa-rIR/strings.xml
```diff
@@ -1,495 +1,495 @@
-<?xml version="1.0" encoding="utf-8"?>
```
bear101, line 1: Something is wrong with line endings. It's not possible to see what has changed.

Client/TeamTalkAndroid/src/main/res/values-pt/strings.xml
hwangsihu: This file originally seemed to have mixed line endings. This issue cannot be fixed, as each time the file is saved, the modified line endings are committed.

bear101 (approved):

PR #3030, opened 2025-12-05
Fix build errors in Android

hwangsihu:


poretsky:
I'd better suggest to declare this value somewhere in resources and then
use it here as well as in
Client/TeamTalkAndroid/src/main/java/dk/bearware/gui/ServerEntryActivity.java.

Client/TeamTalkAndroid/src/main/res/values/strings.xml
```diff
@@ -320,6 +320,7 @@
     <string name="pref_title_channel">Channel</string>
     <string name="pref_title_channel_password">Channel Password</string>
     <string name="pref_title_last_channel">Last Joined Channel</string>
+    <string name="default_port">10333</string>
```
bear101: Why should people translate this?

poretsky:
This string should be marked as not translatable. It could be done by
attribute translatable="false". Or, as an alternative, such strings are
defined in donottranslate.xml file.

hwangsihu:
That's weird, why can't I merge them?

bear101 (approved):

PR #3032, opened 2025-12-05
Enable NASM on OpenSSL build

hwangsihu:


Library/TeamTalkLib/README.md
```diff
@@ -55,6 +55,9 @@ manually.
     * Place `yasm.exe` for x64 in `C:/tt5dist/yasm/x64`
     * Place `yasm.exe` for Win32 in `C:/tt5dist/yasm/win32`
     * LibVPX requires *yasm*
+* Download [NASM](https://www.nasm.us/)
```
bear101, line 58: Can't we download `nasm` and `yasm` as external projects (similar to _depot_tools_), so we don't need these manual steps?

hwangsihu:
This is a bad idea.
It would be better to switch to MSYS2's Mingw-w64 environment and manage dependencies with pacman like other operating systems.

hwangsihu:
I checked and it seems that enabling FFmpeg on Windows will solve the codec issue, although you'll have to give up TeamTalkClassic and several codecs (e.g., MediaFoundation). But I'm not sure about TeamTalk .Net.

PR #3033, opened 2025-12-05
Optimize workflow steps

hwangsihu:
- Fixed an issue where both push and pull_request triggers would work on the same commit
- Fixed debug branch filtering logic
- Replaced ${{runner.workspace}}/TeamTalk5} with ${{github.workspace}}
- Increased the execution cycle of the Update Gradle Wrapper to one week and disabled fork guard
- Change CMD to PowerShell in Windows CMake step
- Upgrade smoke test runner to Ubuntu 24.04
- Optimized install dependencies step in SmokeTest
- Normalize and combine steps

### DIFF
--- a/.github/workflows/android.yml
+++ b/.github/workflows/android.yml
@@ -2,18 +2,13 @@ name: Android
 
 on:
   push:
-    branches:
-      - '**'
-      - '!debug*'
-      - 'debug-android*'
+    branches: [ "master" ]
   pull_request:
-    branches:
-      - '**'
-      - '!debug*'
-      - 'debug-android*'
+    branches: [ "master" ]
 
 jobs:
   build:
+    if: "!startsWith(github.head_ref, 'debug') || startsWith(github.head_ref, 'debug-android')"
     strategy:
       matrix:
         include:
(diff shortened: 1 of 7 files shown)


.github/workflows/android.yml
```diff
@@ -52,9 +52,6 @@ jobs:
     - name: Update Ubuntu repository
       run: sudo apt-get update
 
-    - name: Print env
```
bear101, line 55: I use it when build fails

.github/workflows/smoketest.yml
```diff
@@ -18,7 +18,7 @@ env:
 
 jobs:
   build:
-    runs-on: ubuntu-22.04
+    runs-on: ubuntu-latest
```
bear101: Can you specify ubuntu-24.04 instead. Builds will fail when ubuntu-latest becomes Ubuntu 26.

hwangsihu:
Debug filtering doesn't seem to be working properly. Could you please merge? I think I fixed it correctly.

.github/workflows/android.yml
```diff
@@ -3,17 +3,14 @@ name: Android
...
   pull_request:
     branches:
-      - '**'
-      - '!debug*'
-      - 'debug-android*'
+      - master
 
 jobs:
   build:
+    if: "!startsWith(github.head_ref, 'debug-android')"
```
bear101: This looks incorrect. You build if it does _not_ start with `debug-android` but what we want is to not build windows, ios, etc. workflows (and only android workflow).

hwangsihu:
Is that okay now?
Sorry, I was stupid.

.github/workflows/smoketest.yml
```diff
@@ -2,11 +2,11 @@ name: SmokeTest
 
 on:
   push:
-    branches-ignore:
-      - 'debug*'
+    branches:
```
bear101: Will smoke test now only run on _master_ branch?

hwangsihu:
This creates a risk of conflicts every time a PR is approved. Is there any other way?

bear101:
You can make your PRs depend on commits from other PRs. A small PR is more likely to be merged than a big one.

PR #3035, opened 2025-12-06
Add /LTCG option

hwangsihu:


Library/TeamTalkLib/CMakeLists.txt
```diff
@@ -39,6 +39,9 @@ if (MSVC)
     endif()
   endforeach(flag_var)
 
+  set (CMAKE_EXE_LINKER_FLAGS "${CMAKE_EXE_LINKER_FLAGS} /LTCG")
+  set (CMAKE_SHARED_LINKER_FLAGS "${CMAKE_SHARED_LINKER_FLAGS} /LTCG")
```
bear101: This only applies to optimized/release builds, right?

hwangsihu: I haven't tried a debug build, but maybe...

bear101: Please verify debugger works as before in VS

bear101:
What exactly is the benefit of this change? When I look at the size of TeamTalk5.dll on _main_ branch and _ltcg_ branch then it's exactly the same.

hwangsihu:
> What exactly is the benefit of this change? When I look at the size of TeamTalk5.dll on _main_ branch and _ltcg_ branch then it's exactly the same.

Sorry, I just checked.
When compiling most files, you'll see a warning message saying that using the /ltcg option will improve linker performance.
This is in response to the warning.

bear101:
Looking at the build times I don't see any improvements in build time compared to master branch

hwangsihu:
But wouldn't it be nice if we could get rid of the warning message?

bear101:
By default VS2022 sets _Link Time Code Generation_ to nothing (blank) for _Debug_ and _/LTCG:incremental_ for _Release_ builds on DLL projects.

hwangsihu:
@bear101 Ah, Did you mean to change /LTCG to /LTCG:incremental?

hwangsihu:
Incremental LTCG is not working.

PR #3037, opened 2025-12-06
Fix TeamTalk5 Setup script

hwangsihu:
- Change the compression method to lzma2/ultra64
- Do not install if Microsoft Visual C++ v14 Redistributable Package is already installed
- Added missing official translations
- Refactor file paths into a variable-based structure to make path changes easier
- Add steps to create and upload a TeamTalk Installer in GitHub Actions


Setup/Installer/Windows/TeamTalk5.iss
```diff
@@ -58,43 +78,55 @@ BeveledLabel={cm:LanguageName}
...
+ko.AdditionalSoftware=추가 소프트웨어:
 zh_CN.Client=TeamTalk 5 客户端
 zh_CN.Full=TeamTalk 5 客户端 & 服务器
+zh_CN.InstallVCRedist=安装 Visual C++ 可再发行组件包 (推荐)
+zh_CN.AdditionalSoftware=附加软件:
 
 [Tasks]
 Name: "desktopicon"; Description: "{cm:CreateDesktopIcon}"; GroupDescription: "{cm:AdditionalIcons}"; Flags: unchecked
 Name: "quicklaunchicon"; Description: "{cm:CreateQuickLaunchIcon}"; GroupDescription: "{cm:AdditionalIcons}"; Flags: unchecked
+Name: "vcredist"; Description: "{cm:InstallVCRedist}"; GroupDescription: "{cm:AdditionalSoftware}"; Components: client
```
bear101: How would people know if they need it?

Setup/Installer/Windows/TeamTalk5.iss
```diff
@@ -166,3 +198,45 @@ Type: files; Name: "{app}\Languages\Spanish.qm"
 Type: files; Name: "{app}\Languages\Thai.qm"
 Type: files; Name: "{app}\Languages\Turkish.qm"
 Type: files; Name: "{app}\Languages\Vietnamese.qm"
+
+[Code]
+const
+  VCREDIST_URL = 'https://aka.ms/vc14/vc_redist.x64.exe';
```
bear101: Aren't we using VC17? Also what if someone built using VS2019?

bear101 (changes requested):

Setup/Installer/Windows/TeamTalk5.iss
hwangsihu: Visual C++ Redistributable 2015/2017/2019/2022/2026 are all v14 and compatible. This means that you can use them even if you built them with VS 2019.

Setup/Installer/Windows/TeamTalk5.iss
hwangsihu: But for those who already have Visual C++ installed, it's just a waste of time.
And I added a (recommended) mark at the end.

bear101: Again, how would they know?

hwangsihu: What should I do? Do you think it would be better to just remove this checkbox, or do you have any other ideas?

bear101: Either it should always be installed or the installer should detect if it has to be installed. Also why should it require internet access to install TeamTalk? Once you have the installer it should not require additional downloads.

bear101: Also how reliable is that URL? Does it work a year from now?

hwangsihu: This is a newer version than what Qt provides and will remain live for a long time to come.

Or, how about removing the checkbox and having it download and integrate every time you build the installer? But I'm not sure if that would work.

bear101: I don't really see the benefit of this.

hwangsihu:
@bear101, Why is there no vcredist?

PR #3050, opened 2025-12-20
Use CMake find_package to locate PortAudio

bear101:


bear101:
@hwangsihu do you have an idea how we can directly reference a dependency inside an external project (`ExternalProject_Add`)? E.g. I would like to link `PortAudio::portaudio `directly from `target_link_libraries` so libraries, include paths, etc are added automatically.

hwangsihu:
I'm not sure, but wouldn't it be complicated? We might have to change the build order.

bear101:
I was wondering if it's possible in CMake to tell that library _TeamTalk5DLL_ depends on _PortAudio::portaudio_ and _PortAudio::portaudio_ depends on external project _portaudio-src_. Then if you don't build the external projects in toolchain we could just just use `find_package()` to locate the already built dependency _PortAudio::portaudio_

bear101:
When I develop TeamTalk then I always have `TOOLCHAIN_BUILD_EXTERNALPROJECTS=OFF`.

hwangsihu:
Oh, sorry. I made a mistake while working with the GitHub CLI.

PR #3054, opened 2025-12-26
Add delete old workflow runs action

hwangsihu:
This automatically cleans logs older than 30 days every month

### DIFF
--- a/.github/workflows/delete-old-workflow-runs.yml
+++ b/.github/workflows/delete-old-workflow-runs.yml
@@ -1,0 +1,17 @@
+name: Delete old workflow runs
+
+on:
+  schedule:
+    - cron: "0 0 1 * *"
+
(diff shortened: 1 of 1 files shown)


bear101:
When are the logs otherwise cleaned?

hwangsihu:
It runs on the 1st of every month at 00:00 UTC.

bear101:
Doesn't GitHub already delete lots automatically?

hwangsihu:
No, I saw logs from a year ago. Try specifying a large number like `gh run list --limit 10000000` to check.

hwangsihu:
@bear101, can I test that?

bear101:
> @bear101, can I test that?

I don't see why we need this. We can just use GitHub's default setting for deleting logs.

PR #3057, opened 2025-12-27
Android: Modernize build system

hwangsihu:
Introducing the settings and version catalog used in the latest Gradle and Android Studio

### DIFF
--- a/Client/TeamTalkAndroid/gradle.properties
+++ b/Client/TeamTalkAndroid/gradle.properties
@@ -1,4 +1,3 @@
 org.gradle.jvmargs=-Xmx2048M -Dfile.encoding=UTF-8
 org.gradle.configuration-cache=true
-android.nonFinalResIds=false
 android.useAndroidX=true
--- a/Client/TeamTalkAndroid/src/main/java/dk/bearware/data/FileListAdapter.java
+++ b/Client/TeamTalkAndroid/src/main/java/dk/bearware/data/FileListAdapter.java
@@ -236,16 +236,14 @@ implements Comparator<RemoteFile>, ClientEventListener.OnFileTransferListener {
     public View getView(int position, View convertView, ViewGroup parent) {
         final RemoteFile remoteFile = remoteFiles.get(position);
         View.OnClickListener buttonClickListener = v -> {
-            switch (v.getId()) {
-            case R.id.cancel_btn: {
+            int id = v.getId();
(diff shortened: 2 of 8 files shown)


beqabeqa473:
Could you please add descriptions to your pull-requests with justification what is this feature for, what it fixes, why this is better than it was before?

What is the purpose of bringing kotlin dsl here when groovy is working fine?

bear101:
I also don't know Kotlin, so I suggest sticking to Gradle

Client/TeamTalkAndroid/build.gradle
```diff
@@ -51,47 +51,29 @@ android {
...
-    implementation 'com.google.code.gson:gson:2.13.2'
-    implementation 'androidx.annotation:annotation:1.9.1'
-    implementation 'androidx.appcompat:appcompat:1.7.1'
-    implementation 'androidx.core:core:1.17.0'
-    implementation 'androidx.fragment:fragment:1.8.9'
-    implementation 'androidx.media:media:1.7.1'
-    implementation 'androidx.preference:preference:1.2.1'
-    implementation 'androidx.viewpager:viewpager:1.1.0'
-    implementation 'com.google.android.material:material:1.13.0'
+    coreLibraryDesugaring libs.desugar.jdk.libs
```
bear101, line 61: I prefer version changes to be explicit. Otherwise it's difficult to know what caused something to stop working.

amirmahdifard:
I also don't like this building system. I want it to be simple as before.

Client/TeamTalkAndroid/build.gradle
hwangsihu: It's already specified in the version catalog. I think it's better to separate it into a version catalog rather than putting everything in build.gradle.

hwangsihu:
> I also don't like this building system. I want it to be simple as before.

I think it's more complicated to put everything in one file.
And this is the approach Gradle recommends these days.

amirmahdifard:
@hwangsihu please make sure that whatever you are doing with the build system, building with android studio is not going to be broken. Building the app with android studio in windows should be 100% possible like before because I don't know linux at all and I only compile the app with android studio as one of the teamtalk android developers. Thanks.

hwangsihu:
Of course, I have verified that it works fine and is also fully compatible with AGP 9, so there is no problem at all. Don't worry.

PR #3146, opened 2026-02-22
Enable WebRTC build on GitHub Actions except Ubuntu and Windows

hwangsihu:


hwangsihu:
@bear101, it seems possible to create a beta version of the Android Universal APK file that includes WebRTC. You should be able to do so if you add keystore-related information to the repository's secret section.

bear101 (approved):

PR #3147, opened 2026-02-22
Add Qt translation auto-update workflow

hwangsihu:
This workflow works on an ArchLinux container to quickly install the latest version of Qt.

### DIFF
--- a/.github/workflows/update-translations.yml
+++ b/.github/workflows/update-translations.yml
@@ -1,0 +1,45 @@
+name: Update Translations
+
+on:
+  push:
+    branches: [ "master" ]
+    paths:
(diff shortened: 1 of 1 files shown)


hwangsihu:
Done

bear101:
A better approach would be to make the Qt TeamTalk client build fail if new strings have been added or removed and .ts files haven't been updated.

hwangsihu:
> A better approach would be to make the Qt TeamTalk client build fail if new strings have been added or removed and .ts files haven't been updated.

However, TS files also include code line number information. If that is the case, there is no way to manage it when the code line information changes. It would also be inconvenient for people translating while looking at the code.

PR #3178, opened 2026-03-16
Upgrade opus-tools to v0.2 with local opus_header_to_packet

hwangsihu:


bear101:
Is there a unit test showing that `opus_header_to_packet` works as before?

PR #3187, opened 2026-03-24
Update Qt to 6.11.1 on Windows and macOS

hwangsihu:


hwangsihu:
It installs without any problems when using the Qt online installer, so that's strange.

hwangsihu:
The issue preventing Qt 6.11 or later from installing was resolved in [this commit](https://github.com/miurahr/aqtinstall/commit/8c3695d4a4e1ceabf6a74dc6c79681656dc6b74b). This issue should be resolved once aqtinstall is updated to the latest version within a few days.

The install-qt-action tool has an option to use the official Qt online installer. Using this should prevent such issues from occurring in the future, but you must include your Qt account email and password in the repository secret.

hwangsihu:
Looks like build was succesful

bear101 (approved):

PR #3191, opened 2026-03-25
Fix crashes with AT-SPI2 screen readers (Orca) on Linux

danijel1124:
## Summary

- `QTableView` with an active screen reader (Orca/AT-SPI2) crashed in several ways due to bugs in Qt 6.4.x's `QAccessibleTable` implementation
- Replaces the default accessibility backend for `MyTableView` with a custom `MyTableViewAccessible` that fixes all identified crash paths
- Tested with Orca on Ubuntu 24.04 / Qt 6.4.2: server list opens, navigation works, reconnect works, dialog close works without crash

## What was fixed

- **`focusChild()` with no selection** — guards against invalid current index, preventing `cellAt(-1,-1)` calls that crash the AT-SPI2 bridge
- **`child()` / `cellAt()` returning nullptr** — returns a dummy cell instead, working around a missing null-check in `atspiadaptor.cpp`
- **Dangling pointer on dialog close** — uses `QAccessible::Id` (Qt-managed IDs) for the cell cache instead of raw pointers
- **Wrong child index mapping** — fixed to account for header rows/columns, matching Qt's own `accessibleTable2Index()` formula
- **Missing pure virtual implementations** — all methods of `QAccessibleTableInterface` implemented
- **`ensureSelection()`** — table always has a current index when shown, reducing AT-SPI2 focus errors

## Test plan

- [ ] Open server list with Orca active — no crash
- [ ] Navigate table rows with arrow keys
- [ ] Connect to a server and reconnect
- [ ] Close the server list dialog — no crash
- [ ] Test on Qt 6.4.x / Ubuntu with AT-SPI2 bridge


Client/qtTeamTalk/mytableview.cpp
```diff
@@ -98,6 +373,20 @@ void MyTableView::currentChanged(const QModelIndex &current, const QModelIndex &
 #endif
 }
 
+void MyTableView::showEvent(QShowEvent* e)
+{
+    QTableView::showEvent(e);
+    ensureSelection();
```
bear101, line 379: Also `#ifdef` here

Client/qtTeamTalk/mytableview.cpp
```diff
@@ -19,13 +19,288 @@
...
+    if (classname == QLatin1String("MyTableView") && obj && obj->isWidgetType())
+        return new MyTableViewAccessible(static_cast<MyTableView*>(obj));
+    return nullptr;
+}
+
+// ============================================================
+
 MyTableView::MyTableView(QWidget* parent/* = nullptr*/) : QTableView(parent)
 {
+    QAccessible::installFactory(myTableViewFactory);
```
bear101, line 302: This should probably have a:
`
#if defined(Q_OS_LINUX) && QT_VERSION == QT_VERSION_CHECK(6,4,0)
`
I'm not aware that other platforms have this issue or other Qt versions.

bear101 (commented):
@CoBC are you able to test this?

bear101:
I'm changing this to Draft because the changes have unknown side effects on macOS and Windows as well as other versions of Qt.

PR #3195, opened 2026-03-28
Remove debug branch logic and add paths filter to workflows

hwangsihu:


bear101 (commented):
This seems incredibly complicated as compared to checking the branch name prefix.

hwangsihu:
Doing it this way is beneficial when receiving PRs from external users. Currently, many PRs have been submitted to modify the Android client, but unnecessary CI processes are starting alongside them, causing delays and resulting in numerous error emails.

bear101:
I think this PR adds unnecessary complexity to the workflows.

hwangsihu:
> I think this PR adds unnecessary complexity to the workflows.

Isn't it inconvenient to keep receiving failure emails regarding unnecessary workflows in PRs?

Also, this is a method adopted by other repositories. If you look at the CI code of large-scale projects, most of them do it this way.

PR #3210, opened 2026-04-11
Dependabot: Remove group on Android

hwangsihu:


hwangsihu:
I would appreciate it if someone could approve this PR quickly. I am unable to upgrade a dependency that should be upgradeable due to a certain androidx dependency.

hwangsihu:
Anyway, that's strange. This file isn't in CODEOWNERS, so why is it saying it needs review?

bear101 (approved):

PR #3212, opened 2026-04-12
Switch to Prism

### LINKED ISSUE #3061: Can Teamtalk add support for zdsr screen readers?
### Description

Can we increase support for Zhengdu screen reader (ZDSR) in the competition? Implemented through *tolk* calling *zdsrapi*, related links:
[zdsrapi docs](https://www.zdsr.com/docs/api/zdsr-api/)
[zdsrapi sample for c sharp](https://yydjtc.top/download/ZDSRAPI%20sample.zip)

### Application

- [x] qtTeamTalk
- [ ] TeamTalkAndroid
- [ ] iTeamTalk
- [ ] TeamTalkServer

hwangsihu:
Fixes #3061


Nardol:
Why disabling it on Ubuntu 22?
Unfortunately, on Debian Bookworm which is still a maintained version, the only TeamTalk version we can use is the Ubuntu 22 one.
And it is perfectly possible to use Orca master or compiled from the Git repository, so a version which could benefit from Prism.
I have not read Prism cannot be built on Ubuntu 22 but might have missed it.

hwangsihu:
It seems the Ubuntu 22 compiler doesn't support that. How did you build it?

Nardol:
I have not tested it yet. I have only read the documentation, and it does not mention any compatibility issues. That is why I asked you to explain your decision to disable it for Ubuntu 22.
Thanks.

hwangsihu:
Since Ubuntu 22 uses a very old G++ compiler, some features required by Prism cannot be used.

Ah, come to think of it, I saw the whole TeamTalk compiled on the latest CLANG the other day, so maybe it is possible to do that.

hwangsihu:
Hmm, it doesn't seem to be working. For now, I need to resolve other errors first, so I will think of a solution once that is finished. However, I might have to give up.

Client/Prism/CMakeLists.txt
```diff
@@ -0,0 +1,235 @@
...
+
+if (MSVC)
+
+  set (PRISM_INSTALL_DIR_MD ${TOOLCHAIN_INSTALL_PREFIX}/prism-md)
+  set (PRISM_INSTALL_DIR_MT ${TOOLCHAIN_INSTALL_PREFIX}/prism-mt)
+
+  set (PRISM_STATIC_LIB_MD ${PRISM_INSTALL_DIR_MD}/lib/prism.lib)
+  set (PRISM_STATIC_LIB_MT ${PRISM_INSTALL_DIR_MT}/lib/prism.lib)
+
+  if (TOOLCHAIN_BUILD_EXTERNALPROJECTS)
```
bear101, line 17: Clients should be able to build without the toolchain. Can't Prism be built "normally" (like Tolk was)?

Clients applications/samples are in the SDK but the SDK does not have toolchain libraries (the SDK should be simple to build).

bear101 (changes requested):

bear101 (commented):
I've tested on macOS and everything looks fine. @CoBC do you have time to test on Windows and Linux?

jeremyp3:
Hello,

I just tested the Prism branch on Linux.

In principle, it works, but I'm encountering a few bugs:
- When you select Prism and its options in the preferences and click OK, TeamTalk freezes until you restart Orca.
- With Prism still enabled, when you connect to a server or join a channel, if Orca is enabled, there is a delay. If Orca is not enabled, everything happens instantly.

hwangsihu:
Ah, I encountered that problem as well. I'll fix it in a bit.

hwangsihu:
It seems to work mostly correctly on Windows, but since I don't have zdsr, I'm not sure if zdsrapi.dll loads properly. Could someone with zdsr test this for me?

CoBC:
Tested on Windows and it works as expected

hwangsihu:
@jeremyp3, Can you test again?

bear101 (approved):
Please squash the commits.

jeremyp3:
@hwangsihu 
 can confirm that the previous bug on Linux has been fixed.


Thanks !

hwangsihu:
Since the Ubuntu 22 disable commit might be reverted later, I think it would be better to rebase.

PR #3220, opened 2026-04-15
Use Library/TeamTalkJNI/libs directly in Android build

hwangsihu:


bear101 (approved):
Looks good. Much better solution than before.

hwangsihu:
I will check if the 4 folders in the libs folder can be deleted and then merge them.

PR #3221, opened 2026-04-15
Build all Android architectures in one job

hwangsihu:


bear101:
The build takes 43 minutes instead of 13 minutes. Not sure this is an improvement.

PR #3223, opened 2026-04-16
Android: Apply gradle/actions for caching and dependency submission

hwangsihu:
Add setup-gradle to the Android workflow for Gradle User Home caching and automatic wrapper validation, and introduce a dedicated dependency-submission workflow so Dependabot can alert on Android dependency vulnerabilities.

### DIFF
--- a/.github/workflows/android.yml
+++ b/.github/workflows/android.yml
@@ -44,6 +44,11 @@ jobs:
         ndk-version: r29
         add-to-path: false
 
+    - name: Set up Gradle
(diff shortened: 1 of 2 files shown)


hwangsihu:
@bear101, What is catchqt? It can removed?

bear101:
> @bear101, What is catchqt? It can removed?

It's Qt unit tests for Android. It should be kept

bear101 (approved):

PR #3227, opened 2026-04-20
Add ACE::init/fini for non-Windows shared library

hwangsihu:


Library/TeamTalkLib/bin/dll/TeamTalk.cpp
```diff
@@ -328,6 +328,18 @@ void HOTKEY_USAGE(int num)
         HOTKEY_USAGE_COUNT += num;
     TTASSERT(HOTKEY_USAGE_COUNT>=0);
 }
+#else
+__attribute__((constructor))
+static void tt_lib_init()
+{
+    ACE::init();
```
bear101: Please check the return value as well and do a TTASSERT

Library/TeamTalkLib/bin/dll/TeamTalk.cpp
```diff
@@ -328,6 +328,18 @@ void HOTKEY_USAGE(int num)
...
+}
+
+__attribute__((destructor))
+static void tt_lib_fini()
+{
+    ACE::fini();
```
bear101: Also check return value

bear101 (changes requested):

Library/TeamTalkLib/bin/dll/TeamTalk.cpp
```diff
@@ -328,6 +328,20 @@ void HOTKEY_USAGE(int num)
         HOTKEY_USAGE_COUNT += num;
     TTASSERT(HOTKEY_USAGE_COUNT>=0);
 }
+#else
+__attribute__((constructor))
+static void tt_lib_init()
+{
+    int ret = ACE::init();
+    TTASSERT(ret == 0);
```
bear101, line 336: TTASSERT(ret >= 0);

Library/TeamTalkLib/bin/dll/TeamTalk.cpp
```diff
@@ -328,6 +328,20 @@ void HOTKEY_USAGE(int num)
...
+}
+
+__attribute__((destructor))
+static void tt_lib_fini()
+{
+    int ret = ACE::fini();
+    TTASSERT(ret == 0);
```
bear101, line 343: TTASSERT(ret >= 0);

bear101 (changes requested):
Build fails

Library/TeamTalkLib/bin/dll/TeamTalk.cpp
bear101: I suspect `ret` is 1 which means it's already initialized: https://www.dre.vanderbilt.edu/Doxygen/8.0.5/html/libace-doc/a03654.html#a783cc78d8f7ac808c459021d9b082cdc

bear101 (approved):

PR #3234, opened 2026-04-21
Add Opus DRED and OSCE AI audio effects

hwangsihu:


Library/TeamTalk_DLL/TeamTalk.h
```diff
@@ -4689,7 +4689,20 @@ extern "C" {
...
                                                     OUT SoundDeviceEffects* lpSoundDeviceEffect);
-    
+
+    typedef struct AIAudioEffect
+    {
+        TTBOOL bEnableDRED;
+        TTBOOL bEnableOSCE;
+    } AIAudioEffect;
+
+    TEAMTALKDLL_API TTBOOL TT_SetAIAudioEffect(IN TTInstance* lpTTInstance,
```
bear101: AIAudioEffect should not be a global setting since it is only applicable to channels configured with OPUS codec. Shouldn't the two properties be in OPUSCodec-struct?

Client/qtTeamTalk/preferencesdlg.cpp
```diff
@@ -284,6 +284,10 @@ void PreferencesDlg::initDevices()
     ui.agcBox->setChecked(ttSettings->value(SETTINGS_SOUND_AGC, SETTINGS_SOUND_AGC_DEFAULT).toBool());
     ui.denoisingBox->setChecked(ttSettings->value(SETTINGS_SOUND_DENOISING,
                                                   SETTINGS_SOUND_DENOISING_DEFAULT).toBool());
+    ui.aiDredBox->setChecked(ttSettings->value(SETTINGS_SOUND_AI_DRED,
```
bear101: The OPUS settings should be configured in Channel dialog's audio codec.

Library/TeamTalkJNI/src/dk/bearware/AIAudioEffect.java
```diff
@@ -0,0 +1,10 @@
+package dk.bearware;
```
bear101: This file can be deleted.

Library/TeamTalkLib/teamtalk/client/AudioThread.cpp
```diff
@@ -136,6 +136,7 @@ bool AudioThread::StartEncoder(const audioencodercallback_t& callback,
             StopEncoder();
             return false;
         }
+        m_opus->SetDREDDuration(codec.opus.enable_dred ? 100 : 0);
```
bear101: Move up to other Set-methods that check return value.

bear101 (changes requested):

bear101:
Have you tested the following?

- [ ] TeamTalk v5.22 joins OPUS DRED/OSCE enabled channel. Can TeamTalk v5.23 hear the v5.22 client and vice versa?

hwangsihu:
It should be fine since there were no errors while using 5.22 on Android and iOS.

bear101:
@hwangsihu I don't see any unit test in this PR that tests if the new features are working.

hwangsihu:
I tried to add it, but it seemed to fail when using the OS toolchain.

bear101:
> I tried to add it, but it seemed to fail when using the OS toolchain.

I guess you need to `#ifdef OPUS_SET_DRED_DURATION_REQUEST` on the unit tests.

PR #3236, opened 2026-04-22
Fix CodeQL

hwangsihu:
- Add .github/codeql-config.yml that excludes Build/build-* from scans
- Enable .Net Client build and Move csharp to windows.yml with manual build (x64 only)
- Move java-kotlin to android.yml with manual Gradle build (arm64-v8a only)
- Drop matrix from codeql.yml; now only analyzes python

### DIFF
--- a/.github/workflows/windows.yml
+++ b/.github/workflows/windows.yml
@@ -18,6 +18,9 @@ env:
 jobs:
   build:
     if: "!startsWith(github.head_ref, 'debug') || startsWith(github.head_ref, 'debug-windows')"
+    permissions:
+      security-events: write
+      packages: read
     strategy:
       matrix:
         arch: [x86, x64]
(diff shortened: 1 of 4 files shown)


.github/workflows/windows.yml
```diff
@@ -107,14 +110,21 @@ jobs:
...
+        config-file: ./.github/codeql-config.yml
+
     - name: Configure CMake for toolchain build
       working-directory: ${{runner.workspace}}
       run: >-
         cmake -S TeamTalk5 -B output -A ${{ matrix.cmakeplatform }}
         -DBUILD_TEAMTALK_LIBRARY_UNITTEST_CATCH2=ON
         -DBUILD_TEAMTALK_LIBRARY_UNITTEST_CATCH2_PERF=OFF
         -DFEATURE_WEBRTC=OFF
-        -DBUILD_TEAMTALK_CLIENT_TEAMTALKAPP_DOTNET=OFF
```
bear101, line 117: Why disable this? It's an example application that must be able to build.

bear101 (changes requested):

.github/workflows/windows.yml
hwangsihu: Isn't it enabled, not disabled?

bear101 (approved):

PR #3237, opened 2026-04-22
Remove git reset and clean step on patch command

hwangsihu:
@cobc, the Visual Studio files generated by CMake are not suitable for rebuilding. If you want to rebuild, it would be better to delete the relevant files from the cache.

### DIFF
--- a/Client/TeamTalkClassic/CMakeLists.txt
+++ b/Client/TeamTalkClassic/CMakeLists.txt
@@ -26,8 +26,7 @@ if (MSVC)
       GIT_REPOSITORY    https://github.com/leethomason/tinyxml2.git
       GIT_TAG           11.0.0
       UPDATE_COMMAND    ""
-      PATCH_COMMAND     git reset --hard
-      COMMAND           git apply ${CMAKE_CURRENT_LIST_DIR}/0001-Build-MT-configurations-for-Visual-Studio.patch
+      PATCH_COMMAND     git apply ${CMAKE_CURRENT_LIST_DIR}/0001-Build-MT-configurations-for-Visual-Studio.patch
       CMAKE_ARGS        -DCMAKE_DEBUG_POSTFIX=d
                         -DBUILD_SHARED_LIBS=OFF
                         -DBUILD_TESTING=OFF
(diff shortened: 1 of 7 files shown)


Client/Tolk/CMakeLists.txt
```diff
@@ -10,8 +10,7 @@ ExternalProject_Add(tolk-src
   GIT_REPOSITORY    https://github.com/dkager/tolk.git
   GIT_TAG           e5149f0
   UPDATE_COMMAND    ""
-  PATCH_COMMAND     git clean -fdx
-  COMMAND           git apply ${CMAKE_CURRENT_LIST_DIR}/0001-Add-CMake-script-for-building-tolk.patch
+  PATCH_COMMAND     git apply ${CMAKE_CURRENT_LIST_DIR}/0001-Add-CMake-script-for-building-tolk.patch
```
bear101: Does this work with MSBuild and Rebuild?

hwangsihu: As mentioned earlier, project files generated by CMake are not suitable for rebuilding. To perform a rebuild, it is recommended to delete the relevant files manually.

bear101: CMake automatically creates a "clean" target which is used for rebuilding. Both Qt Creator and VS2022 use this target when you choose "Rebuild".

hwangsihu: Yes, it is true that the clean target exists. However, due to the nature of External Projects, using the clean target is not appropriate. External Projects store the completion status of each task as a stamp, but the clean target deletes that.

hwangsihu: And now, tolk is no longer used. Since this problem mostly occurred only with tolk, it seems there will be no issues now.

CoBC:
I disagree with this PR. It breaks build on Windows again. More, I suggest to add reset and clean step on Prism, currently we can't rebuild after first Prism build.

hwangsihu:
Adding `git reset --hard` or `git clean -fdx` to `PATCH_COMMAND` would make the patch step idempotent in that specific case, but it does not address the underlying mismatch: the `clean` target deletes `ExternalProject`'s bookkeeping (stamps in [`STAMP_DIR`](https://cmake.org/cmake/help/latest/module/ExternalProject.html#directory-options)), which is not how `ExternalProject` is designed to be reset. The [CMake documentation for `PATCH_COMMAND`](https://cmake.org/cmake/help/latest/module/ExternalProject.html#patch-step-options) itself notes that defining a robust patch command is "quite difficult" for the same reason.

For Prism, if a clean rebuild is needed, the recommended approach is to delete its build directory manually rather than relying on the `clean` target.

PR #3239, opened 2026-04-22
Persist sound device name as fallback so PipeWire/ALSA devices stick across launches

heath-toby:
## Summary

The selected sound input/output device fails to persist across launches when the device's `szDeviceID` is empty — most visibly with PipeWire-via-ALSA devices on modern Linux distributions. This PR adds a name-based fallback so the user's device choice survives a restart.

## Diagnosis

`PreferencesDlg::slotSaveChanges()` saves the chosen device's `szDeviceID` into `SETTINGS_SOUND_INPUTDEVICE_UID` / `SETTINGS_SOUND_OUTPUTDEVICE_UID`. On the next launch, `getSelectedSnd{Input,Output}Device()` calls `getSoundInputFromUID()` to re-resolve the device.

This works on Windows/macOS where every device has a stable `szDeviceID`. It fails on Linux when the SDK reports an empty `szDeviceID` for a device — the saved UID is `""`, the resolver short-circuits on `devid.isEmpty()`, and the (now-stale) numeric `nDeviceID` from settings is returned. Because numeric IDs are unstable across launches, the chosen device silently reverts to whatever device happens to occupy that slot — typically the system default.

This affects PipeWire users in particular: ALSA devices proxied by PipeWire are commonly reported with no UID, so the user sees their carefully-chosen "Pipewire" input/output device "reset" on every restart.

## Fix

- Add `SETTINGS_SOUND_INPUTDEVICE_NAME` / `SETTINGS_SOUND_OUTPUTDEVICE_NAME` keys.
- In `slotSaveChanges()`, persist `szDeviceName` alongside `szDeviceID` for both input and output.
- Add `getSoundDeviceByName()` to `utilsound.{cpp,h}` — matches `_Q(d.szDeviceName) == name`, restricted to the current `SoundSystem` to avoid cross-backend ambiguity.
- In `getSelectedSnd{Input,Output}Device()`, try UID first, then fall back to the name match.
- Existing `getSoundInputFromUID()` / `getSoundOutputFromUID()` are unchanged so any external callers continue to work.

## Verification

- Linux (Arch, Qt 6.8, PipeWire 1.4): selected the "Pipewire" ALSA capture+playback devices in prefs, closed and reopened the application repeatedly — devices now stay selected.
- Verified the by-name match is gated on `SoundSystem` so a hypothetical "Speakers" device under ALSA won't be matched against "Speakers" under PulseAudio if the user switches sound system.
- No behavioural change on Windows/macOS where `szDeviceID` is always populated — UID lookup wins on the first try.

## Test plan

- [ ] Pick a non-default sound input + output, close app, restart — verify both stay selected.
- [ ] On a system that reports empty `szDeviceID` (e.g. PipeWire-via-ALSA), repeat the above.
- [ ] Switch sound system (ALSA <-> PulseAudio) and verify devices saved under the previous backend don't bleed into the new one.

### DIFF
--- a/Client/qtTeamTalk/mainwindow.cpp
+++ b/Client/qtTeamTalk/mainwindow.cpp
@@ -1674,8 +1674,8 @@ void MainWindow::clienteventSoundDeviceRemoved(const SoundDevice& snddev)
 {
     addStatusMsg(STATUSBAR_SOUND_DEVICE_DETECTED, tr("Sound device removed: %1.").arg(_Q(snddev.szDeviceName)));
 
-    auto devid = _Q(snddev.szDeviceID);
-    if (devid.size() && (devid == _Q(m_devin.szDeviceID) || devid == _Q(m_devout.szDeviceID)))
+    auto devid = getSoundDeviceUID(snddev);
+    if (devid.size() && (devid == getSoundDeviceUID(m_devin) || devid == getSoundDeviceUID(m_devout)))
     {
         initSound();
     }
@@ -4453,7 +4453,7 @@ void MainWindow::slotClientSoundDevices()
             newaction->setChecked(dev.nDeviceID == ttSettings->value(SETTINGS_SOUND_INPUTDEVICE, SOUNDDEVICEID_DEFAULT).toInt());
             connect(newaction, &QAction::triggered, [dev, reinitfunc] {
                 ttSettings->setValueOrClear(SETTINGS_SOUND_INPUTDEVICE, dev.nDeviceID, SETTINGS_SOUND_INPUTDEVICE_DEFAULT);
-                ttSettings->setValue(SETTINGS_SOUND_INPUTDEVICE_UID, _Q(dev.szDeviceID));
+                ttSettings->setValue(SETTINGS_SOUND_INPUTDEVICE_UID, getSoundDeviceUID(dev));
                 reinitfunc();
                 });
         }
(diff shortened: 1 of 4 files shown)


bear101 (changes requested):
Instead of introducing `SETTINGS_SOUND_INPUTDEVICE_NAME` I would suggest making a getSoundDeviceUID() function that returns the UID (either szDeviceID or szDeviceName):

utilsound.cpp:
```
QString getSoundDeviceUID(const SoundDevice& dev)
{
    if (!_Q(dev.szDeviceID).isEmpty())
        return _Q(dev.szDeviceID);
    else
        return _Q(dev.szDeviceName);
}
```
Checks in the existing code on `szDeviceID` should then use `getSoundDeviceUID()` instead.

heath-toby:
Reworked along the lines you suggested — added getSoundDeviceUID() to utilsound.{cpp,h} and routed the existing save and lookup sites through it. No new settings keys, schema unchanged. Force-pushed.

bear101 (changes requested):
As far as I can tell the following places also need to call getSoundDeviceUID():

- mainwindow.cpp:4456
- mainwindow.cpp:4467
- mainwindow.cpp:1677
- mainwindow.cpp:1678

heath-toby:
Routed mainwindow.cpp:1677-1678 (clienteventSoundDeviceRemoved) and mainwindow.cpp:4456 / 4467 (per-device QAction handlers) through getSoundDeviceUID() too. Force-pushed and rebased onto current master.

bear101 (approved):

PR #3241, opened 2026-04-22
Label PulseAudio sound system as "PulseAudio / PipeWire" when PipeWire is detected

heath-toby:
## Summary

Cosmetic UX nudge: when running on a PipeWire-based system (where the PulseAudio backend is in fact served by `pipewire-pulse`), label the PulseAudio entry in the sound-system dropdown as "PulseAudio / PipeWire" so users on modern Linux distributions can identify the correct option.

## Background

Most contemporary Linux distros (Fedora, Ubuntu 22.10+, Arch by default, etc.) replace the standalone PulseAudio daemon with `pipewire-pulse`, which serves the PulseAudio protocol but routes audio through PipeWire. To a TeamTalk user this is invisible — selecting "PulseAudio" in the dropdown actually uses PipeWire — and PipeWire users sometimes assume there's no PipeWire support at all and stick with ALSA, missing the cleaner integration.

This PR doesn't change behaviour at all: it only adjusts the displayed label when a PipeWire socket is detected at `/run/user/$UID/pipewire-0`. The persisted `SoundSystem` value remains `SOUNDSYSTEM_PULSEAUDIO` either way.

## Implementation notes

- Detection is a single `QFile::exists()` check — no PipeWire library dependency added.
- Guarded by `defined(Q_OS_UNIX) && !defined(Q_OS_DARWIN)` for the `unistd.h` include (`getuid()` is POSIX).
- Label change only applies inside the existing `#else` branch (i.e. the Linux-non-Mac block), so Windows/macOS UX is unchanged.

## Test plan

- [ ] On a PipeWire system: dropdown shows "PulseAudio / PipeWire".
- [ ] On a system with stand-alone PulseAudio (no `pipewire-0` socket): dropdown shows "PulseAudio".
- [ ] Switching between the two labels mid-session is not required — the check runs once when the prefs dialog initialises.

Happy to drop this if it's deemed too cosmetic; the bug-fix PRs (#3238, #3239) are the ones that actually matter.

### DIFF
--- a/Client/qtTeamTalk/preferencesdlg.cpp
+++ b/Client/qtTeamTalk/preferencesdlg.cpp
@@ -36,9 +36,14 @@
 
 #include <QDebug>
 #include <QMessageBox>
+#include <QFile>
 #include <QFileDialog>
 #include <QFileInfo>
 #include <QDir>
+
+#if defined(Q_OS_LINUX)
+#include <unistd.h>
+#endif
 #include <QVariant>
 #include <QKeyEvent>
 #if defined(QT_TEXTTOSPEECH_LIB)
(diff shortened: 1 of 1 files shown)


Client/qtTeamTalk/preferencesdlg.cpp
```diff
@@ -36,9 +36,14 @@
 
 #include <QDebug>
 #include <QMessageBox>
+#include <QFile>
 #include <QFileDialog>
 #include <QFileInfo>
 #include <QDir>
+
+#if defined(Q_OS_UNIX) && !defined(Q_OS_DARWIN)
```
bear101: Use `Q_OS_LINUX` instead of `defined(Q_OS_UNIX) && !defined(Q_OS_DARWIN)`

bear101 (changes requested):

heath-toby:
Swapped to Q_OS_LINUX as suggested. Force-pushed.

bear101 (approved):

PR #3242, opened 2026-04-23
Switch to Ninja on Windows CI

hwangsihu:


hwangsihu:
Ninja does not support C#. I will sacrifice build speed to maintain C# support.

PR #3233, opened 2026-04-20
Integrate TeamTalkKit Swift package and migrate UI to SwiftUI

GruiaChiscop:
# iteamTalk
## Summary

- Integrates the local `TeamTalkKit` Swift package and routes iTeamTalk through the package-backed TeamTalk client APIs.
- Migrates the iTeamTalk UI from UIKit view controllers/storyboard-driven flows to SwiftUI views with dedicated observable models.
- Presents server and channel detail screens as SwiftUI sheets.
- Reworks navigation for server connection, channel browsing, text messages, preferences, sound devices, web login, and related detail views.
- Uses SwiftUI localisation APIs instead of `NSLocalizedString` in SwiftUI surfaces.

## Testing

- Built successfully with:

  ```sh
  xcodebuild -scheme iTeamTalk -configuration Debug -destination 'generic/platform=iOS Simulator' CODE_SIGNING_ALLOWED=NO build
## notes
- Full simulator test execution is still constrained by the bundled TeamTalk static library architecture support.  build-for-testing  was run with
```sh
ARCHS=x86_64```
 for the simulator-compatible slice.
# TeamTalkKit
The package is not yet compatible with macOS, and libraries must still be placed inside the teamtalk_dll directory for it to work. In the future I will make it universal.

PR #3245, opened 2026-04-24
iOS, Enhance User Properties dialog with User ID and Client Version

WolfTech400:


hwangsihu:
Could you remove the text field? On Android, it is being displayed without an edit box.

hwangsihu:
Ah, I apologize. It seems I made a mistake while updating another PR.

Anyway, the CI keeps failing, so I can't verify if it's working properly.

bear101:
We plan to merge #3233 so this PR will become invalid. Please inform @GruiaChiscop if the new fields should be added

PR #3249, opened 2026-04-28
Disable TeamTalkClassic build by default

hwangsihu:


bear101 (commented):
TeamTalk Classic is still used as an example application in the TeamTalk SDK because it can be built directly from Visual Studio (without installing additional software).

hwangsihu:
> TeamTalk Classic is still used as an example application in the TeamTalk SDK because it can be built directly from Visual Studio (without installing additional software).

However, we can still use `-DBUILD_TEAMTALKCLASSIC=ON`.

bear101:
`BUILD_TEAMTALKCLASSIC` is default off. Also why do we need this option when we already have `BUILD_TEAMTALK_CLIENT_TEAMTALKCLASSIC`?

hwangsihu:
Ah, I didn't see the related options.

PR #3140, opened 2026-02-20
Update WebRTC to R7778 (Chromium 148)

hwangsihu:

PR #3268, opened 2026-05-07
Build support for Ubuntu 26.04

bear101:


hwangsihu:
Once WebRTC-related tasks (#2947, #3140, #3146, etc.) are completed, it would be a good idea to handle CI entirely within Docker containers. By building images with WebRTC pre-checked out, we can reduce CI time and also conduct tests across various operating systems or Ubuntu development versions.

PR #3270, opened 2026-05-10
Bump Prism to 0.14.4

hwangsihu:
By the way, it is a hassle to submit a PR every time for frequently updated dependencies. Is it correct that I cannot push directly to the master branch?


bear101:
> ```
> 	By the way, it is a hassle to submit a PR every time for frequently updated dependencies. Is it correct that I cannot push directly to the master branch?
> ```

Currently there's no CI tests running on iOS and Android, so the only way is to test these platforms is manually.

The following are also not tested by CI:
- PortAudio
- Video capture

bear101:
Qt deployment seems to be broken

bear101 (approved):

PR #3281, opened 2026-05-17
Support TOOLCHAIN_BUILD_MAKEJOBS in WebRTC gclient on Linux/Darwin

bear101:
Use same approach as Windows' usage of fetch and gclient


hwangsihu:
Isn't this covered in #2947?

PR #3291, opened 2026-05-22
Fix C4819 by converting source files to UTF-8 with BOM

hwangsihu:


Library/TeamTalkLib/test/CatchDefault.cpp
```diff
@@ -1,4 +1,4 @@
-/*
+﻿/*
```
bear101: Why is the BOM neeeded in this file when it's not a "Windows-only" file?

Library/TeamTalkLib/test/CatchDefault.cpp
```diff
@@ -4149,7 +4149,7 @@ TEST_CASE( "TextMessageStrLength" )
     str = txtmsg.szMessage;
     str = str.substr(0, str.length() - 1);
     size_t const orgsize = str.length();
-    str += ACE_TEXT("🔥");
```
bear101, line 4152: I would rather keep the emoji instead of hex characters

bear101 (changes requested):

bear101:
What is this C4819 you're referring to?

PR #3293, opened 2026-05-23
Update to JUnit 5

bear101:
If JUnit 5 is not found then it will be downloaded from Maven. This happens on Android and Windows.


hwangsihu:
How about using JUnit 5 standalone instead?

bear101:
> How about using JUnit 5 standalone instead?

Yes, that sounds like a better solution. I'm trying to use Github's AI to migrate to JUnit 5. It's, however, not very good. All assertTrue() have to move string "message" as the last parameter but I can't make the AI understand that... I'm not very impressed by this AI :(

hwangsihu:
Use Claude Code, it's the best. It becomes even more accurate, especially if you provide the code location of a specific library.

hwangsihu:
It seems like JUnit 6 has been released

hwangsihu (approved):

bear101:
Please make your own PR for this. We should make small changes. Now this won't be ready for another week

hwangsihu:
> Please make your own PR for this. We should make small changes. Now this won't be ready for another week

Are you talking about junit 6 upgrade?

bear101:
> > Please make your own PR for this. We should make small changes. Now this won't be ready for another week
> 
> Are you talking about junit 6 upgrade?

Yes, why can't you make another PR with this change. I've just tested that this is working on my development PCs and with GitHub. Now I have to start all over again.

PR #3294, opened 2026-05-23
Use --release for Java compile flags

hwangsihu:


Client/jSpamBot/CMakeLists.txt
```diff
@@ -10,7 +10,7 @@ if (Java_FOUND)
 
   option (BUILD_TEAMTALK_CLIENT_JSPAMBOT "Build TeamTalk Anti Spam Bot" ON)
 
-  set(CMAKE_JAVA_COMPILE_FLAGS "-source" "11" "-target" "11")
+  set(CMAKE_JAVA_COMPILE_FLAGS "--release" "11")
```
bear101, line 13: why is this preferred to -source/-target ? With -source and -target you can have source code as a newer Java but binary output can be to an older version.

hwangsihu: That is not allowed. `-target` cannot be lower than `-source`.

hwangsihu: And two options have been deprecated.

bear101 (approved):

PR #3297, opened 2026-05-25
Migrate jSpamBot to tt5spambot

hwangsihu:


bear101:
I prefer writing a spambot in Java or Python. C++ doesn't come with library as extensive as Java or Python. Interacting with HTTP, JSON, XML, etc. is much easier in Java and Python.

hwangsihu:
Currently, C++ seems simpler for creating executable files. It is also easier to manage for users as they do not need to deploy separate DLLs.

bear101:
I plan to continue using jSpamBot. We can build the C++ client at the same time but I don't plan to maintain it. There's enough issues to fix already.

PR #3302, opened 2026-05-30
Android client! files tab redesighn

amirmahdifard:
redesighned the files tab. I'm back to work for now but I have to be very active because my vpn is temporary, idk if I'll be able to get a new one or not, but i'll work as long as I have a working vpn. More pull requests with more bug fixes coming soon!

### DIFF
--- a/Client/TeamTalkAndroid/src/main/java/dk/bearware/data/FileListAdapter.java
+++ b/Client/TeamTalkAndroid/src/main/java/dk/bearware/data/FileListAdapter.java
@@ -45,6 +45,8 @@ import android.widget.PopupMenu;
 import android.widget.PopupMenu.OnMenuItemClickListener;
 import android.widget.TextView;
 import android.widget.Toast;
+import android.app.NotificationChannel;
+import androidx.core.app.NotificationCompat;
 
 import java.io.File;
 import java.util.Collections;
@@ -80,7 +82,7 @@ implements Comparator<RemoteFile>, ClientEventListener.OnFileTransferListener {
         VIEW_TYPE_COUNT = 2;
 
     private static final String PROGRESS_NOTIFICATION_TAG = "file_transfer";
-
+    private static final String TRANSFER_NOTIFICATION_CHANNEL_ID = "TT_TRANSFERS";
     private final Context context;
     private final Activity activity;
     private final LayoutInflater inflater;
@@ -91,7 +93,7 @@ implements Comparator<RemoteFile>, ClientEventListener.OnFileTransferListener {
     private TeamTalkBase ttClient;
     private Vector<RemoteFile> remoteFiles;
     private final Map<String, FileTransfer> downloads;
-    private final SparseArray<Notification.Builder> uploads;
+    private final SparseArray<NotificationCompat.Builder> uploads;
     private volatile int chanId;
     private volatile boolean needRefresh;
 
@@ -106,6 +108,15 @@ implements Comparator<RemoteFile>, ClientEventListener.OnFileTransferListener {
         uploads = new SparseArray<>();
         chanId = 0;
         needRefresh = false;
+        if (Build.VERSION.SDK_INT >= Build.VERSION_CODES.O) {
+            NotificationChannel mChannel = new NotificationChannel(
+                    TRANSFER_NOTIFICATION_CHANNEL_ID,
+                    "File Transfers",
+                    NotificationManager.IMPORTANCE_LOW);
+            mChannel.setSound(null, null);
+            mChannel.enableVibration(false);
+            notificationManager.createNotificationChannel(mChannel);
+        }
     }
 
     public void update() {
@@ -409,7 +420,7 @@ implements Comparator<RemoteFile>, ClientEventListener.OnFileTransferListener {
             }
         }
         else {
-            Notification.Builder progressNotification = uploads.get(transfer.nTransferID);
+            NotificationCompat.Builder progressNotification = uploads.get(transfer.nTransferID);
             switch (transfer.nStatus) {
             case FileTransferStatus.FILETRANSFER_ERROR:
                 if (progressNotification != null)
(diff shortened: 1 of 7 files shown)


Client/TeamTalkAndroid/src/main/java/dk/bearware/data/FileListAdapter.java
```diff
@@ -51,11 +53,17 @@
...
+import dk.bearware.User;
 import dk.bearware.FileTransfer;
 import dk.bearware.FileTransferStatus;
 import dk.bearware.RemoteFile;
 import dk.bearware.TeamTalkBase;
+import dk.bearware.UserAccount;
+import dk.bearware.UserRight;
 import dk.bearware.backend.TeamTalkService;
 import dk.bearware.events.ClientEventListener;
 import dk.bearware.gui.AccessibilityAssistant;
```
hwangsihu, line 71: Is the import order appropriate? It's all mixed up.

Client/TeamTalkAndroid/src/main/java/dk/bearware/data/FileListAdapter.java
```diff
@@ -51,11 +53,17 @@
 import java.util.Map;
 import java.util.Vector;
 
+import dk.bearware.ClientFlag;
+import dk.bearware.ClientStatistics;
+import dk.bearware.ClientEvent;
```
hwangsihu, line 60: Several of these newly added imports appear to be unused (`ClientFlag`, `ClientStatistics`, `ClientEvent`, `User`, and also `Button` / `OnMenuItemClickListener` above). Could you please remove the unused imports and sort the remaining ones consistently? This may also be related to the current Android workflow failure.

Client/TeamTalkAndroid/src/main/java/dk/bearware/data/FileListAdapter.java
```diff
@@ -306,22 +251,50 @@ public void onClick(DialogInterface dialog, int whichButton) {
...
+                ttClient.getMyUserAccount(myuseraccount);
+                boolean downloadRight = (myuseraccount.uUserRights & UserRight.USERRIGHT_DOWNLOAD_FILES) != UserRight.USERRIGHT_NONE;
+                popup.getMenu().findItem(R.id.action_downloadfile).setEnabled(downloadRight).setVisible(downloadRight);
+                popup.setOnMenuItemClickListener(item -> {
+                    if (item.getItemId() == R.id.action_downloadfile) {
+                        downloadFile(remoteFile);
+                        return true;
+                    } else if (item.getItemId() == R.id.action_removefile) {
+                        removeFile(remoteFile);
+                        return true;
```
hwangsihu, line 280: Should this action also be hidden or disabled when the user does not have permission to remove/delete files? The download action checks `USERRIGHT_DOWNLOAD_FILES`, but `Remove file` appears to always be visible.

Client/TeamTalkAndroid/src/main/java/dk/bearware/data/FileListAdapter.java
```diff
@@ -306,22 +251,50 @@ public void onClick(DialogInterface dialog, int whichButton) {
...
             if((convertView == null) || (convertView.findViewById(R.id.progress) == null))
                 convertView = inflater.inflate(R.layout.item_file_transfer, parent, false);
             FileTransfer transferinfo = downloads.get(remoteFile.szFileName);
             ((TextView)convertView.findViewById(R.id.progress)).setText(context.getString(R.string.download_progress, getPercentage(transferinfo)));
-            Button cancelButton = convertView.findViewById(R.id.cancel_btn);
-            cancelButton.setOnClickListener(buttonClickListener);
-            cancelButton.setAccessibilityDelegate(accessibilityAssistant);
+            convertView.setOnClickListener(v -> cancelTransfer(remoteFile));
+
+            convertView.setOnLongClickListener(v -> {
```
hwangsihu, line 296: Is it intentional that a normal short tap immediately cancels the active transfer? This may be easy to trigger accidentally, especially for accessibility users. It might be safer to keep cancellation behind the popup action or add confirmation.

Client/TeamTalkAndroid/src/main/java/dk/bearware/data/FileListAdapter.java
```diff
@@ -332,6 +305,67 @@ public void onClick(DialogInterface dialog, int whichButton) {
...
+
+        alert.setNegativeButton(android.R.string.no, null);
+        alert.show();
+    }
+
+    private void cancelTransfer(RemoteFile remoteFile) {
+        FileTransfer transfer = downloads.get(remoteFile.szFileName);
+        if (ttClient.cancelFileTransfer(transfer.nTransferID)) {
+            downloadCancellationCleanup(transfer);
+            notifyDataSetChanged();
```
hwangsihu: Could you add a null check here? If `downloads.get(remoteFile.szFileName)` returns null while the list/transfer state is being refreshed, this would crash when accessing `transfer.nTransferID`.

hwangsihu (commented):
I left a few review comments. The main blocker seems to be the Android workflow failure, likely related to cleanup needed in this file.

hwangsihu:
Ah, the AI ​​left a comment. Anyway, it would be good to try to resolve what can be resolved.

amirmahdifard:
@hwangsihu About the imports, I just added them all, just incase later we add anything else and we forgot to add the imports, the file would not complane compilation error. About remove action, I want to do this, but sadly there is no user can delete files user right on the server, server allows file deletion if the user is channel operator, or the file owner, so like windows, I left that handling to the server. Anyway, the code supports this whenever a new user right  has been added to server, We can add checks.
about file canceling, It was as it is before, I just moved it to the file it self. Confurmation is not needed i think.

amirmahdifard:
I don't see the ai comment why?

amirmahdifard:
aaaaaaaaaaaaaa! fuck this android studio! I thoart button remove string is only for file remove, I didn't think it's also there for other things as well!

amirmahdifard:
@hwangsihu I added confirmation to canceling file downloads. The reason I didn't add this confirmation at the first place was because I didn't make any changes to the existing functions, I just moved them in to helper methods, and I used them in popups/clicks, I thort the desighn is going to be cool like telegram. Clicking a file starts download, clicking again cancels download. Holding on normal file shows download and remove file options, and holding on a downloading file shows cancel download option.
But currently, it behaves like this.
clicking on a file = starts downloading.
clicking again on a downloading file = ask confirmation question to cancel the download.
holding a file opens popup menu for download file, and remove file.
holding on a downloading file opens popup menu for cancel download option.
I love this way, And Finally my internet allowed me to turn my dreams in to real.

bear101:
Please create your PR as draft until it's ready for review. I.e. it can build and commit history has been cleaned up.

PR #3305, opened 2026-05-31
Remove OpenSSL on Windows Qt

hwangsihu:
Since Schannel is used on Windows, OpenSSL is not needed.


bear101 (commented):
Did you test that WebLogin is still working?

hwangsihu:
Yes, it's working properly.

bear101 (approved):

PR #3307, opened 2026-05-31
Change to sign the APK with the actual keystore

hwangsihu:
@bear101, could you please add the information about the keystore used for Google Play releases to your GitHub Secret? The required items are as follows:
- `ANDROID_KEYSTORE_BASE64` (You need to base64 encode your keystore)
- `ANDROID_KEYSTORE_PASSWORD`
- `ANDROID_KEY_ALIAS`
- `ANDROID_KEY_PASSWORD`

### DIFF
--- a/.github/workflows/android.yml
+++ b/.github/workflows/android.yml
@@ -59,15 +59,21 @@ jobs:
       env:
         ANDROID_NDK_HOME: ${{ steps.setup-ndk.outputs.ndk-path }}
 
-    - name: Generate Android signing keystore
+    - name: Set up Android signing keystore
       working-directory: ${{github.workspace}}/Client/TeamTalkAndroid
+      env:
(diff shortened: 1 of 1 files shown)


bear101:
I tried adding the secrets but it didn't seem to work. Also how do I prevent file `key.keystore` from being stolen? Or people echo'ing `ANDROID_KEY_PASSWORD`

hwangsihu:
Since Secret is encrypted with ***, I don't think you need to worry about that.

Nardol:
If I agree about `echo`, I won't really agree to say all is OK.
A member of the organisation which would make a bad PR could transform the key to send it somewhere else.
Adding the secret is only the beginning:
it should **not** be used in a PR, only in some precise scenarios, in a protected environment...
For example this kind of secret should be used in a workflow which should only be manually dispatched.
The [GitHub documentation](https://docs.github.com/en/actions/concepts/security/secrets) gives a lot of information and links.

bear101:
I don't see how this can be done safely without exposing the keystore properties.

PR #3311, opened 2026-06-02
Android client. Messages history fixes and redesighns

amirmahdifard:
1. Added popup menu actions for messages.
2. Replaced the old broken buggy single link support system with a new proper multiple links support system, Just like the desktop qt teamtalk's multiple links support for messages!


Client/TeamTalkAndroid/src/main/java/dk/bearware/data/TextMessageAdapter.java
```diff
@@ -175,6 +184,37 @@ public View getView(int position, View convertView, ViewGroup parent) {
...
                 msgdate.setTextColor(text_color);
                 msgtext.setTextColor(text_color);
+
+                convertView.setOnClickListener(v -> copyToClipboard(v.getContext(), txtmsg.szMessage));
+                convertView.setOnLongClickListener(v -> {
+                    PopupMenu popup = new PopupMenu(v.getContext(), v);
+
+                    List<String> urls = extractUrls(txtmsg.szMessage);
+                    for (int i = 0; i < urls.size(); i++) {
+                        popup.getMenu().add(0, 1000 + i, 0, urls.get(i));
```
bear101: Is this 1000 something that should a constant?

bear101: Or where does it come from?

bear101 (dismissed):

Client/TeamTalkAndroid/src/main/java/dk/bearware/data/TextMessageAdapter.java
amirmahdifard: It lists the links available in a message. See my changelog and pr discription. I made a helper to find all the links on a message and list them in the popup menu for users to click. it's multiple link support. If a message contane 5 links, it'll list all of those 5 links. Better than that old single link support system for messages. It's same as desktop clients. Install it on your phone and you'll see how cool it is

hwangsihu (dismissed):

amirmahdifard:
@hwangsihu thanks for this, I could also do this, but I mean, what's the problem with that 1000 thingy, for avoid it, we had to creat a extra more helper, which now you did.

hwangsihu (approved):

amirmahdifard:
@hwangsihu thanks for aprooving. Is merge button not available for you to merge my pull requests? because it looks that bear101 doesn't have any plan

hwangsihu:
Yes I unable to merge your PRs

bear101:
@hwangsihu did you test it?

amirmahdifard:
@bear101 if you care about my testings, I test them, and anyone who sees even this code and test it, will confurm that it's fine. I'm a programmer, a java programmer

amirmahdifard:
@hwangsihu why did you remove delete message action and added only clear history, That delete message action was there to delete only that single message. Add it back and also keep your clear action too. This one will clear all messages, while the other one deletes a single message

hwangsihu:
It is due to a functional mismatch. Since QtTeamTalk does not have a single message deletion feature, I do not want to create a functional difference. If that feature is needed, I think it would be better to add it later along with Qt.

amirmahdifard:
@hwangsihu That's not correct. Even corentin wanted to add single delete message option to qt teamtalk's context menu but he forgot to do it later. About seperet status message per server, I added it to android client first, and then corentin added to qt teamtalk. I added this delete message action because It's needed And I need it. Please add it back and keep both.

amirmahdifard:
matching 2 clients is ok, but it doesn't mean prevent an action to be existed in one client when it's not yet in the other client. I'm the one who is matching android client with qt client by all my pull requests, but it doesn't work like this. Add that action back, or do you want me to do it?

amirmahdifard:
@hwangsihu Thanks, your changes are correct. Now if you've also tested, tell him to merge my prs

bear101:
@hwangsihu please do the following:

- Fix the title and description of this PR
- Add notice in ChangeLog

PR #3312, opened 2026-06-03
Android client. Added the forgotan Subscription and transmission changes to the chat list

amirmahdifard:
When I was implementing these announcements in 5.18, because you took so long to review and merge my pull request, I forgot to complete it and add the chatlogs for them as whell like I did for the other things. But now has been fadded

### DIFF
--- a/Client/TeamTalkAndroid/src/main/java/dk/bearware/backend/TeamTalkService.java
+++ b/Client/TeamTalkAndroid/src/main/java/dk/bearware/backend/TeamTalkService.java
@@ -1050,6 +1050,13 @@ public class TeamTalkService extends Service
 
     @Override
     public void onCmdUserUpdate(User user) {
+
+        User olduser = users.get(user.nUserID);
+
+        if (olduser != null) {
+            Utils.subscriptionLogChanged(getBaseContext(), olduser, user) .ifPresent(text -> getChatLogTextMsgs().add( MyTextMessage.createLogMsg( MyTextMessage.MSGTYPE_LOG_INFO, text)));
+        }
+
         users.put(user.nUserID, user);
     }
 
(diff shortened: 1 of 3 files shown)


Client/TeamTalkAndroid/src/main/java/dk/bearware/backend/TeamTalkService.java
```diff
@@ -1175,6 +1182,13 @@ public void onCmdChannelNew(Channel channel) {
 
     @Override
     public void onCmdChannelUpdate(Channel channel) {
+
+        Channel oldchannel = channels.get(channel.nChannelID);
+
+        if (oldchannel != null) {
+            Utils.transmitUsersLogChanged(getBaseContext(), oldchannel, channel, getUsers()) .ifPresent(text -> getChatLogTextMsgs().add( MyTextMessage.createLogMsg( MyTextMessage.MSGTYPE_LOG_INFO, text)));
```
bear101: Why would you be interested in transmission changes in channels you're not participating in?

bear101 (dismissed):
These changes will overflow the chat history. I suggest they should be disabled by default.

Client/TeamTalkAndroid/src/main/java/dk/bearware/backend/TeamTalkService.java
amirmahdifard: no, It's only for Current channel. I followed the same patten I did for tts of this feature in 5.18 if you remember.

hwangsihu (dismissed):

amirmahdifard:
@hwangsihu I didn't see this statement in main activity I think that's why I forgot to add it at the first place. Thanks! this pr is now ready

hwangsihu (approved):

bear101:
@hwangsihu please do the following:

- Fix the title and description of this PR
- Add notice in ChangeLog

PR #3318, opened 2026-06-07
Update WebRTC to R7827 (Chromium 149)

hwangsihu:


bear101:
It takes many hours to clone WebRTC so we should only update it if there's new features that we need.

hwangsihu:
I don't think it's good to focus only on new features. Should we sacrifice compatibility and bug fixes? Also, existing users just need to change the branch to 7827 in WebRTC and sync.

bear101:
I suggest merging this PR into #3351

PR #3322, opened 2026-06-09
Fix broken .tt import on Android

hwangsihu:
Co-Authored-By: Claude Opus 4.8 <noreply@anthropic.com>

### DIFF
--- a/Client/TeamTalkAndroid/src/main/java/dk/bearware/gui/Utils.java
+++ b/Client/TeamTalkAndroid/src/main/java/dk/bearware/gui/Utils.java
@@ -517,13 +517,21 @@ public class Utils {
     
     public static DocumentBuilderFactory newSecureDocumentBuilderFactory() throws ParserConfigurationException {
         DocumentBuilderFactory dbf = DocumentBuilderFactory.newInstance();
-        dbf.setFeature("http://apache.org/xml/features/disallow-doctype-decl", true);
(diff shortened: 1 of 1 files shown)


amirmahdifard:
@hwangsihu @claude please just revurt the all java files changed in this commit https://github.com/BearWare/TeamTalk5/commit/3697813e9c2dc7a0cb20608d7e44335c53952696 because we don't know yet what else it has broke. before changes in that commit things that based on it has became broken in tt android So the proper fix would be revurting entire java files of that commit because now I don't know what else became broken during that change, I only figgered out about tt files import for now

hwangsihu:
First, please verify if the current code resolves the issue. Since the BearWare.DK login wasn't working either, it seems like this will fix that problem as well.

amirmahdifard:
@hwangsihu yes because the entire document factory has been modifyed, and I don't know what was the goal. i'll try it when i go home

amirmahdifard:
@hwangsihu stil it's broken. nothing is fixed, I just tryed it on my phone. And how would you even expect this pull request to fix the issues? it just added a try block around the faktory. Please don't completely trust ai. My request is to please revurt java files of that commit because onnistly I don't know what that commit is even trying to fix. It broke a feature in tt android that my friends used it a lot all this time when internet in iran was disconnected.
if you don't know java, please when you wanted to make any changes to teamtalk android, first ask a review from me, then do it. Thanks

PR #3323, opened 2026-06-11
Revert Android CodeQL changes breaking .tt import

hwangsihu:


amirmahdifard:
@hwangsihu yeah! I just tested this on my phone and I realised the following problems have been solved.
importing .tt files.
uploading files
web login activity
surprisingly, even the show official servers in server list and show unofficial servers in server list features!
You can now merge this.
desktop client is not my job, but that commit also modifyed a fue cpp files related to teamtalk sdk and qt teamtalk, so somehow make sure that something is not  broken in qt client like android client. Thanks.

hwangsihu:
Since my main device is an iPhone, I haven't been able to test Android very often, but I use Windows and ArchLinux frequently, so that is not the case. You don't need to worry about that. If any issues are found, I will revert or fix them.

amirmahdifard:
@hwangsihu ok. As I said, I don't involv myself to the other clients, i'm maintaning the android client only for now at least. So You can now merge this pr. It fixes everything. Thanks again.

bear101 (approved):

PR #3325, opened 2026-06-14
Upgrade to JUnit 6

hwangsihu:


CoBC:
Do we need every dependencies update? For some months now, it seems everytime a dependency is updated, we have a PR to update in TT. And, in majority of cases, we have to test, rebuild from scratch, sometimes fix errors on every build PC... Can't we focus on bug fixes and improvements, as before, and stop updating dependencies every days?

amirmahdifard:
actually it's right corentin. If you don't work on qt client, no body will work on it. It was 5.17 that You worked hevvyly on qt teamtalk. And qt teamtalk saw its huge update in the history. If you don't work on qt client and I don't work on android client and @hwangsihu doesn't add some features or some bug fixes, Teamtalk will have nothing to introduce in its next updates.

bear101:
> Do we need every dependencies update? For some months now, it seems everytime a dependency is updated, we have a PR to update in TT. And, in majority of cases, we have to test, rebuild from scratch, sometimes fix errors on every build PC... Can't we focus on bug fixes and improvements, as before, and stop updating dependencies every days?

I agree. The main focus should be on fixing issues (something that changes ChangeLog). We should only update dependencies if they contain bug fixes or address build issues.

hwangsihu:
In that case, is it okay to focus on bug fixes but occasionally update dependencies? Up-to-date dependencies are necessary for general reasons. However, leaving dependencies that are too old is also not good.

hwangsihu:
Also, is there a reason why you use version 5.9.3 instead of the latest 5.x version, 5.14.4, when upgrading to JUnit 5?

bear101:
> Also, is there a reason why you use version 5.9.3 instead of the latest 5.x version, 5.14.4, when upgrading to JUnit 5?

I tested with version 5.9.3 and found it to be working. If I remember correctly then it was the one provided by Ubuntu 24.

Ubuntu 24 and 26 do not come with JUnit 6 packages, so I don't see a reason to introduce JUnit 6 unless it comes with some new features that we need.

PR #3330, opened 2026-06-19
Android client changelogs in this branch

amirmahdifard:
from now on, to prevent conflicts, this branch exist to only update the changelog entries. Whatever Changes I make to teamtalk android, I will add its changelog entrie to this branch only and not to the original pr.


bear101:
ChangeLog is updated with the PR

PR #3340, opened 2026-06-23
Android client. Added Clear indication for private and channel messages sending rights

amirmahdifard:


bear101:
It is better to keep the channel-message-send-button so the user gets an error message when trying to send a message. By hiding the button no one knows why it's missing.

amirmahdifard:
@bear101 Sure! i can even just add a text that says message right disabled for you or you are not allowed to chat. Do you want me to do that? the reason this is needed, is that user doesn't start typing a message and then figger out that they are not allowed to chat.

hwangsihu:
> @bear101 Sure! i can even just add a text that says message right disabled for you or you are not allowed to chat. Do you want me to do that? the reason this is needed, is that user doesn't start typing a message and then figger out that they are not allowed to chat.

Good idea

amirmahdifard:
@hwangsihu Done!

bear101:
@amirmahdifard please clean up the commits. There should only be one commit.

`git rebase -i origin/master`

amirmahdifard:
@bear101 done. Ready to merge

bear101:
Can't you use existing edit fields (e.g. `channel_im_edittext`) for the text message "Sending channel messages disabled for you"? Then mark the text field as read-only.

The UI designer looks really bad with this new label, because the edit-field becomes tiny.

amirmahdifard:
@bear101 I Placed the text to be at the senter of the screen exactly in the place of edit box and edit box is completely gone removed from the leyowt, so only text is visible
About your question, no, this text view is better for screen readers and displaying error disallowed texts on android.

amirmahdifard:
This text is displayed on the senter at the bottom of the screen exactly in the place of edit box, and it's just an allturnet text that displayed when user doesn't have right. So it's better to keep it accessible and as it is.

amirmahdifard:
@bear101 this is a simple pr And I don't think it should be kept open for longer than this. Can you allow the merge so I can delete my branch? thanks

amirmahdifard:
@bear101 I've fixed the text view was a bit small. It's now as wide as the editbox itself. There is now no diffrence between them. Please, Test and look again. Thanks.

amirmahdifard:
@bear101 I have fixed this pr. Now it is good sise for both sited users and the screen reader users. Please merge it. Please.

amirmahdifard:
@bear101 i have fixed this for sited users visually grately. Please merge it bro it's open for a long time it is small change.

amirmahdifard:
@bear101 please also merge this pr and milestone it to 5.23. Thanks.

PR #3341, opened 2026-06-25
Android client. Added toggel transmissions for everyone for classroom channels

amirmahdifard:


hwangsihu (approved):

bear101:
@hwangsihu if you want to continue being code owner on TeamTalkAndroid then you need to ensure the quality of the code that comes in is good.

The PR has the following issues
- Commit message is misleading
  - It refers to "missing toggel actions". When did they go missing? And there's a typo.
- There is no updated ChangeLog and the PR is not linked to a milestone.
  - This mean no one knows about this feature so beta testers will not be informed

bear101:
Use "Labels" instead of "Android client" title prefix.

amirmahdifard:
@bear101 i'm sorry but only my commit messages have typo mistake. My code is ok

amirmahdifard:
@bear101 why you don't see the fakt that my code is clean and the app that finally user is going to use does not have any bug? i'm sorry only about my commit messages, i will also take care of them, but I feal that you don't see my code value

PR #3342, opened 2026-06-27
Android client: Fix channel listview items to be exposed as clickable items for accessibility services to work correctly from keyboard

beqabeqa473:
The problem is that doubletap was beeing propagated to a system and adapter onClick and onLongClick handlers were invoked, but these channel and user rows were not exposing clickable action, so accessibility services were not able to correctly act.


amirmahdifard:
@beqabeqa473 Thanks. I don't know if you noticed or not, but we had the same problem for files list where files items were read only items for talkback, But I fixed it when I was adding popup menu and direct clicking on files items them selves raather than buttons beside them haha

hwangsihu (approved):

poretsky (approved):

PR #3351, opened 2026-06-30
Update dependencies

hwangsihu:


CoBC:
Do we need to update all these dependencies? Aside from Prism, which has an open issue related to this dependency, what other bugs affecting TeamTalk are fixed by these updates?

bear101:
Prism should be updated in a separate PR if it contains important changes. Also Prism is not part of the toolchain.

@hwangsihu we should only update to official release. There no need to update libvpx and PortAudio to some git that may or may not have been tested.

bear101:
Please keep this PR then we can merge it when we need to rebuild the toolchain anyway.

hwangsihu:
@bear101, Would you mind submitting the patch currently in PortAudio as a PR to the upstream, or would you prefer to handle it yourself? I would like to get your approval since you wrote this patch. It will definitely be helpful if merged.

hwangsihu:
I plan to use tags for libvpx when a new version is released. However, this will not be the case for PortAudio. This is because it is too old, to the point where even some distributions have switched to the Git version. However, I will do the same for this when a new version becomes available.

bear101:
> @bear101, Would you mind submitting the patch currently in PortAudio as a PR to the upstream, or would you prefer to handle it yourself? I would like to get your approval since you wrote this patch. It will definitely be helpful if merged.

I've submitted it here: https://github.com/PortAudio/portaudio/pull/1160

Unfortunately it has an issue with msys2

bear101:
> @bear101, Would you mind submitting the patch currently in PortAudio as a PR to the upstream, or would you prefer to handle it yourself? I would like to get your approval since you wrote this patch. It will definitely be helpful if merged.

@hwangsihu do you have a setup where you can fix the msys2 build error? https://github.com/PortAudio/portaudio/pull/1160

hwangsihu:
I have MSYS2. Would you please give me repository permissions?

hwangsihu:
I've made the changes. Does the modification look okay? For now, there were no issues on my local MSYS2.

Also, I think we need to modify the code in a way that doesn't break the API. What would be the best way to do that?

hwangsihu:
By the way, I recommend switching from Cygwin to MSYS2. Since Cygwin is primarily for Linux emulation and MSYS2 is primarily for development, that would be appropriate.

PR #3355, opened 2026-07-03
Stop announcement burst when reconnecting after connection loss

beqabeqa473:


Client/TeamTalkAndroid/src/main/java/dk/bearware/backend/TeamTalkService.java
```diff
@@ -183,6 +183,8 @@ public class TeamTalkService extends Service
     CountDownTimer eventTimer;
     SparseArray<CmdComplete> activecmds = new SparseArray<>();
 
+    private boolean synchronizing = false;
```
bear101, line 186: Instead of introducing `synchronizing` it should be enough to check `activecmds.size() == 0` or `(activecmds.contains(CmdComplete.CMD_COMPLETE_JOIN) || activecmds.contains(CmdComplete.CMD_COMPLETE_LOGIN)) == false`

amirmahdifard: I agree

PR #3358, opened 2026-07-08
Fix navigation logic for physical phone's back button

amirmahdifard:
Previously, users could only navigate TeamTalk safely using the "Navigate Up" button. The phone's physical back button was not supported, causing users to be disconnected immediately and lose their session. This change handles the phone's back button to act the same as the navigation button. Refactored the code into a clean helper method for both to use, with clear comments indicating each step of the logic.

### DIFF
--- a/Client/TeamTalkAndroid/src/main/java/dk/bearware/gui/MainActivity.java
+++ b/Client/TeamTalkAndroid/src/main/java/dk/bearware/gui/MainActivity.java
@@ -90,6 +90,7 @@ import androidx.fragment.app.FragmentManager;
 import androidx.fragment.app.FragmentPagerAdapter;
 import androidx.fragment.app.ListFragment;
 import androidx.viewpager.widget.ViewPager;
+import androidx.activity.OnBackPressedCallback;
 
 import com.google.android.material.tabs.TabLayout;
 
@@ -310,6 +311,12 @@ extends AppCompatActivity
             mMediaPlayer.setOnCompletionListener(mediaPlayer -> mMediaPlayer.release());
             mMediaPlayer.start();
         }
+        getOnBackPressedDispatcher().addCallback(this, new OnBackPressedCallback(true) {
+            @Override
+            public void handleOnBackPressed() {
+                handleBackNavigation();
+            }
+        });
     }
 
     @Override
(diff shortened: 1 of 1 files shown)


amirmahdifard:
@bear101 Finally, I did my best to wright clean and explaning commit messages, clean comments, with no spelling mistakes😊
do you like this way?

bear101:
I only have the emulator to test Android. When i press Back button then it still returns immediately to the server list.

amirmahdifard:
@bear101 what's your android version? may be I need to use the onBackPressed dispachor anyway instead of the discontinued onBackPressed method.

amirmahdifard:
I will push a new commit, please tell me if that fixes it or not.

amirmahdifard:
@bear101 would you please try it one more time?
I have updated the navigation logic to use OnBackPressedDispatcher. This should fix the issue on modern emulators (Android 13+) where the legacy onBackPressed() method is often bypassed.
I was awhere of this, but I didn't use this new callback because I didn't think we will need it. But this is the new way introduced by google to replace the old one for apps who wants to controle the back button.
Please let me know if it works for you now. Thanks!

amirmahdifard:
@bear101 please when you tested this again, also tell me is the shrinking exit animation for android works fine with this? because I also want to handle android's shrinking exit animation

amirmahdifard:
@bear101 bro, remember in 2024 you promiced me that you will give me a teamtalk sdk lisence soon?
will you give it to me now please? thanks!

Nardol:
Well, now we all know exactly why you’re so eager to get your contributions merged into the project 🤣

amirmahdifard:
@Nardol no today my friend asked me to run a pro server, and it's not possible to run it without lisence. One of my friends gave me a tt sdk lisence But i don't like to use it. And it's not linked to my account anyway. Also, it's not so much important, I just remembered and asked.
For my contributions value, mentioning my name in updates news leter makes me more happy than getting a lisence.

bear101:
Is there an issue or discussion linked to this PR?

amirmahdifard:
@bear101 what should I say to you? This is a bug, And I have fixed the bug! what is your problem right now?

amirmahdifard:
@bear101 i'm receaving many complanes about this bug. People tell me that the action bar navigate up button is a bit hard to reach, and the phone's back button is not supported, so if user is downloading a large file or uploading, if their finger hits their phone's back button, everything is lost and user got out of the server and throw back in to the server list.
Back button is supported in every app, before I forgot to support it.
In teamtalk, people expect to navigate parent channels with back button, and at the end, if everything is safe, they would exit the server safely. This makes it so much comftable to move users between channels, or join other channels.
Everyone is not used to navigate up button at the top, moste of the people use their back button for these things.
This is incomplete feature that causes bug. Because in android, if navigate up button is doing something in the app, the back button should also do the same.
So many people are saying "oops, my finger hit the back button and I kicked out from the server"
Please merge. Thank you!
Did you test the last commit on your device? does the back button works for your emulator now?

amirmahdifard:
@bear101 Did you test the last commit on your device? does the back button works for your emulator now?

bear101:
@amirmahdifard when I press the emulator's back button then it return to the server list instead of the parent channel. The idea is to make the back button work like the app's top-left button, right?

amirmahdifard:
@bear101 yes. Did you try the laytist commit? because it works here correctly.

bear101:
I used commit 04697971a

amirmahdifard:
@bear101 It works here correctly. I will try on more phones and see if there are any problem

bear101 (approved):

PR #3364, opened 2026-07-13
Added the MOTD table column to qt client

amirmahdifard:
Previously, the motd columb was missing, And we could not move to the motd column by right arrow key and move over the motd column. But now I added it and tested it. It's great

### DIFF
--- a/Client/qtTeamTalk/serverlistdlg.cpp
+++ b/Client/qtTeamTalk/serverlistdlg.cpp
@@ -41,6 +41,7 @@ enum
     COLUMN_INDEX_SERVERNAME,
     COLUMN_INDEX_USERCOUNT,
     COLUMN_INDEX_COUNTRY,
+    COLUMN_INDEX_MOTD,
     COLUMN_COUNT,
 };
 
@@ -93,6 +94,7 @@ QVariant ServerListModel::headerData(int section, Qt::Orientation orientation, i
             case COLUMN_INDEX_SERVERNAME: return tr("Name");
             case COLUMN_INDEX_USERCOUNT: return tr("Users");
             case COLUMN_INDEX_COUNTRY: return tr("Country");
+            case COLUMN_INDEX_MOTD: return tr("MOTD");
         }
     }
     return QVariant();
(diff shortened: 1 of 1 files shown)


CoBC:
I'm not convinced of the usefulness of this addition. In my opinion, the MOTD should only be accessible to users actually connected to the server. It serves as a welcome or informational message; its display in the server list is, in my view, unnecessary.

amirmahdifard:
I didn't add this feature, motd is available in server list in all clients for public servers allready, but it was only accessible for screen readers from accessible text role in the name column. The motd column it self was missing, I just added the column to complete it and make it easy to see and navigate motd for screen reader users like other table view columns inclooding myself. 😊
Also, Actually I see this feature useful Because some people may be comftable to see the server's information and rules that the owner provides before entering the server, same as number of users and cuntry. And I think that's why This feature is added. at the first And I really like it.

computertech1012:
Hi,
I have to agree with @CoBC here. Also, the "Stats" bot connecting and disconnecting from a server is very noisy and anoying just to make your server public.
Also, having to add something like #teamtalkskip# makes the server name look really ugly and nasty just to get rid of the stats bot. I am only putting this here because only servers that are in the official and unaficial lists show the MOTD, not local servers.
Thanks for reading!

amirmahdifard:
@computertech1012 hi. The stats bot is not related to this pr. The stats bot is handled entirely by @bear101 The stats bot gives server name, users numbers, cuntry and motd value for all public servers and clients retreave and shows them to the user. This was already shown to qt client users, I just fixed the accessibility by making the tabel view look the same as other tabel views So We can navigate to this column.
And about stats bot connecting to server constantly, it is because the stats bot needs to make sure your public server is available stil, and to update your server's laytist information such as name, number of users, cuntry and motd. This is very nice because In client, you can filter servers by name, cuntry users etc.
For example, if you want to join very full servers, you can filter the client to only show you the servers that has more than 15 users count. Or you can filter to show you only the servers that are hosted in iran. I love this And using it so much when exploring and navigating public server.
@bear101 did a good job by building this system of public servers.
And if anyone doesn't likes them, they can disable inclood public servers from server list.

amirmahdifard:
Anyway, I have just found a better idea

PR #3365, opened 2026-07-13
fixed teamtalk crash when you opened context menu on empty server list

amirmahdifard:
when the server list is completely empty and you have no selection and you open the context menu on it, causes crash. This is because in 5.23, someone incorrectly handled this that the code would grab the nul.
But in 5.22 and older, the validations were correctly seperated, so the code didn't crash, And we would stil have the possible options.
protected the crash cause behind safe gard, like the same stile as the other views.
No Crash: If the list is empty, the && stops the code from looking for [srcIndex.row()].

### DIFF
--- a/Client/qtTeamTalk/serverlistdlg.cpp
+++ b/Client/qtTeamTalk/serverlistdlg.cpp
@@ -971,8 +971,6 @@ void ServerListDlg::slotTreeContextMenu(const QPoint& /*point*/)
     const QString country = "country";
     sortCountry->setChecked((ttSettings->value(SETTINGS_DISPLAY_SERVERLIST_SORT, SETTINGS_DISPLAY_SERVERLIST_SORT_DEFAULT).toString() == country)?true:false);
     sortMenu->addAction(sortCountry);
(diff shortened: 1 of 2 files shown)


Client/qtTeamTalk/serverlistdlg.cpp
```diff
@@ -981,12 +979,14 @@ void ServerListDlg::slotTreeContextMenu(const QPoint& /*point*/)
     QAction* publishServ = menu.addAction(tr("&Publish Publicly"));
     QAction* genTTServ = menu.addAction(tr("&Generate .tt file"));
     QAction* publishJoinCode = menu.addAction(tr("Generate &Join Code"));
+    auto srcIndex = m_proxyModel->mapToSource(ui.serverTableView->currentIndex());
```
bear101, line 982: If there's no selection then I would assume `ui.serverTableView->currentIndex()` returns an invalid `QModelIndex`. Passing an invalid model index to m_proxyModel->mapToSource() returns what? Probably also an invalid model index. So `m_model->getServers()[srcIndex.row()]` is called on an empty list.

bear101 (changes requested):

bear101:
Superceeded by #3390

PR #3371, opened 2026-07-16
Ability to filter servers by country in qt teamtalk, Same as filtering servers by name!

### LINKED ISSUE #2018: Qt client: also allow us to filter servers with cuntry
### Description

since public servers have been removed and the unofficial servers are now the main servers and the list is also long, it's good that we can type a name and filter servers by name or filter them with specific amount of users, but also please add an edit box to type cuntry, to filter servers with only that cuntry specificly: i'll use my self as an example.
i'm a persian user, so i want to only see servers wich are hosted in iran: so in the cuntry edit box i type ir, and then it will show me the servers with ir cuntry only, as long as i don't delete the ir from the cuntry edit box.
Qt client: 5.14

### Application

- [X] qtTeamTalk
- [ ] TeamTalkAndroid
- [ ] iTeamTalk
- [ ] TeamTalkServer

amirmahdifard:
In qt teamtalk, it is possible to filter servers by name, So only the servers that has the words you typed will be shown to you.
Added the same useful feature to country! Now, I can filter my servers list and See only the servers hosted in iran with "ir"
fixes https://github.com/BearWare/TeamTalk5/issues/2018
Coded exactly as the same stile as the original existing code
Fixed accessibility for motd column


ziyadmohamed:
it's nice

PR #3372, opened 2026-07-17
Fixed the MOTD column accessibility

amirmahdifard:
in qt teamtalk tables, we can move over the columns with left and right arrow keys to access them better. Only the motd column was broken for this, Fixed it.

### DIFF
--- a/Client/qtTeamTalk/serverlistdlg.cpp
+++ b/Client/qtTeamTalk/serverlistdlg.cpp
@@ -41,6 +41,7 @@ enum
     COLUMN_INDEX_SERVERNAME,
     COLUMN_INDEX_USERCOUNT,
     COLUMN_INDEX_COUNTRY,
+    COLUMN_INDEX_MOTD,
     COLUMN_COUNT,
 };
 
@@ -93,6 +94,7 @@ QVariant ServerListModel::headerData(int section, Qt::Orientation orientation, i
             case COLUMN_INDEX_SERVERNAME: return tr("Name");
             case COLUMN_INDEX_USERCOUNT: return tr("Users");
             case COLUMN_INDEX_COUNTRY: return tr("Country");
+            case COLUMN_INDEX_MOTD: return tr("MOTD");
         }
     }
     return QVariant();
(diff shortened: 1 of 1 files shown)


bear101:
MOTD are typically multi line messages and therefore does not work well with a list widget. The row sizes will be uneven

amirmahdifard:
@bear101 Before this, motd was only announced on the accessible text role on the first column like all other tables, Just the column was missing for screen reader users.
In other places, we have everything (note,), etc, exposed as table columns.
Also, I tryed this version with even multy line motds, it doesn't cause any unexpected bug and it works as expected.

amirmahdifard:
@bear101 please first give a try to this. It's ok for sited users this is just an accessibility improvement like other tables

bear101:
Who has requested this feature?

amirmahdifard:
tabel views are useful in this case that screen readers can see each information in its own column. Country in the country column, name in the name column, Note in the note column, file size in the file column. And it works everywhere, Just this one column was missing. Is it really that hard for you to understand?

tohidbareza:
hello
@bear101 
I want to explane how this is useful for screen readers and blind users.
Visual users, when looking at screen, they can see everything at a glance
These table columns allows blind users to have the same ability.
We can press left and right arrow keys on the keyboard to move between columns and read only that column, without reading the other information.
e.g. If I want to read the motd or country quickly, I can press right arrow key to go to the column that I wanted and read it. This feature introduced from teamtalk v5.17, And I very like the accessibility. Everything is not long. Every information is in its own column.
So, If you allow this improvement to land on the new releace, it would be great. Thanks, I hope I have explaned it wel.

amirmahdifard:
@bear101 As explaned in the above comments, this pr gives the same feel to blind and screen reader users that sited users have, + it's available everywhere else. It's minimal. Can you please merge it so it lands to 5.23? Thanks a lot!

amirmahdifard:
@bear101 please also merge this. This tested by all my friends and it is completely fine. Thanks.

bear101 (approved):
When users initially starts the server list dialog then the headers will appear broken because `restoreGeometry()` does not match the new columns. I guess this is just something we have to accept..

amirmahdifard:
@bear101 You approved this. Please also merge it to master so I can delete the branch.
Grate. Now columns in every screen look the same for screen readers. Files, accounts, online users, bans, etc.

PR #3374, opened 2026-07-17
Ability to filter servers by country

### LINKED ISSUE #2018: Qt client: also allow us to filter servers with cuntry

amirmahdifard:
In qt teamtalk, it is possible to filter servers by name, So only the servers that has the words you typed will be shown to you.
Added the same useful feature to country! so Now, I can filter my servers list and See only the public servers hosted in iran with "ir"
fixes https://github.com/BearWare/TeamTalk5/issues/2018
Coded exactly as the same stile as the original existing code
exclooded local servers just like the user count filtering.


amirmahdifard:
@bear101 can I ask when will you view and merge my prs? this should've been merged and appear under beta 202. but you haven't merged it yet.

bear101:
Who has requested this feature? Anyway, if the feature should be there then make a drop-down where the active countries are listed. Then blank means all.

amirmahdifard:
@bear101 That's what I did. I Added a nice filtering. Ok, it looks you don't want to merge any of my prs for some reason.

amirmahdifard:
@bear101 if you always refuse to merge prs, then what you have to introduce in next updates of teamtalk?
if you are not working yourself on this project, don't also disallow others from working on it.
You are not doing anything yourself, And if anyone is also doing something valuable, you are refusing to merge it.
I hope something to happen. I'm really tired! It is not a good fealing at all to code and then see your code is being lost for no reason!

amirmahdifard:
@bear101 why you closed this? This feature is wel implemented and it works completely fine

PR #3375, opened 2026-07-19
keyboard navigation accessibility fix. Stop using Qt's Fusion style for dark Color

amirmahdifard:
I was fasing an accessibility navigation bug for months, And I couldn't find the reason.
Last night, I stayed awake the entire night and made and debugged a sample qt test application, and finally I got the bug cause.
when Qt's Fusion style is active and being used, every disabled menu items in menus/context menus ( disabled grade out QActions) would not receave focus and skipped for screen readers during keyboard navigation with up and down arrow keys.
Currently, If We set windows Color to light from windows settings and open teamtalk, this bug will be solved and every item would receave focus properly during keyboard navigation
when windows Color is set to dark in windows settings, Qt's Fusion style will be activated for teamtalk, And cause this accessibility bug.
So, Fixed it by stop using Qt's Fusion style on dark Color
This bug is for Qt's Fusion style, and I've reported it to Qt, so they will fix it, so that layter Qt's Fusion style won't cause this annoying bug anymore, So that anyone doesn't have to try to hunt the bug for months.

### DIFF
--- a/Client/qtTeamTalk/main.cpp
+++ b/Client/qtTeamTalk/main.cpp
@@ -25,7 +25,6 @@
 #include <QFileOpenEvent>
 #include <QPalette>
 #include <QSettings>
-#include <QStyleFactory>
 #include <QUrl>
 #include <QtPlugin>
 
@@ -51,7 +50,6 @@ public:
         QSettings settings("HKEY_CURRENT_USER\\Software\\Microsoft\\Windows\\CurrentVersion\\Themes\\Personalize", QSettings::NativeFormat);
         if (settings.value("AppsUseLightTheme", -1).toUInt() == 0)
         {
-            setStyle(QStyleFactory::create("Fusion"));
             QPalette darkPalette;
             QColor darkColor = QColor(45, 45, 45);
             QColor disabledColor = QColor(127, 127, 127);


bear101:
@CoBC are you familiar with this issue? I've never noticed any problems with dark mode.

amirmahdifard:
@bear101 Do you have screen reader? if yes, I will tell you how to reproduce it. The problem is not due to dark mode, it's about qt's fution style being active. I also tested this in other qt test sample

CoBC:
> @CoBC are you familiar with this issue? I've never noticed any problems with dark mode.

I'm not aware of this issue, nobody has talked about it to me

amirmahdifard:
@CoBC remember some time ago I told you teamtalk has a problem for me that in context menus, when I press arrow keys, my keyboard skips grade out menu items, and you told me why this is happening? I spent one night, made a qt test sample application, And after long tests, I figgered out that when qt's fution style is active, it causes this bug.
Steps to reproduce.
1: First, go to your windows settings, in culler, set your windows culler to dark.
2: open your current teamtalk.
3: open a context menu that has some disabled options inside it.
4: press your up and down arrow keys, and you will see that it only focuses on the enabled options. It skips and jumps over the disabled options during keyboard navigation, even though they are visible in the screen and for sited users.
5: now close teamtalk, go to windows settings, set windows culler to light.
6: now open teamtalk again, open a context menu, and press arrow keys.
You will see that every item is focused correctly, and just like how sited users can see, keyboard can also reach disabled items, and nvda announces disabled for them.
This issue only happens when windows culler is set to dark because teamtalk was configgered to activate qt's fution style for dark systems, And this pr fixes that issue by not using fusion style for dark culler.
So with this pr, it doesn't matter if your windows is dark or light, it will no longer cause this bug for teamtalk.
I have widely tested and confurmed it.
@bear101 you can also follow the reproduce steps I provided. I think you don't even need screen reader to reproduce, you can just press your arrow keys and see where your keyboard focus lands, And how it skips the grade out options and jumps to the next enabled item. Just try the above steps in light mode and dark mode, and in both cases, watch your keeboard focus pointer on the screen. Thanks.

bear101:
Thanks for the description on how to reproduce the issue. It makes it much easier to understand what issue you're trying to fix.

However, this change makes menus and dialog boxes unreadable to sighted users. Text and background color becomes the same in dark mode.

CoBC:
I'm not sure it's really a problem to not focus disabled items. If an option is disabled, why we should focus it?

amirmahdifard:
@bear101 oh! did  you try the build of this pr on a dark system? i'm blind, So I don't know how it looks like for sited users. But I didn't remove any other dark settings, I just removed the Qt fusion style from dark, to use the windows default style for dark and light both. Is it ok?
@CoBC The disabled items are focusable and visible for both sited users and everywhere, But they are not clickable, and NVDA says disabled for them for the blind user to understand that the item is disabled, So this expectation matches how sited users feal. They are able to see those items and they are awhere that it exist, but it's disabled, So with this fix, the blind users will feal the same as sited users for dark culler systems

bear101:
@amirmahdifard, yes, on a Dark system. Qt Fusion is only activated in dark mode.

amirmahdifard:
@bear101 can't we somehow give the dark systems the same look but without qt's fusion style, fusion style is causing accessibility problems and i have reported it to qt, so my suggestion is merge this pr, and whenever qt fixed the fusion style, i will add the fusion style back to teamtalk like before. This is a temporary workaround. So will you merge it please?

amirmahdifard:
@bear101 do you acsept this as a temporary workaround until qt fixes the fusion style on there end?

bear101:
> @bear101 do you acsept this as a temporary workaround until qt fixes the fusion style on there end?

No, it's completely unusable to sighted users.

amirmahdifard:
@bear101 ok. So if you have any other suggestion, tell me and I will try to implement it. If not, close this pr and i will delete the branch. Actually, i'm wundering how else we can avoid fusion style without effecting sited users. I'm not sited i'm completely blind, so I have no idea.

PR #3392, opened 2026-08-15
Reject 4096-block desktop sessions

### LINKED ISSUE #3386: TeamTalk Server hangs on 4096-block desktop frame due to off-by-one limit
### Description

TeamTalk Server hangs when a client repeatedly shares a 3264x1280 desktop using `BMP_RGB32` and `DESKTOPPROTOCOL_ZLIB_1`. Smaller desktop frames, such as 160x90 and 320x180, work normally.

### Application

- [ ] qtTeamTalk
- [ ] TeamTalkAndroid
- [ ] iTeamTalk
- [ ] TeamTalkClassic
- [x] TeamTalkServer

### Platform

- [ ] Windows
- [ ] macOS
- [ ] Android
- [ ] iOS
- [x] Linux

### TeamTalk version

5.22.0.5198

### Steps to reproduce

1. Start TeamTalk Server on Linux.
2. Log in with an account that can transmit desktop sessions and join a channel with a receiving client.
3. Share a 3264x1280 desktop using `BMP_RGB32` and `DESKTOPPROTOCOL_ZLIB_1`.
4. Continue sending normal desktop updates.

### Expected behavior

The desktop is delivered to receiving clients, or the server rejects the unsupported desktop size without affecting other users.

### Actual behavior

The server repeatedly logs:

```
Failed assertion m_blocks.size() == (size_t)GetBlocksCount() in DesktopCache.cpp at line 342
```

A receiver may see the same frame delivered repeatedly.

Removing the desktop-transmit right prevents this because the desktop session is rejected before reaching this code path.

### Update: measurements on an isolated server

I went back and tested this properly on a stock 5.22.0.5198 Linux server with two SDK clients, rather than relying on what I saw the first time.

The assertion is completely deterministic. Every large frame produces exactly one:

| Desktop | Blocks | Frames sent | Assertions |
| --- | --- | --- | --- |
| 3213x1300 | 63 x 65 = 4095 | 30 | 0 |
| 3264x1280 | 64 x 64 = 4096 | 30 | 30 |
| 3213x1300 | 4095 | 30 | 0 |
| 3264x1280 | 4096 | 30 | 30 |

I alternated those four times against the same running server, and separately sent 250 frames at 4096 blocks and got 250 assertions.

It also depends on the frame contents, which I think explains why this has been hard to see. Keeping the same 4096-block size and changing only what is in the buffer:

| Frame contents | Assertions in 30 frames |
| --- | --- |
| Uniform bytes, as in a zeroed buffer | 0 |
| Varying pattern | 30 |

A uniform bitmap makes every duplicate set contiguous, so `IsBlockRange()` sends it through the range branch, which never writes the `0xFFF` terminator. The ambiguity only exists on the single-list branch.

### Correction on the hang

My original report said the server stops accepting connections and that a service stop can time out. I could not reproduce that in this round of testing. Across more than 280 large frames the server kept logging, kept accepting connections, and RSS moved only from 406 MB to 413 MB. I also checked every server log I still have on two hosts going back to March and there is no `DesktopCache` assertion in any of them other than the ones I generated during this testing.

So I should not have stated the hang as a consequence of this bug. What I can support is the cache corruption and the assertion, on every frame, from a desktop size the server currently accepts as valid.

#3395 describes an infinite loop in `ConvertFromUInt12Array` that traps the reactor thread while it holds the `ServerNode` mutex, which locks out TCP commands. That matches the behaviour I originally described here much better than this issue does, and I now think that is the more likely explanation for what I saw. The two are separate: that one needs a malformed field length, and I checked that `ConvertToUInt12Array` can only ever emit lengths that are 0 or 2 more than a multiple of 3, so this bug cannot produce the input that one needs.

Leaving this issue open for the block-number collision itself, which is reproducible on demand.

### Technical details

RGB32 desktop blocks are 51x20 pixels. A 3264x1280 desktop therefore produces exactly 64x64, or 4096, blocks.

PR #2884 added the large-desktop limit. Its test and API documentation use a maximum of 4095 blocks, but `DesktopSession::IsValid()` currently accepts:

```cpp
GetBlocksCount() <= DesktopPacket::BLOCKNUMS_MAX
```

`BLOCKNUMS_MAX` is `0x1000`, so 4096 blocks are still accepted. The highest block index is then `0xFFF`, which is also used as the delimiter in the single duplicate-block encoding in `PacketLayout.cpp`. This can leave the desktop cache with fewer blocks than `GetBlocksCount()` and trigger the assertion in `DesktopCache.cpp`.

The likely correction is to enforce the documented 4095-block maximum and reject an incomplete initial desktop cache before making it current. Regression tests around block indexes `0xFFE` and `0xFFF` would also cover this boundary.

PR #2893 appears to address a separate desktop payload-size/OOM problem.

computertech1012:
Hi,

This fixes #3386 by rejecting desktop sessions that require 4096 blocks.

Desktop block numbers are stored as 12-bit values, but `0xFFF` is also used as the end marker for a single duplicate-block list. This makes `0xFFE` the highest usable block number. A 3264x1280 RGB32 desktop currently creates 4096 blocks, so its final block uses the reserved `0xFFF` value.

When that value is interpreted as the duplicate-list marker, the server can finish receiving the update with fewer cached blocks than the desktop dimensions require. This leads to the repeated `m_blocks.size() == GetBlocksCount()` assertion described in the issue.

The change sets `DesktopPacket::BLOCKNUMS_MAX` to `4095` as an exclusive limit. Valid block numbers remain `0` through `4094`, and sessions requiring 4096 blocks are rejected. It also adds a Catch2 regression test for all four bitmap formats, confirming that 4095 blocks remain valid and 4096 blocks are rejected.

@bear101, could you please review this when you have time?

Thanks for reading!

### DIFF
--- a/Library/TeamTalkLib/teamtalk/DesktopSession.cpp
+++ b/Library/TeamTalkLib/teamtalk/DesktopSession.cpp
@@ -56,8 +56,7 @@ DesktopSession::DesktopSession(const DesktopWindow& wnd, int bytes_per_line)
 
 bool DesktopSession::IsValid() const
 {
-    // 0xFFF terminates duplicate-block lists and cannot be used as a block index.
-    return GetBlocksCount() > 0 && GetBlocksCount() < DesktopPacket::BLOCKNUMS_MAX &&
+    return GetBlocksCount() > 0 && GetBlocksCount() - 1 <= DesktopPacket::BLOCKNUMS_MAX &&
            GetWidth() <= DesktopPacket::WIDTH_MAX && GetHeight() <= DesktopPacket::HEIGHT_MAX;
 }
 
--- a/Library/TeamTalkLib/teamtalk/DesktopSession.cpp
+++ b/Library/TeamTalkLib/teamtalk/DesktopSession.cpp
@@ -56,7 +56,8 @@ DesktopSession::DesktopSession(const DesktopWindow& wnd, int bytes_per_line)
 
 bool DesktopSession::IsValid() const
 {
-    return GetBlocksCount() > 0 && GetBlocksCount() <= DesktopPacket::BLOCKNUMS_MAX &&
+    // 0xFFF terminates duplicate-block lists and cannot be used as a block index.
+    return GetBlocksCount() > 0 && GetBlocksCount() < DesktopPacket::BLOCKNUMS_MAX &&
            GetWidth() <= DesktopPacket::WIDTH_MAX && GetHeight() <= DesktopPacket::HEIGHT_MAX;
 }
 
--- a/Library/TeamTalkLib/teamtalk/DesktopSession.cpp
+++ b/Library/TeamTalkLib/teamtalk/DesktopSession.cpp
@@ -56,7 +56,7 @@ DesktopSession::DesktopSession(const DesktopWindow& wnd, int bytes_per_line)
 
 bool DesktopSession::IsValid() const
 {
-    return GetBlocksCount() > 0 && GetBlocksCount() - 1 <= DesktopPacket::BLOCKNUMS_MAX &&
+    return GetBlocksCount() > 0 && GetBlocksCount() <= DesktopPacket::BLOCKNUMS_MAX &&
            GetWidth() <= DesktopPacket::WIDTH_MAX && GetHeight() <= DesktopPacket::HEIGHT_MAX;
 }
 
--- a/Library/TeamTalkLib/test/CatchDefault.cpp
+++ b/Library/TeamTalkLib/test/CatchDefault.cpp
@@ -36,6 +36,7 @@
 #include "settings/Settings.h"
 #include "teamtalk/Commands.h"
 #include "teamtalk/Common.h"
+#include "teamtalk/DesktopSession.h"
 #include "teamtalk/StreamHandler.h"
 #include "teamtalk/client/AudioMuxer.h"
 #include "teamtalk/client/Client.h"
@@ -126,6 +127,29 @@ TEST_CASE( "Init TT", "" )
     TTInstance* ttinst = nullptr;
     REQUIRE( (ttinst = TT_InitTeamTalkPoll()) );
     REQUIRE( TT_CloseTeamTalk(ttinst) );
+}
+
+TEST_CASE("DesktopSession block limit")
+{
+    auto verify_block_limit = [](teamtalk::RGBMode rgb_mode, int block_width, int block_height)
+    {
+        auto const max_session = teamtalk::MakeDesktopSession(block_width * 63,
+                                                              block_height * 65,
+                                                              rgb_mode);
+        REQUIRE(max_session.GetBlocksCount() == 4095);
+        REQUIRE(max_session.IsValid());
+
+        auto const oversized_session = teamtalk::MakeDesktopSession(block_width * 64,
+                                                                    block_height * 64,
+                                                                    rgb_mode);
+        REQUIRE(oversized_session.GetBlocksCount() == 4096);
+        REQUIRE_FALSE(oversized_session.IsValid());
+    };
+
+    verify_block_limit(teamtalk::BMP_RGB8_PALETTE, RGB8_BLOCK_PIXEL_W, RGB8_BLOCK_PIXEL_H);
+    verify_block_limit(teamtalk::BMP_RGB16_555, RGB16_BLOCK_PIXEL_W, RGB16_BLOCK_PIXEL_H);
+    verify_block_limit(teamtalk::BMP_RGB24, RGB24_BLOCK_PIXEL_W, RGB24_BLOCK_PIXEL_H);
+    verify_block_limit(teamtalk::BMP_RGB32, RGB32_BLOCK_PIXEL_W, RGB32_BLOCK_PIXEL_H);
 }
 
 #if defined(ENABLE_OGG) && defined(ENABLE_SPEEX)
--- a/Library/TeamTalkLib/test/CatchDefault.cpp
+++ b/Library/TeamTalkLib/test/CatchDefault.cpp
@@ -37,6 +37,7 @@
 #include "teamtalk/Commands.h"
 #include "teamtalk/Common.h"
 #include "teamtalk/DesktopSession.h"
+#include "teamtalk/PacketLayout.h"
 #include "teamtalk/StreamHandler.h"
 #include "teamtalk/client/AudioMuxer.h"
 #include "teamtalk/client/Client.h"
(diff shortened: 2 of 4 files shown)


Library/TeamTalkLib/teamtalk/DesktopSession.cpp
```diff
@@ -56,7 +56,8 @@ DesktopSession::DesktopSession(const DesktopWindow& wnd, int bytes_per_line)
 
 bool DesktopSession::IsValid() const
 {
-    return GetBlocksCount() > 0 && GetBlocksCount() <= DesktopPacket::BLOCKNUMS_MAX &&
+    // 0xFFF terminates duplicate-block lists and cannot be used as a block index.
+    return GetBlocksCount() > 0 && GetBlocksCount() < DesktopPacket::BLOCKNUMS_MAX &&
```
bear101: If this is the case then DesktopPacket::BLOCKNUMS_MAX should be 4094 instead of 4095

computertech1012: Hi,

I've corrected `DesktopPacket::BLOCKNUMS_MAX` to 4094 and updated the related comparisons.

Thanks for reading!

bear101 (changes requested):
Hm, valid blockno is 0..4094, so `BLOCKNUMS_MAX` should be 4095.

Then remove `GetBlocksCount() - 1` and change `blockno <= BLOCKNUMS_MAX` to `blockno < BLOCKNUMS_MAX`

computertech1012:
Hi,

@bear101, I've updated this as suggested. `BLOCKNUMS_MAX` is now 4095, `GetBlocksCount()` no longer subtracts one, and the block-number checks now use `< BLOCKNUMS_MAX`.

Thanks for reading!

Library/TeamTalkLib/test/CatchDefault.cpp
```diff
@@ -128,6 +129,29 @@ TEST_CASE( "Init TT", "" )
     REQUIRE( TT_CloseTeamTalk(ttinst) );
 }
 
+TEST_CASE("DesktopSession block limit")
```
bear101, line 198: This unit test doesn't prove that there is an issue with 4096 blocks. In commit 18de30159c8ab073e02a60b6ecd44a2ced24d4eb I've tried to make a unit test that creates a desktop window which results in 4096 blocks but it does not fail despite all video blocks are duplicates

computertech1012: Hi,

@bear101, you're right that my current unit test verifies the block-count boundary but does not reproduce the reported hang.

I reviewed `18de301`. Because its framebuffer is zero-filled, the duplicate blocks form a continuous range and are encoded through `FIELDTYPE_BLOCK_DUP_RANGE`. That does not exercise `FIELDTYPE_BLOCK_DUP`, where `0xFFF` is both the delimiter and a possible block number.

I think a more targeted integration test would make block 4095 duplicate a non-adjacent block while the intervening blocks differ, forcing the single-duplicate encoding path. I can revise the regression test that way. Does that match the failure path you would like tested?

Thanks for reading!

computertech1012:
Hi,

@bear101, I set up an isolated Linux server and got a reliable reproduction, so I wanted to share the numbers in case they help on your end.

### Setup

Stock `tt5srv` 5.22.0.5198 (compiled Apr 5 2026) on Debian, two SDK clients in the same channel, sender using an account with `USERRIGHT_TRANSMIT_DESKTOP`, `BMP_RGB32` and `DESKTOPPROTOCOL_ZLIB_1`.

### Result

RGB32 blocks are 51x20, so the block count comes straight from the dimensions:

| Desktop | Blocks | 30 frames |
| --- | --- | --- |
| 3213x1300 | 63 x 65 = 4095 | 0 assertions |
| 3264x1280 | 64 x 64 = 4096 | assertion on every frame |

I alternated the two sizes four times against the same running server. 4095 was clean every time, 4096 asserted every time:

```
Failed assertion m_blocks.size() == (size_t)GetBlocksCount() in file
/BearWare/TeamTalk5/Library/TeamTalkLib/teamtalk/server/DesktopCache.cpp at line 342
```

Over a longer run of 250 frames at 4096 blocks I got 250 assertions, one per frame.

### Why this may not have reproduced for you

The assertion is completely deterministic, but the hang is not. Across roughly 280 large frames the server kept accepting connections, kept logging, and RSS moved from 406 MB to 413 MB. If you were watching for the hang or a crash, a short test would look clean even while the desktop cache was being corrupted on every single frame.

The assertion is the reliable signal here, not the hang.

### On the review

`BLOCKNUMS_MAX` is 4095 now as you asked, and `GetBlocksCount() - 1` is gone.

On the third point, I think changing `GetBlocksCount() <= BLOCKNUMS_MAX` to `<` would be wrong, and I would rather raise it than quietly do it. That comparison in `DesktopSession::IsValid()` is against a block *count*, while the three asserts in `PacketLayout.cpp` are against a block *number*:

```cpp
// PacketLayout.cpp - block numbers, already exclusive
assert(ii->first < BLOCKNUMS_MAX);

// DesktopSession.cpp:59 - a count, so inclusive is correct
GetBlocksCount() > 0 && GetBlocksCount() <= DesktopPacket::BLOCKNUMS_MAX
```

With `BLOCKNUMS_MAX` at 4095, a count of 4095 gives block numbers 0..4094, so the highest number is `0xFFE` and `0xFFF` stays reserved. Making that check exclusive would cap the count at 4094 and reject 3213x1300, which is the size I verified above runs cleanly. I am happy to change it if you still want it, I just did not want to introduce an off-by-one in the other direction.

### One follow-up

`BLOCKNO_INDEX_MAX` is declared in `PacketLayout.h` but is not referenced anywhere in the repository, and its value `0xFFF` is the same reserved value the duplicate-block list uses as its terminator in `PacketLayout.cpp`:

```cpp
blocknums_single_input.push_back(0xFFF);   // encode, line 1539
uint16_t block_no = 0xFFF;                 // decode, line 1795
```

Giving that terminator a name, and either removing `BLOCKNO_INDEX_MAX` or reusing it for this, would make the reserved value obvious at both ends. I left it out of this PR to keep the change small, but I can open a separate one if you would like it.

Thanks for reading!

computertech1012:
Hi,

@bear101 you asked for a unit test that proves the bug rather than describes it, so I have added one to `CatchDefault.cpp`. It builds a `DesktopPacket` with a non-contiguous duplicate set, serialises it, reads it back and checks what survives:

```
highest usable block number survives     0xFFE round-trips intact
the reserved block number does not       0xFFF is dropped
```

The reason it has to be a non-contiguous set is that a contiguous one takes the range branch, which never writes the terminator. That is also why a zeroed frame buffer will not show any of this, since every block is identical and the duplicates come out as one contiguous range.

The dropped block is the whole mechanism. The receiving `DesktopCache` then holds fewer blocks than the session dimensions call for, which is the assertion at `DesktopCache.cpp:342`.

I left a note in the second section saying that if the encoding is ever changed so every 12-bit value is usable, that section should be inverted to require `0xFFF` survives too. That way the test does not get in the way if you would rather make 4096 blocks work, which is the direction #3393 points.

The Catch unit tests step passed on ubuntu-24.04 in CI, so it builds and runs clean.

I have also opened #3400 for the two problems @Ivankolp reported in #3395, since one of those is an actual hang and I think it is the better explanation for what I originally described in #3386. I have corrected that issue accordingly.

Thanks for reading!

bear101:
superceeded by #3393

PR #3393, opened 2026-08-15
Assign duplicate blocks DesktopPacket::BLOCKNO_INDEX_DUPLICATE

### LINKED ISSUE #3386: TeamTalk Server hangs on 4096-block desktop frame due to off-by-one limit

bear101:


computertech1012:
Hi,

@bear101, naming the reserved value is exactly what I think that code needed, so this is good to see. I ran into the same area in #3392 and I have some test results that I think are relevant here, particularly for `testSendDesktopWindowMaximum`.

### The new test may not fail even without a fix

The test fills the frame with `new byte[wnd.nWidth * wnd.nHeight * 4]`, which is all zeros. Every block is then identical, so the duplicates of block 0 are the contiguous set 1..4095, and `IsBlockRange()` returns true. That takes the range branch in `PacketLayout.cpp`, which never writes the terminator:

```cpp
if(IsBlockRange(dbi->second))
{
    blocknums_range_input.push_back(dbi->first);   // no terminator here
    ...
}
else
{
    blocknums_single_input.push_back(dbi->first);
    ...
    blocknums_single_input.push_back(BLOCKNO_INDEX_DUPLICATE);   // only this path
}
```

The ambiguity only exists on the single-list path, so a uniform bitmap sidesteps it.

I checked this on a stock 5.22.0.5198 Linux server with 51x64 by 20x64, the same dimensions as the test, sending 30 frames each way and changing only the frame contents:

| Frame contents | Assertions in 30 frames |
| --- | --- |
| Uniform bytes, like a zeroed buffer | 0 |
| Varying pattern | 30 |
| Uniform bytes again | 0 |
| Varying pattern again | 30 |

Same server, same session, same desktop size. If the test used a non-uniform pattern I think it would fail on master, and it would then be protecting the actual fix.

### This PR does not change behaviour on its own

Unless I have misread it, the C++ side here is a rename: `BLOCKNO_INDEX_DUPLICATE` is still `0xFFF`, and `BLOCKNUMS_MAX` stays at `0x1000`. So a 4096-block desktop is still accepted and its highest block number is still `0xFFF`, which is still indistinguishable from the terminator on the single-list path.

### Where this meets #3392

This is the part I want to get right before either of us goes further, because the two PRs point in opposite directions.

`testSendDesktopWindowMaximum` asserts that a 4096-block desktop transfers correctly. #3392 rejects that size outright by lowering `BLOCKNUMS_MAX` to 4095. Both cannot be right, so it comes down to whether 4096 blocks should be supported.

If it should be rejected, then #3392 covers it and the test here only needs its dimensions reduced by one block, and I have verified 63x65 at 4095 blocks runs clean.

If it should be supported, then #3392 is the wrong approach and I am happy to close it. In that case the terminator needs to stop being a reserved block number, for example by length-prefixing the single list instead of terminating it with a sentinel, and this PR's rename is a good first step toward that.

I have no strong preference, I would just rather not land a size limit while you are working toward making that size work. Let me know which way you want it and I will adapt #3392 to match, or drop it.

Thanks for reading!

bear101:
I finally managed to write a unit test that reproduces the crash.

PR #3394, opened 2026-08-16
Reject negative ban list ranges

### LINKED ISSUE #3391: TeamTalk Server 5.22 segfaults during channel-ban race
### Description

TeamTalk Server 5.22.0.5198 on Linux can segfault while applying a channel ban in a newly created temporary channel. I saw the same crash four times in about six minutes while users were rapidly entering, leaving and reconnecting.

Each crash had the same executable offset and fault address:

```
segfault at ffffffffffffff18 ... error 5 in tt5srv[d0468,...]
```

The packaged server is stripped. An unstripped server binary from the same 5.22 SDK maps offset `0xd0468` to `teamtalk::ServerUser::GenerateBan(BanTypes, ACE_TString const&) const`, so that symbol should be treated as a lead rather than a complete backtrace.

This looks related to #1659, but that crash was in `GetChannelPath()` when the target was not in a channel. The current crash occurs later in `GenerateBan()` after the target has joined the temporary channel.

### Application

- [ ] qtTeamTalk
- [ ] TeamTalkAndroid
- [ ] iTeamTalk
- [ ] TeamTalkClassic
- [x] TeamTalkServer

### Platform

- [ ] Windows
- [ ] macOS
- [ ] Android
- [ ] iOS
- [x] Linux

### TeamTalk version

5.22.0.5198

1. Start TeamTalk Server 5.22.0.5198 on Linux.
2. Connect a user that can create temporary channels and apply channel bans.
3. Connect several other users and have them repeatedly leave, reconnect and change channels.
4. Create a new temporary channel and join the first user to it.
5. Have one or more of the other users join that channel.
6. Apply a channel ban to one of those users while users are entering, leaving or disconnecting.
7. Delete/recreate the temporary channel and repeat if the race does not occur on the first attempt.

In the four observed failures, the server crashed between roughly 0.5 and 1.5 seconds after the temporary channel was created, immediately after other users began joining it.

### Expected behavior

The ban succeeds, or the server returns `USER_NOT_FOUND` / `CHANNEL_NOT_FOUND` if the user or channel changed during the command.

### Actual behavior

`tt5srv` exits with `SIGSEGV`. A service manager can restart it, but all connected users are disconnected.

There was no out-of-memory event. All four failures used the same `tt5srv+0xd0468` offset.

computertech1012:
Hi,

This fixes #3391 by rejecting negative ban-list ranges before the server indexes the ban vector.

The retained crash evidence pointed to a `BannedUser` pointer 232 bytes before the start of the list, and 232 bytes matches `sizeof(BannedUser)`. That is consistent with `nIndex == -1` on an empty ban list. This keeps the existing login, channel and authorization checks first, then returns `TT_CMDERR_SYNTAX_ERROR` for a negative `nIndex` or `nCount`.

I also added an in-process server regression for `doListBans(0, -1, 1)` and `doListBans(0, 0, -1)`. It checks the exact syntax error, confirms `doListBans(0, 0, 0)` still succeeds, and verifies the server remains responsive afterward.

Validation performed:

- Built `TeamTalk5Pro-jni` and `TeamTalk5SrvTest` on Linux.
- Ran `testListBansInvalidRange` 10 times plain and 10 times encrypted.
- Ran `testBanUser` and `testAutoOperatorChannelUpdate` plain and encrypted.
- Ran the in-process server test class excluding one local IPv4/v6 hostname-resolution environment failure: 37/37 passed plain and 37/37 passed encrypted.
- `git diff --check` passed, apart from normal Windows line-ending warnings.

@bear101, could you please review this when you have time?

Thanks for reading!

### DIFF
--- a/Library/TeamTalkLib/teamtalk/server/ServerNode.cpp
+++ b/Library/TeamTalkLib/teamtalk/server/ServerNode.cpp
@@ -3466,9 +3466,6 @@ ErrorMsg ServerNode::UserListServerBans(int userid, int chanid, int index, int c
             return ErrorMsg(TT_CMDERR_NOT_AUTHORIZED);
     }
 
-    if (index < 0 || count < 0)
-        return ErrorMsg(TT_CMDERR_SYNTAX_ERROR);
-
     std::vector<BannedUser> bans;
     if (banchan)
     {
--- a/Library/TeamTalkLib/teamtalk/server/ServerNode.cpp
+++ b/Library/TeamTalkLib/teamtalk/server/ServerNode.cpp
@@ -3466,6 +3466,9 @@ ErrorMsg ServerNode::UserListServerBans(int userid, int chanid, int index, int c
             return ErrorMsg(TT_CMDERR_NOT_AUTHORIZED);
     }
 
+    if (index < 0 || count < 0)
+        return ErrorMsg(TT_CMDERR_SYNTAX_ERROR);
+
     std::vector<BannedUser> bans;
     if (banchan)
     {
--- a/Library/TeamTalkJNI/test/dk/bearware/TeamTalkServerTestCase.java
+++ b/Library/TeamTalkJNI/test/dk/bearware/TeamTalkServerTestCase.java
@@ -587,6 +587,29 @@ public class TeamTalkServerTestCase extends TeamTalkTestCaseBase {
     }
 
     @Test
+    public void testListBansInvalidRange() {
+
+        TeamTalkSrv server = newServerInstance();
+        TeamTalkBase client = newClientInstance();
+
+        ServerInterleave interleave = new RunServer(server);
+        connect(server, client);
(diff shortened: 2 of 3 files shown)


Library/TeamTalkLib/teamtalk/server/ServerNode.cpp
```diff
@@ -3466,6 +3466,9 @@ ErrorMsg ServerNode::UserListServerBans(int userid, int chanid, int index, int c
             return ErrorMsg(TT_CMDERR_NOT_AUTHORIZED);
     }
 
+    if (index < 0 || count < 0)
```
bear101: Put this in `ServerUser::HandleListServerBan()` instead.

bear101 (changes requested):

computertech1012:
Hi,

Completed, @bear101. I moved the validation into `ServerUser::HandleListServerBans()` and pushed the update.

Thanks for reading!

bear101 (approved):

PR #3400, opened 2026-08-19
Reject malformed 12-bit arrays and bounds check desktop block sizes

### LINKED ISSUE #3395: Fix infinite loop in ConvertFromUInt12Array and heap OOB read in DesktopPacket::GetBlocks
### Description
Two related vulnerabilities in `TeamTalkLib/teamtalk/PacketLayout.cpp` allow malformed UDP packets to cause a complete server deadlock (hang) or heap out-of-bounds read (crash):

1. **Infinite Loop in `ConvertFromUInt12Array` (Server & Client Hang):**
   - **File:** `TeamTalkLib/teamtalk/PacketLayout.cpp` (lines ~67-93)
   - **Details:** 12-bit uint array decoding unpacks 2 integers per 3 bytes (`src_len % 3 == 0`). When a packet containing `FIELDTYPE_BLOCKNUMS_AND_SIZES` or `FIELDTYPE_ENCFRAMESIZES` has a length where `length % 3 != 0` (e.g. 16 bytes or 1 byte), the decoding loop does not terminate properly, entering an infinite loop.
   - **Impact:** On the server, the UDP reactor thread gets trapped in a 100% CPU loop while holding the `ServerNode` mutex (`GUARD_OBJ`), locking up all TCP commands (login, join, message) and UDP traffic. On clients, incoming malformed audio frame sizes lock the audio processing thread at 100% CPU.

2. **Heap Out-Of-Bounds Read in `DesktopPacket::GetBlocks` (Server Crash):**
   - **File:** `TeamTalkLib/teamtalk/PacketLayout.cpp` (lines ~1731-1742)
   - **Details:** The parser iterates through block numbers and sizes declared in `FIELDTYPE_BLOCKNUMS_AND_SIZES` and copies data from `FIELDTYPE_BLOCKS_DATA` without validating if the total available buffer matches the cumulative block sizes.
   - **Impact:** If declared block sizes exceed the actual payload size, `memcpy` reads past the receive buffer bounds, causing an immediate segmentation fault (SIGSEGV / Access Violation).

---

#### 1. Validate length in `ConvertFromUInt12Array`:
```cpp
bool ConvertFromUInt12Array(const uint8_t* src, size_t src_len, ...) {
    if (src_len % 3 != 0) {
        return false; // Reject malformed 12-bit array length
    }
    // ...
}
```

#### 2. Add bounds check in `DesktopPacket::GetBlocks`:
```cpp
if (byte_pos + block_size > total_blocks_data_len) {
    return false; // Bounds check failure: discard corrupt packet
}
```

computertech1012:
Hi,

This fixes both halves of #3395. @Ivankolp found them and the analysis is his.

### What happens

Two 12-bit values pack into three bytes, so a valid array ends either on a whole pair or on a two byte tail. If a field length leaves exactly one byte over, the final `else` in the decode loop is reached and it does not advance `i`:

```cpp
for(uint16_t i=0;i<source_size;)
{
    if(source_size - i >= 3)        { ...; i += 3; }
    else if(source_size - i == 2)   { ...; i += 2; }
    else assert(0);                 // i unchanged
}
```

With `assert()` compiled out, that is an infinite loop.

I pulled the loop out into a standalone file, stubbed the bit unpacking since only the index arithmetic decides termination, and built it without the assert. Before:

```
  length  3 (len % 3 = 0): terminated, 2 values
  length  6 (len % 3 = 0): terminated, 4 values
  length  2 (len % 3 = 2): terminated, 1 values
  length  5 (len % 3 = 2): terminated, 3 values
  length  4 (len % 3 = 1): NO TERMINATION: i stuck at 3, 1 byte remaining, 100M iterations
  length  7 (len % 3 = 1): NO TERMINATION: i stuck at 6, 1 byte remaining, 100M iterations
  length 16 (len % 3 = 1): NO TERMINATION: i stuck at 15, 1 byte remaining, 100M iterations
  length  1 (len % 3 = 1): NO TERMINATION: i stuck at 0, 1 byte remaining, 100M iterations
```

After:

```
  length  3 (len % 3 = 0): terminated, 2 values
  length  6 (len % 3 = 0): terminated, 4 values
  length  2 (len % 3 = 2): terminated, 1 values
  length  5 (len % 3 = 2): terminated, 3 values
  length  4 (len % 3 = 1): terminated, 2 values
  length  7 (len % 3 = 1): terminated, 4 values
  length 16 (len % 3 = 1): terminated, 10 values
  length  1 (len % 3 = 1): terminated, 0 values
```

Valid lengths decode to the same number of values as before, so this does not change good traffic.

### Why not reject everything that is not a multiple of three

The issue suggests `src_len % 3 != 0`. I think that would drop valid packets. An odd number of values legitimately encodes to a two byte tail:

| Values | Bytes | Bytes % 3 |
| --- | --- | --- |
| 1 | 2 | 2 |
| 3 | 5 | 2 |
| 4095 | 6143 | 2 |

The `== 2` branch already handles those. Only a remainder of one is unhandled, so that is all this rejects.

### The change

`ReadUInt12Array` now checks its field length up front, which is what `ReadUInt16Array` already does a few lines below:

```cpp
// ReadUInt16Array, existing
if((field_size == 0u) || ((field_size % 2) != 0))
    return false;

// ReadUInt12Array, added
if((field_size == 0u) || ((field_size % 3) == 1))
    return false;
```

The loop also breaks rather than spinning, since four other call sites reach it without going through `ReadUInt12Array`.

### Second half: the out of bounds read in `GetBlocks`

The block sizes in `FIELDTYPE_BLOCKNUMS_AND_SIZES` come off the wire, and nothing checked them against how much `FIELDTYPE_BLOCKS_DATA` actually holds, so the block pointer walks past the receive buffer.

`GetBlockFragments()` a few lines below already guards this exact thing, so I used the same shape:

```cpp
// GetBlockFragments, existing
assert(byte_pos+bf.frag_size<=data_size);
if(byte_pos+bf.frag_size>data_size) //buffer overflow check
    return false;

// GetBlocks, added
assert(byte_pos + bb.block_size <= blockdata_len);
if(byte_pos + bb.block_size > blockdata_len) //buffer overflow check
    return false;
```

I ran the loop standalone against a few declared size lists before and after:

```
  honest packet
    current : declared total=300, payload=300 -> within bounds
    fixed   : accepted

  one block oversized
    current : declared total=1100, payload=300 -> READS PAST THE BUFFER
    fixed   : rejected, returns false

  sizes far exceed payload
    current : declared total=54464, payload=300 -> READS PAST THE BUFFER
    fixed   : rejected, returns false

  single huge block
    current : declared total=65535, payload=10 -> READS PAST THE BUFFER
    fixed   : rejected, returns false
```

Worth noting the third case. Two blocks of 60000 declare 120000 bytes, but `byte_pos` is `uint16_t` so it wraps to 54464. A check on the total after the loop would miss that, which is why this one is per block.

### On a unit test

You asked for a test on my other issue and that was fair, so I want to be straight about this one. `ConvertFromUInt12Array` and `ReadUInt12Array` are both file-static, so there is no clean seam to call them from `CatchDefault.cpp` without changing their linkage, and a test driving the loop through a hand-built packet would hang rather than fail if it ever ran without this fix.

`GetBlocks` is reachable, so a test for the second half is straightforward if you want one. Say where you would like it and I will add it.

One thing worth noting: `ConvertToUInt12Array` can only ever produce a length that is 0 or 2 more than a multiple of three, so the encoder cannot generate this itself. It has to arrive from outside.

Thanks for reading!

### DIFF
--- a/Library/TeamTalkLib/teamtalk/PacketLayout.cpp
+++ b/Library/TeamTalkLib/teamtalk/PacketLayout.cpp
@@ -88,7 +88,15 @@ namespace teamtalk
                 target.push_back(v1);
                 i += 2;
             }
-            else assert(0);
+            else
+            {
+                // A single trailing byte cannot hold a 12-bit value. This
+                // branch used to leave 'i' unchanged, so a field length of
+                // 1 modulo 3 looped forever wherever assert() is compiled
+                // out. Stop instead of spinning.
+                assert(0);
+                break;
+            }
         }
     }
 
@@ -96,7 +104,11 @@ namespace teamtalk
                          std::vector<uint16_t>& output)
     {
         uint16_t const field_size = READFIELD_SIZE(ptr);
-        if(field_size == 0u)
+        // Two 12-bit values pack into three bytes, so a valid array ends on a
+        // whole pair or on a two byte tail. One byte left over cannot be
+        // decoded, which makes the field malformed. ReadUInt16Array already
+        // rejects its own bad lengths the same way.
+        if((field_size == 0u) || ((field_size % 3) == 1))
             return false;
         const uint8_t* field_ptr = READFIELD_DATAPTR(ptr);
         ConvertFromUInt12Array(field_ptr, field_size, output);
--- a/Library/TeamTalkLib/teamtalk/PacketLayout.cpp
+++ b/Library/TeamTalkLib/teamtalk/PacketLayout.cpp
@@ -816,12 +816,8 @@ namespace teamtalk
         if(ptr == nullptr)
             return v_frm_sizes;
 
-        uint16_t const length = READFIELD_SIZE(ptr);
-        ptr = READFIELD_DATAPTR(ptr);
-
-        const auto* frm_sizes = ptr;
-
-        ConvertFromUInt12Array(frm_sizes, length, v_frm_sizes);
+        if(!ReadUInt12Array(ptr, FIELDTYPE_ENCFRAMESIZES, v_frm_sizes))
+            v_frm_sizes.clear();
 
         return v_frm_sizes;
     }
@@ -1720,9 +1716,6 @@ namespace teamtalk
         if(blocknums_ptr == nullptr)
             return false;
 
-        uint16_t const blocknums_len = READFIELD_SIZE(blocknums_ptr);
-        blocknums_ptr = READFIELD_DATAPTR(blocknums_ptr);
-
         const uint8_t* blockdata_ptr = FindField(FIELDTYPE_BLOCKS_DATA);
         if(blockdata_ptr == nullptr)
             return false;
@@ -1732,9 +1725,9 @@ namespace teamtalk
 
         
         std::vector<uint16_t> blocks_n_sizes;
-        const auto* blocks_array = blocknums_ptr;
-
-        ConvertFromUInt12Array(blocks_array, blocknums_len, blocks_n_sizes);
+        if(!ReadUInt12Array(blocknums_ptr, FIELDTYPE_BLOCKNUMS_AND_SIZES,
+                            blocks_n_sizes))
+            return false;
 
         assert(blocks_n_sizes.size() % 2 == 0);
         if((blocks_n_sizes.size() % 2) != 0u)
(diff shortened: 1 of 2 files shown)


computertech1012:
Hi,

@bear101, I added focused Catch regressions for a malformed 12-bit array tail and a declared block size that exceeds the available block data. Both are now rejected safely, and the focused Linux test passed all 6 assertions.

Thanks for reading!

bear101 (approved):

PR #3401, opened 2026-08-19
Bounds check desktop range decoders against out-of-bounds reads

computertech1012:
Hi,

While looking at the desktop parsing code around #3386 I found two more spots in `PacketLayout.cpp` where a value read off the wire is used to walk a vector without checking its length first. Both are out of bounds reads reachable on the server from a client, so I wanted to fix them the same way `GetBlocks` already handles its own array.

Same threat model as #3386: an authenticated client sending a crafted packet, not a pre-auth remote.

**1. `GetDuplicateBlocks`, the `BLOCK_DUP_RANGE` branch**

It reads three values per range entry:

```cpp
for(size_t i=0;i<blocknums_range.size();i+=3)
{
    for(uint16_t block_no=blocknums_range[i+1];
        block_no<=blocknums_range[i+2];block_no++)
        ...
    ... blocknums_range[i] ...
}
```

There was no check that the decoded length is a multiple of three. A three byte range field decodes to two values, so the first iteration reads `[2]` past the vector. `GetBlocks` guards its own array with `% 2 == 0`; this branch had nothing.

Reached on the server from `DesktopCache::UpdateCurrentDesktopWindow`, the same function as the assertion in #3386.

**2. `GetPacketsAcked`, the `PACKETRANGE_ACK` branch**

It reads two values per range:

```cpp
ReadUInt16Array(ptr, FIELDTYPE_PACKETRANGE_ACK, packetnums);
assert(packetnums.size() % 2 == 0);
for(size_t i=0;i<packetnums.size();i+=2)
{
    ... packetnums[i+1] ...
}
```

The parity check is an `assert`, so it is gone in a release build. `ReadUInt16Array` requires an even byte length but still returns `field_size/2` values, which can be odd (a six byte field gives three values), so an odd count reads `[i+1]` past the vector on the last step.

Reached on the server from `ReceivedDesktopAckPacket` (`PACKET_KIND_DESKTOP_ACK`).

**How I confirmed it**

I could not build the full library locally, so I pulled the two loops out exactly as written and drove them with `vector::at()`, which throws precisely where `operator[]` would read past the end:

```
GetDuplicateBlocks range branch
  3-byte field  -> 2 values (size%3=2): OUT-OF-BOUNDS READ
  6-byte field  -> 4 values (size%3=1): OUT-OF-BOUNDS READ
  well-formed   -> 3 values (size%3=0): ok
  well-formed   -> 6 values (size%3=0): ok

GetPacketsAcked range branch
  6-byte field -> 3 values (size 3): OUT-OF-BOUNDS READ
  4-byte field -> 2 values (size 2): ok
  8-byte field -> 4 values (size 4): ok
```

**The fix**

One guard each, both mirroring what `GetBlocks` already does:

```cpp
// GetDuplicateBlocks
if((blocknums_range.size() % 3) != 0u)
    return false;

// GetPacketsAcked
if((packetnums.size() % 2) != 0u)
    return true;
```

**On a unit test**

These functions are file-static in reach, so the only way to exercise them is a hand-crafted malformed packet, and the normal packet builder never produces one. I did not want to hand-corrupt bytes into a test I cannot run against the current code to confirm it actually reaches the bug, since that would prove nothing. The extracted-loop output above is what demonstrates the read. If you would like an in-tree test, tell me where you would want it and I will add one against a crafted packet.

Thanks for reading!

### DIFF
--- a/Library/TeamTalkLib/teamtalk/PacketLayout.cpp
+++ b/Library/TeamTalkLib/teamtalk/PacketLayout.cpp
@@ -1831,6 +1831,7 @@ namespace teamtalk
             // step. GetBlocks() guards its own array the same way.
             if((blocknums_range.size() % 3) != 0u)
                 return false;
+            assert(blocknums_range.size() % 3 == 0);
 
             for(size_t i=0;i<blocknums_range.size();i+=3)
             {
@@ -2003,9 +2004,9 @@ namespace teamtalk
         // length but still yields field_size/2 values, which can be odd, so the
         // count has to be checked at run time and not only with assert(), which
         // is compiled out of a release server.
-        assert(packetnums.size() % 2 == 0);
         if((packetnums.size() % 2) != 0u)
-            return true;
+            return false;
+        assert(packetnums.size() % 2 == 0);
         //insert ranges
         for(size_t i=0;i<packetnums.size();i+=2)
         {
--- a/Library/TeamTalkLib/teamtalk/PacketLayout.cpp
+++ b/Library/TeamTalkLib/teamtalk/PacketLayout.cpp
@@ -1825,6 +1825,13 @@ namespace teamtalk
             std::vector<uint16_t> blocknums_range;
             ConvertFromUInt12Array(u8_info_ptr, info_size, blocknums_range);
 
+            // Each range entry is three values: a source block and a low/high
+            // pair. The values come off the wire, so the count can be anything;
+            // without this the loop below reads past the vector on the last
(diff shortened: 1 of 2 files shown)


computertech1012:
Hi,

@bear101, I added focused Catch regressions for incomplete duplicate-block and acknowledgement range fields. Both malformed cases are now rejected safely, and the focused Linux test passed all 8 assertions.

Thanks for reading!

bear101 (approved):

PR #3403, opened 2026-08-19
Ability to filter and search public servers by country

amirmahdifard:
some people in some limited countrys that have their internet limited, Are able to connect to only some servers hosted in some countrys only.
For example, Iranian people are only able to connect to midle east countrys, not the others.
This feature allows user to only see the countrys that they are able to connect or want to see.


bear101 (changes requested):
Change country to QComboBox with drop-down list that extracts the countries from the server model.

bear101:
Then reopen afterwards

PR #3405, opened 2026-08-21
Fix screen reader focus bugs in android 15 and up with edg to edg

amirmahdifard:
When we introduced edg to edg at first, we didn't handle it properly.
It was custom and incomplete.
Talkback and other screen readers didn't cover the action bars properly, skipped over them and they were placed at the last, incorrect auder.
By switching to the standard edg to edg handling that android provides in all newly created empty templates projects and also properly grabbing the actual content of the activity and set the padding to it, Everything is fixed for everyone. Now action bar, focus, and everything acts as normal
Also, portrate, landskape, and any side by side support is complete because we are handling this like how android suggest, raather than manually handling the insets

### DIFF
--- a/Client/TeamTalkAndroid/src/main/java/dk/bearware/gui/EdgeToEdgeHelper.java
+++ b/Client/TeamTalkAndroid/src/main/java/dk/bearware/gui/EdgeToEdgeHelper.java
@@ -27,18 +27,16 @@ import android.app.Activity;
 import android.os.Build;
 import android.view.View;
 
+import androidx.core.graphics.Insets;
 import androidx.core.view.ViewCompat;
 import androidx.core.view.WindowInsetsCompat;
 
 public class EdgeToEdgeHelper {
-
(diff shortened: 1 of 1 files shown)


beqabeqa473:
Please test portrait and landscape modes.

beqabeqa473 (approved):

beqabeqa473:
@hwangsihu i didn't merge this pr intentionally to wait visual confirmation. what is the reason for such a rush?

hwangsihu:
Oh, I apologize. I have always assumed that ‘approval = it is okay to merge.’ I will be more careful next time.

Edit: There was a comment above as well... I'll really be careful. I'm used CLI before.

amirmahdifard:
@beqabeqa473 I've tested portrate and landskape modes, They are completely ok! I'm just waiting for a sited user to conferm! Today I will have someone sited user to conferm it for me, and It will be merged after that! This pr fixes that issue in the standard edg to edg introduced by android. it self, It's even on empty template projects.

PR #3410, opened 2026-08-26
Add channel types to TeamTalk files

### LINKED ISSUE #31: Channel type should be included in .tt files
A channel in a .tt file is always created as "default channel type". Classroom, no voice activation, etc channel types should be possible to specify.

computertech1012:
Hi,

This adds channel-type support to Qt TeamTalk `.tt` files and fixes #31.

The Qt client can now:
- select any supported channel-type combination from the Server dialog;
- store the selection for saved servers and recent hosts;
- read and write `<channel-type>` while keeping existing `.tt` files compatible;
- apply the selected type when creating a missing channel;
- mask unsupported channel-type bits while preserving supported values.

Tests:
- focused Qt harness covering the GUI bitmask, XML round trips, legacy/default behavior, invalid and unknown values, and real QSettings save/load/delete paths;
- full Qt TeamTalk client build with MSVC 19.44 and Qt 6.10.1;
- isolated TeamTalk Server 10332 canary creating and observing combined type 52, followed by automatic temporary-channel removal.

Thanks for reading!

### DIFF
--- a/Client/qtTeamTalk/utilxml.cpp
+++ b/Client/qtTeamTalk/utilxml.cpp
@@ -18,6 +18,17 @@
 #include "utilxml.h"
 #include "appinfo.h"
 
+namespace
+{
+constexpr UINT32 CHANNELTYPE_MASK = CHANNEL_PERMANENT |
+                                    CHANNEL_SOLO_TRANSMIT |
+                                    CHANNEL_CLASSROOM |
+                                    CHANNEL_OPERATOR_RECVONLY |
+                                    CHANNEL_NO_VOICEACTIVATION |
+                                    CHANNEL_NO_RECORDING |
+                                    CHANNEL_HIDDEN;
+}
+
 void processTrustedXML(const QDomElement& hostElement, HostEntry& entry)
 {
     QDomElement trusted = hostElement.firstChildElement("trusted-certificate");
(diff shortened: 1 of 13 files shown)


Client/qtTeamTalk/utilxml.cpp
```diff
@@ -75,6 +76,14 @@ void processJoinXML(const QDomElement& hostElement, HostEntry& entry)
         tmp = join.firstChildElement("join-last-channel");
         if (!tmp.isNull())
             entry.lastChan = tmp.text() == "true";
+        tmp = join.firstChildElement("channel-type");
+        if (!tmp.isNull())
+        {
+            bool ok = false;
+            UINT32 chantype = tmp.text().toUInt(&ok);
+            if (ok && !(chantype & ~CHANNELTYPE_ALL))
```
bear101: This expression is a bit complicated. Maybe it would be simpler like this:
```
if (ok)
    entry.chantype = static_cast<ChannelTypes>(chantype & CHANNELTYPE_ALL);
```

bear101 (commented):
I moved the supported channel types to utiltt.h instead of having them in the utilxml.cpp file.

Client/qtTeamTalk/utilxml.cpp
computertech1012: Hi,

@bear101, completed. I simplified the channel-type parsing as suggested so unsupported bits are masked with CHANNELTYPE_ALL while supported bits are preserved.

Thanks for reading!

bear101 (approved):

Issue #2903, opened 2025-09-30
A critical and annoying bug in qt teamtalk. All files in a channel doesn't load when you join the channel normally


amirmahdifard:
### Description

When you login to a server, and join a channel that has 100 files in it for example, you only see a fue of them for example 18. If you want all of the files to load for you, you must move yourself to another channel, and move your self to this channel again, or if you don't have the right, ask someone to move you to the channel for the files to load properly. Not sure from which version this bug happened, I see it from 5.18 for example. I think it happned from 5.17 due to some changes to joining channels and added checkbox to join last channel, etc. Please fix this bug.

### Application

- [x] qtTeamTalk
- [ ] TeamTalkAndroid
- [ ] iTeamTalk
- [ ] TeamTalkServer

PR #3411, opened 2026-08-26
Preserve classroom transmit permissions from other channels

### LINKED ISSUE #1928: Transmiting Issue in Classroom Channel
### Description

As an administrator of the TeamTalk server, I able to transmit and untransmit other people from another channel. Let's say I'm on channel B, I'm able to transmit and untransmit each user or all user in channel A. In this case, I put on channel A so that all user can transmit voice . When I set 1 user in channel A to able to transmit voice regarless of allow all voice transmition settings while I'm in the same channel has no problem. The problem came when I do that from another channel. What will happend is all user will not able to transmit voice and the user that I set so that he/she able to transmit voice can continue talking. I need to return to channel A and allow all voice transmition again.

### Application

- [x] qtTeamTalk
- [ ] TeamTalkAndroid
- [ ] iTeamTalk
- [ ] TeamTalkClassic
- [ ] TeamTalkServer

### Platform

- [x] Windows
- [ ] macOS
- [ ] Android
- [ ] iOS
- [ ] Linux

### Expected behavior

Allow all transmition setting should not be effected (spesificly turned off) while you toggle allow transmition for spesific user from another channel.

### Actual behavior

Allow all transmition setting is effected (spesificly turned off) while you toggle allow transmition for spesific user from another channel.

1. Set channel A to allow all transmition settings (doesn't matter which settings).
2. Join to channel B.
3. Toggle 1 of the user in channel to allow transmition settings (doesn't matter which settings).

computertech1012:
Hi,

This fixes #1928 in the Qt client.

When an administrator viewed a classroom channel other than their current channel, the channel tree cleared the transmission check states. Changing one permission could therefore rebuild the channel's transmission list from missing states and clear existing permissions.

The channel tree now loads the actual check states for any classroom channel the current user is allowed to modify. `ChannelsTree::getTransmitUsers()` follows the same checkbox-model path for both the current channel and other channels.

Tests:
- Focused classroom permission testing covering allow-everyone and individual transmission entries.
- Clean build of the complete Qt TeamTalk client with MSVC and Qt.

Thanks for reading!


bear101:
This doesn't look like the correct way to fix the issue. There should be no difference to ChannelsTree::getTransmitUsers() whether it's "my channel" or another channel. An administrator can control transmit users both in "my channel" and other channels.

I think the issue is related to the model in channel's tree that controls QTreeWidgetItem::checkState(). It looks like it's not being updated correctly when you're administrator.

computertech1012:
Hi,
HM, @bear101 I think your right.
I will check tomorrow.
thanks for reading!

computertech1012:
Hi,

@bear101, fixed. You were right that the issue was in the channel tree model. It now loads the transmission check states for classroom channels that an administrator or channel modifier is allowed to control, even when it is not their current channel. ChannelsTree::getTransmitUsers() now uses the same checkbox-model path for both cases.

Thanks for reading!

amirmahdifard:
@computertech1012 hi
Can you also fix this bug? or you haven't faced it?
https://github.com/BearWare/TeamTalk5/issues/2903

computertech1012:
Hi,
@amirmahdifard I could look into it.
I have also faced this problem.
Thanks for reading!

bear101:
Looks good. Thanks!

PR #3419, opened 2026-09-02
Android client. Label parent channel shortcut in channel list

trypsynth:
## Problem

In the Android client's channel list, the row that navigates up to the parent channel is unlabelled for screen readers.

`ChannelListAdapter.getView()` inflates `item_channel_back.xml` for `PARENT_CHANNEL_VIEW_TYPE` but never populates it:

```java
case PARENT_CHANNEL_VIEW_TYPE :
    // show parent channel shortcut
    if (convertView == null ||
            convertView.findViewById(R.id.parentname) == null)
        convertView = inflater.inflate(R.layout.item_channel_back, parent, false);
    break;
```

The layout supplies a hardcoded `".."` in `parentname` and leaves `population` and `chantopic` empty, and the back arrow is `importantForAccessibility="no"`. So the row's entire accessible label is `".."`, which TalkBack does not announce as anything useful.

This matters more than it might look. Rows are ordered sticky channels → users → parent shortcut → subchannels, so on a server with a flat channel structure, this row is the *only* way back to the channel list once you have joined a channel. A TalkBack user swipes past every user in the channel, reaches an item that announces nothing, and has no way to know it is the way back.

## Change

- `parentname` now carries a static `@string/text_parent_channel` ("Parent channel") set from the layout, replacing the hardcoded `".."`.
- `chantopic` and `population` are filled in from the parent channel, so the row also says something about where it leads.

The label is deliberately *not* the parent's name — that would make the row read like just another subchannel, which is exactly what it needs to be distinguishable from (thanks for catching that).

Note that the user count is set unconditionally rather than behind the `nMaxUsers > 0` guard used in `CHANNEL_VIEW_TYPE`, so a recycled view cannot keep a stale count.

## Testing

Tested on device: a debug build installed on a Pixel 9a announces the row as "Parent channel" under TalkBack, where previously it was silent.

The build was made against the v5.22a Android SDK's prebuilt `TeamTalk5.jar` and JNI libraries (`Library/TeamTalkJNI/src` is identical between v5.22a and this tree, so the wrapper API matches):

- `./gradlew assembleDebug` — BUILD SUCCESSFUL, all four ABIs plus universal.
- Verified in the packaged APK with `aapt2` that `item_channel_back.xml`'s `parentname` binds `android:text` to `0x7f10024a` = `string/text_parent_channel` = "Parent channel".
- `./gradlew lintDebug` — no new findings in the changed files; `item_channel_back.xml` is clean and the `MainActivity.java` and `strings.xml` findings are all pre-existing and at unrelated lines.

### DIFF
--- a/Client/TeamTalkAndroid/src/main/java/dk/bearware/gui/MainActivity.java
+++ b/Client/TeamTalkAndroid/src/main/java/dk/bearware/gui/MainActivity.java
@@ -1317,6 +1317,28 @@ private EditText newmsg;
                         if (convertView == null ||
                                 convertView.findViewById(R.id.parentname) == null)
                             convertView = inflater.inflate(R.layout.item_channel_back, parent, false);
+
+                        TextView parentname = convertView.findViewById(R.id.parentname);
+                        TextView parenttopic = convertView.findViewById(R.id.chantopic);
+                        TextView parentpopulation = convertView.findViewById(R.id.population);
+
+                        String parentchanname;
+                        if (channel.nParentID == 0) {
+                            // show server name as channel name for root channel
+                            ServerProperties srvprop = new ServerProperties();
+                            getClient().getServerProperties(srvprop);
+                            parentchanname = srvprop.szServerName;
+                        }
+                        else {
+                            parentchanname = channel.szName;
+                        }
+                        parentname.setText(parentchanname);
+                        parentname.setContentDescription(getString(R.string.text_parent_channel, parentchanname));
+                        parenttopic.setText(channel.szTopic);
+
+                        int parentusers = Utils.getUsers(channel.nChannelID, getService().getUsers()).size();
+                        parentpopulation.setText((parentusers > 0) ? String.format(Locale.ROOT, "(%d)", parentusers) : "");
+
                         break;
 
                     case CHANNEL_VIEW_TYPE :
--- a/Client/TeamTalkAndroid/src/main/java/dk/bearware/gui/MainActivity.java
+++ b/Client/TeamTalkAndroid/src/main/java/dk/bearware/gui/MainActivity.java
@@ -1318,22 +1318,9 @@ private EditText newmsg;
                                 convertView.findViewById(R.id.parentname) == null)
                             convertView = inflater.inflate(R.layout.item_channel_back, parent, false);
 
-                        TextView parentname = convertView.findViewById(R.id.parentname);
                         TextView parenttopic = convertView.findViewById(R.id.chantopic);
                         TextView parentpopulation = convertView.findViewById(R.id.population);
 
-                        String parentchanname;
-                        if (channel.nParentID == 0) {
-                            // show server name as channel name for root channel
-                            ServerProperties srvprop = new ServerProperties();
-                            getClient().getServerProperties(srvprop);
-                            parentchanname = srvprop.szServerName;
-                        }
-                        else {
-                            parentchanname = channel.szName;
-                        }
-                        parentname.setText(parentchanname);
-                        parentname.setContentDescription(getString(R.string.text_parent_channel, parentchanname));
                         parenttopic.setText(channel.szTopic);
 
                         int parentusers = Utils.getUsers(channel.nChannelID, getService().getUsers()).size();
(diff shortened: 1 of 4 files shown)


Client/TeamTalkAndroid/src/main/java/dk/bearware/gui/MainActivity.java
```diff
@@ -1317,6 +1317,28 @@ public View getView(int position, View convertView, ViewGroup parent) {
                         if (convertView == null ||
                                 convertView.findViewById(R.id.parentname) == null)
                             convertView = inflater.inflate(R.layout.item_channel_back, parent, false);
+
+                        TextView parentname = convertView.findViewById(R.id.parentname);
```
bear101: Shouldn't text label just be "parent channel". Otherwise it's spoken like a subchannel.

bear101: Hm, ok, you prefix the name with "Go to parent channel:"

Client/TeamTalkAndroid/src/main/res/layout/item_channel_back.xml
```diff
@@ -1,6 +1,5 @@
 <?xml version="1.0" encoding="utf-8"?>
 <RelativeLayout xmlns:android="http://schemas.android.com/apk/res/android"
-    xmlns:tools="http://schemas.android.com/tools"
```
bear101, line 3: Why is this line removed?

trypsynth:
Agreed, and fixed in a2477cc — the label is now a static "Parent channel" string set straight from the layout, so it can't be mistaken for a subchannel. That also let me drop the name lookup and the separate content description entirely, so the change is smaller than before.

I kept the topic and user count populated, since those `TextView`s are in the layout and were never filled. Happy to drop them too if you'd rather the row stayed a plain label.

Also verified this build: `assembleDebug` against the v5.22a SDK's prebuilt JNI libraries succeeds, `aapt2` confirms the binding in the packaged APK, and `lintDebug` reports nothing new in the changed files.

Client/TeamTalkAndroid/src/main/res/layout/item_channel_back.xml
trypsynth: It was only there for the `tools:ignore="HardcodedText"` on the hardcoded `".."`. Once that text became `@string/text_parent_channel` the suppression was no longer needed, which left the namespace declaration unused, so I removed it in the same edit.

Purely a tidy-up — nothing depends on it. I checked whether lint would complain about leaving it in and it does not (`UnusedNamespace` deliberately exempts the `tools` namespace), so if you would rather keep the diff to the lines that actually matter, say the word and I will put it back.

Client/TeamTalkAndroid/src/main/java/dk/bearware/gui/MainActivity.java
trypsynth: That prefix is gone now — it was a content description on top of the name, which is exactly the confusion you flagged in the other thread. As of a2477cc the row is just a static "Parent channel" label set from the layout, with no name lookup and no separate content description.

Client/TeamTalkAndroid/src/main/res/layout/item_channel_back.xml
trypsynth: Put it back in e1d80dc. The layout diff is now just the one `android:text` attribute.

trypsynth:
Tested on device: I installed a debug build on a Pixel 9a and confirmed with TalkBack that the row is now announced as "Parent channel", where previously it was silent. That closes the on-device gap I noted in the description.

`xmlns:tools` is restored in e1d80dc, so the layout diff is down to the single `android:text` attribute.

amirmahdifard:
@trypsynth hi. This is correct, but I think it is not wurth it this way anymore.
I wanted to rewrite the channels list, that on each channel, talkback says expanded or collapsed. If we click a channel, it should become expanded, and right below it, it shows all users of that channels and below those, it lists the subchannels of that channel. This way, we have the possibility to have multiple channels expanded and look at all of them at the same time. You can look at the teamtalk flutter client written by beka.
The implementation should look like this. A set that holds all expanded channel ids, and a helper that adds all users and subchannels of a channel below it if the id of that channel is inside that set. And clicking a channel once adds that channel id to the expanded channels set, clicking on it again, removes that channel id from the expanded channels set.

bear101 (approved):

PR #3420, opened 2026-09-02
Android client. Add neutral gender option

trypsynth:
Follows up on discussion #2896.

The Android gender setting is a single checkbox — unchecked for a male icon, checked for a female one. TeamTalk has supported a third gender, neutral, for a while, and the Qt client already offers all three. The Android client could already *display* `STATUSMODE_NEUTRAL` for other users, but gave no way to select it.

## Change

- `pref_general.xml` swaps the `CheckBoxPreference` for a `ListPreference` offering Male, Female and Neutral, in the same order `preferencesdlg.cpp` uses.
- Entries reuse the existing `user_state_male` / `_female` / `_neutral` strings, so no new translations are needed.
- Adds `STATUSMODE_GENDER_MASK`, mirroring `qtTeamTalk/common.h`, and uses it when clearing the gender bits. `PreferencesActivity` previously cleared only `STATUSMODE_FEMALE`, which would have left `STATUSMODE_NEUTRAL` set when switching away from neutral.

The old boolean lives under `gender_checkbox` and a `ListPreference` persists a String, so reusing that key would throw `ClassCastException` on upgrade. The list uses a new key and `Utils.migrateGenderPreference()` carries the old value over the first time it is needed, so existing users keep their setting. Male stays the default, matching current Android behaviour rather than the Qt client's `GENDER_NEUTRAL` default — happy to switch if you would rather the clients agree.

## Why the diff touches 24 files

The old summary string "Check for female icon" describes a checkbox that no longer exists, so it is removed — one line each from the default strings and 15 translations. That is 16 of the 24 files. The actual change is `Utils.java`, `pref_general.xml`, the new `arrays.xml`, and three small edits in `TeamTalkConstants`, `Preferences` and `TeamTalkService`.

## Testing

Tested on device: installed on a Pixel 9a and exercised with TalkBack, including selecting neutral.

- `./gradlew assembleDebug` succeeds; `./gradlew lintDebug` reports nothing new in the changed files. In particular there is no `ExtraTranslation`, confirming the removed string is gone from every locale.
- Two of the translation files (`values-fa-rIR`, `values-pt`) use CRLF line endings; they were edited byte-exactly so the diff stays at one line each.

The upgrade path is verified too: seeding the old `gender_checkbox=true` into the app's shared preferences and then opening Settings showed Female in the new picker, and selecting Neutral persisted across an app restart.

Neutral users still get the male icon, since Android uses PNG assets (`man_*` / `woman_*`) and there is no neutral artwork — the Qt client sidesteps this with emoji. The TalkBack announcement is already correct. Happy to follow up if you want an icon.


amirmahdifard:
@trypsynth i don't trust ai coding so much, i'd raather implement this feature manually myself.

trypsynth:
Please don't bullshit me, you've seen my projects and should know I'm smarter than that. Not only am I being told to not use AI coding when there is no contributing guideline for this and based on nothing but the way my PR body was written, but you seem so damn set against this feature for some weird reason. Multiple issues got opened about it and converted to discussions, and now you're giving me hell because I used Claude to draft the PR body after testing on-device and reviewing every single line of code by hand? Happy to address specific technical feedback on the diff from the maintainers, but this feels unconstructive. I did work for you that you've been saying you would for a while, I don't understand the pushback at all.

BlindMaster24:
> Please don't bullshit me, you've seen my projects and should know I'm smarter than that. Not only am I being told to not use AI coding when there is no contributing guideline for this and based on nothing but the way my PR body was written, but you seem so damn set against this feature for some weird reason. Multiple issues got opened about it and converted to discussions, and now you're giving me hell because I used Claude to draft the PR body after testing on-device and reviewing every single line of code by hand? Happy to address specific technical feedback on the diff from the maintainers, but this feels unconstructive. I did work for you that you've been saying you would for a while, I don't understand the pushback at all.

Don't pay attention to it, and most importantly, don't make excuses. Those who understand have already agreed on everything a long time ago. @amirmahdifard First, add a ban to the contributing guidelines and show why coding manually is better than writing it with something like Claude and then testing it on a device and reviewing the code by hand.

amirmahdifard:
The reason implementing features in android client is slow is due to the pull requests were merged sloly. Otherwise I would've implemented these simple and small things over a long time.

trypsynth:
Okay, good for you? That literally has nothing to do with anything you previously said, and doesn't explain why you're giving me hell for having Claude generate the prose on this PR, after I've made it clear I tested it and it works. Please just let the code reviewers who have actually been assigned to this do their job, and stop wasting energy talking about a feature that has now been coded, mostly by me, not claude.

beqabeqa473 (approved):

beqabeqa473:
Seems ok. Icons would be good to have.

amirmahdifard:
@trypsynth I will look at the dif tomorrow. Coding with ai or yourself is no diffrence if you conferm that it's standard that how any other human would code this (even myself). The correct implementation for this would be adding the options male, nutral and female to arrays.xml, And Use them correctly in services where the connection is made and the client is logged in to the server, and in settings activity where it's being changed. If your implementation is this? I don't see any problem on it.

trypsynth:
@beqabeqa473 thanks, I'm fully blind so might not be the best to generate an icon but I can try to do it, or someone else can. Thoughts?
@amirmahdifard Please avoid commenting on the quality of my pull requests until at the very least after you've reviewed the code in them, thanks.

beqabeqa473:
Me too, but showing male for neutral may have false-positive for sighted people.

trypsynth:
@beqabeqa473 Done by slightly modifying the existing male icon, admittedly with the help of an AI assistant to write a python script using PIL and a sighted person to confirm it looks okay.

amirmahdifard:
@trypsynth hi. I have reviewed this pr, a fue things I want to say.
Why did you added gender migration etc? please just remove and rip out all old gender checkbox related things, and start this new one fresh. This reduces code and users can go to their settings and set their gender directly once.
Also, thanks so much for adding new icons for neutral gender. Can you add more please? Currently, sited users doesn't have any idea of so many things in teamtalk android that I have fixed and added for blind users with content description. So if you can, add a new icon for channel operator, a new one for administrator. Thanks.
While we are at this adding arrays.xml files, it is the best time to add app languge selection to general pref. Not being able to switch langs is very annoying for teamtalk, I wanted to implement it months ago, but the pull requests are being merged very sloly.

bear101:
Thanks @trypsynth! Look good.

I also think the migration is a bit unnecessary but it's ok. In a few releases we can remove it.

PR #3421, opened 2026-09-02
Android client. Let screen readers reach the action bar first

trypsynth:
## Problem

On every screen, TalkBack reaches "Navigate up", the screen title and the menu only *after* swiping through every control the screen contains, and then wraps around to the start. On Preferences that means all seven headers come before the title; on the server list, the whole server list and three version labels.

## Root cause

AppCompat draws the decor action bar and the window content on top of each other. Dumped from a device on the server list screen:

```
content              [0,0][1080,2424]   <- spans the whole screen
action_bar_container [0,0][1080,277]
```

Accessibility services order siblings by their bounds, not by child order. Both start at `top=0, left=0`, so the comparison falls through to size and the larger view goes first. The full-screen content wins every time.

Two things that look like fixes are not:

- `setAccessibilityTraversalBefore` pointing the action bar at `android.R.id.content` — the target is a non-focusable full-screen container, and nothing changes.
- Reordering the children so the action bar comes first — I confirmed the reorder applies and survives layout, and the traversal order is still unchanged, because it is derived from the bounds.

For the action bar to come first its bottom has to be above the content's top, which the decor action bar never is.

## Change

Give the affected activities `AppTheme.NoActionBar` and put a `Toolbar` above the content instead. A toolbar occupies a band of its own, so the two no longer overlap and are ordered by position.

`AccessibilityHelper.installToolbar()` wraps whatever content view the activity has already set, so this stays in one place instead of an edit in all ten layouts. The activities already had a common call site for it, next to where they set up edge-to-edge insets. `PreferencesActivity` goes through `getDelegate()`, being a `PreferenceActivity`.

`AudioCodecActivity` deliberately keeps the decor action bar: it uses `ActionBar.NAVIGATION_MODE_TABS`, and a toolbar-backed action bar throws on `setNavigationMode()`. Converting it means replacing the tabs with a `TabLayout`, which felt like a separate change. Its swipe order is still wrong. `WebLoginActivity` has no action bar at all and is untouched.

## Testing

Verified on a Pixel 9a, by dumping the accessibility tree before and after and by swiping with TalkBack.

Preferences, which was the worst case, now reads:

```
 3  toolbar          [0,53][2424,160]
 4  ImageButton      d=Navigate up
 5  TextView         t=Preferences
 6  prefs_container  [107,213][2272,1080]
 7    ... General, Sound Events, Text-To-Speech, ...
```

Same on the General sub-screen (Navigate up, then "General", then Nickname/Status message/Gender) and on the server list (title, the two action buttons, More options, then the search box and the list). The toolbar band and the content no longer overlap, which is what makes the ordering work.

Every converted screen has since been walked with TalkBack on the device: the server list, server entry, main window, channel properties, user properties, text messages, online users, stream media, and Preferences including its sub-screens. Navigate up and the title come first on each of them, and nothing looked out of place.

### DIFF
--- a/ChangeLog.txt
+++ b/ChangeLog.txt
@@ -31,6 +31,7 @@ Android Client
 - Fixed bug where not all selected files could be uploaded
 - Fixed bug where sounds were not muted during outgoing calls
 - Fixed channel list items not being exposed as clickable for accessibility services
+- Fixed screen readers reaching the action bar only after every other control on the screen
 - Minimum supported Android version is now 6.0 (Marshmallow)
 iOS Client
 - User interface has been redesigned using a more modern UI toolkit
--- a/Client/TeamTalkAndroid/src/main/AndroidManifest.xml
+++ b/Client/TeamTalkAndroid/src/main/AndroidManifest.xml
@@ -50,6 +50,7 @@
 
         <activity
             android:name=".ServerListActivity"
+            android:theme="@style/AppTheme.NoActionBar"
             android:icon="@drawable/teamtalk_blue"
             android:launchMode="singleTop"
             android:exported="true">
@@ -70,12 +71,14 @@
 
         <activity
             android:name=".ServerEntryActivity"
+            android:theme="@style/AppTheme.NoActionBar"
             android:label="@string/title_activity_server_entry"
             android:exported="false">
         </activity>
 
         <activity
             android:name=".MainActivity"
+            android:theme="@style/AppTheme.NoActionBar"
             android:configChanges="orientation|screenSize"
             android:label="@string/title_activity_main"
             android:launchMode="singleTop"
@@ -85,12 +88,14 @@
 
         <activity
             android:name=".ChannelPropActivity"
+            android:theme="@style/AppTheme.NoActionBar"
             android:label="@string/title_activity_channel_prop"
             android:exported="false">
         </activity>
 
         <activity
             android:name=".StreamMediaActivity"
+            android:theme="@style/AppTheme.NoActionBar"
             android:label="@string/title_activity_stream_media"
             android:exported="false">
         </activity>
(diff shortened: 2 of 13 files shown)


amirmahdifard:
@trypsynth i have already fixed this bug by taking a simpler path, I'm just waiting for a sited user to confurm it for me as well.

trypsynth:
@amirmahdifard what was your path, and is there a PR opened for it? I've already had a sighted person confirm it looks the same.

amirmahdifard:
@trypsynth Here's my fix for this issue.
https://github.com/BearWare/TeamTalk5/pull/3405
It works perfectly for screen readers, I just need a sited user to conferm it for me that anything won't break for them. If there is no problem with it, That single small change to edg to edg would fix this and we no longer need to change anything else. Can you try it if you have any sited user around?

trypsynth:
I'll try it yeah, but why is that pull request closed? That's why I missed it and did this fix myself.

beqabeqa473:
need to investigate #3405 for visual side, as that change is more compact.

trypsynth:
I had someone look at it for me and they said it looks pretty much the same, so feel free to merge that instead. I just don't know why it was closed in the first place.

amirmahdifard:
@trypsynth I already told the reason! I closed it because I didn't find any sited user to test it for me. But today, There will be someone who will test it for me and conferm it, Then we can close this pr and merge my pr.
I always try to fix issues in standard ways instead of taking another path. I figgered out that I can simply fix this issue by properly supporting and fixing the edg to edg over the entire content view of the entire app raather than the current activity leyowt! That was the cause for this issue!

amirmahdifard:
@trypsynth My fix for this issue had only a single problem, That I've fixed it and provided the final fix here.
https://github.com/BearWare/TeamTalk5/pull/3437
Now, You can close this pr. Thanks for your try anyway.

trypsynth:
Nice, much smaller diff.

PR #3422, opened 2026-09-02
Android client. Offer options menu items as accessibility actions

trypsynth:
## Problem

Reaching an item hidden behind the overflow button with a screen reader means opening the menu and then swiping through it. Every one of them is at least two gestures away, and on the server list the menu holds seven items.

## Change

`MenuActionsHelper.addMenuActions()` publishes the items behind the overflow button as accessibility actions on it, so a screen reader can offer them directly from its actions gesture. Nothing about the menu itself changes: it looks and behaves the same, and this is invisible to anyone not using an accessibility service.

Items the menu shows as buttons of their own are left out, since those are already a swipe away and repeating them would only pad the list. How many of them there are depends on how much room the menu has, so on the server list:

| | Shown as buttons | Offered as actions |
| --- | --- | --- |
| Portrait | New Server, Refresh | Enter Join Code, Import from a .tt File, Export server list, Settings, Exit |
| Landscape | New Server, Refresh, Import from a .tt File, Export server list | Enter Join Code, Settings, Exit |

Neither the menu nor the buttons are read up front. Both are looked up when an accessibility service asks for the button's node, which means items that appear, disappear or are renamed in `onPrepareOptionsMenu()` — something `MainActivity` does a lot of — are picked up on their own, the list follows the screen it is actually on, and the actions cannot drift out of sync with the menu.

Selecting an action routes through `onOptionsItemSelected()`, the same entry point the menu itself uses, so there is no second copy of the handling to keep in step.

The overflow button is found by walking to the `ActionMenuView` and taking its icon child, the items themselves being text views. That works whether the activity has a decor action bar or a toolbar, so this is independent of #3421 and the two can merge in either order.

`AudioCodecActivity` is not included: its `onCreateOptionsMenu()` inflates nothing.

## Testing

Verified on a Pixel 9a with TalkBack. Instrumenting the delegate confirms it is installed on the overflow button, consulted on every node query, and that the list tracks the orientation:

```
portrait   buttons=[new server, refresh]
             Enter Join Code / Import from a .tt File / Export server list / Settings / Exit
landscape  buttons=[new server, import from a .tt file, export server list, refresh]
             Enter Join Code / Settings / Exit
```


trypsynth:
@bear101 any thoughts on this PR? Just noticed it got left sitting when all the other earlier android PR's got merged, so wanted to make sure it didn't fall off the radar.

bear101:
> @bear101 any thoughts on this PR? Just noticed it got left sitting when all the other earlier android PR's got merged, so wanted to make sure it didn't fall off the radar.

I'm hoping @beqabeqa473 or @amirmahdifard would have an opinion on whether this feature is useful. I'm only familiar with VoiceOver on iPhone.

amirmahdifard:
if my opinion is wanted, i'd say this is useless. Why would you want to add accessibility actions to globel actions? please work on adding accessibility actions to channel items, user items, and file items, and user items in online users activity, just like there's accessibility actions for server items in the server list. The popup menu items for these cases i mentioned, should be exposed as accessibility actions. I was planning to do it, but if you want to do it, then i'm happy to hand it to you.

trypsynth:
Thanks, and you're right that the per-item menus are the bigger win. I checked and `ServerListActivity` already does this for server entries via `ViewCompat.addAccessibilityAction()`, so there's a pattern to follow and channel, user and file items are the obvious gap.

One thing though, I don't think these compete. #3422 is about the overflow button in the action bar, which on the server list holds seven items that are each two gestures away with a screen reader. The per-item menus you're describing are a different surface. Doing one doesn't rule out the other.

If you were already planning the per-item work then it's yours and I'd rather not duplicate it. Happy to take a piece if you want to split it, say channel and file items to me and users to you, or whatever split suits you.

@bear101 for what it's worth, I use this pattern myself and it works well in practice. Paperback, my document reader for Android, builds its top bar overflow items and their TalkBack actions from a single definition for exactly the reason behind this PR, so the menu and the actions can't drift apart. It's my main mobile app, going paid shortly, and the feedback on it from blind users has generally been good. On that basis I'd like to keep #3422, but it's your call.

amirmahdifard:
@trypsynth i haven't actually compiled this to test on my device to see how it looks like? but can you tell me where should the talkback focus be to have this actions? for example, do you focus the more options button, and then opening talkback menu to see these actions? or everywhere?

trypsynth:
As detailed in the PR description under "Change", the actions are attached directly to the overflow button itself; they are not global. You just focus the overflow button and swipe through the actions rotor, without needing to open the full TalkBack menu or the Android options menu.
I’d would suggest checking out the branch and building it locally when you get a chance; testing the code usually clears up how it works much faster than back-and-forth comments calling it useless.

amirmahdifard:
I have a fue more works to do with qt client, then i'll go back to teamtalk android, and i'll build and give a try to this.

bear101 (commented):
Isn't this a duplicate of #3447?

CoBC:
It's not a duplicate of #3447. However, I'm not convinced this change is worthwhile. The TeamTalk menu isn't that full. No other application I know adds menu options to TalkBack actions; it's not standard practice, and I doubt users will see any real benefit.

amirmahdifard:
I also don't see this nesiserry as I said before if my opinion is wanted on this. It's a bit complication for no reason.

trypsynth:
Fair enough. Paperback, my Android document reader, does this on its top bar, but I'll grant it isn't standard practice and three people who use TalkBack daily don't want it. Closing.

@amirmahdifard the offer stands on the per-item actions for channel, user and file items. That's the work everyone agrees is worth doing, and `ServerListActivity` already has the pattern to follow. Tell me which parts you want and I'll take the rest, or I'll stay out of it entirely if you'd rather do all of it.

PR #3379, opened 2026-07-28
Fix action bar in text messages activity and add copy all and clear

amirmahdifard:
Fixed the action bar in text message activity And correctly Added the globel actions.
this fixes https://github.com/BearWare/TeamTalk5/issues/3413

PR #3423, opened 2026-09-02
Add Copy All action to Android private messages

### LINKED ISSUE #3413: The Copy All option is missing from the private message history in the TeamTalk Android client
### Description

As you know, the Windows client has a feature called "Copy All" that lets users copy their entire message history to the clipboard with a single click. However, this feature is missing from the Android client.
At first, it might seem like adding this feature to the Android client isn't really necessary, but let's look at the issue from the perspective of a blind user.
Blind users rely on TalkBack or the CSR screen reader. There are many situations where we need to back up our private message history with someone and save those messages somewhere for later. Right now, the only way to do this is to copy each message individually. Imagine how difficult and frustrating that becomes when the history contains 50 or even 100 messages. We have to select the Copy option separately for every single message, or use TalkBack's or CSR's “Copy Last Spoken Phrase” feature. Either way, it's a pretty tedious process.
Having a “Copy All” option would make this much easier and allow users to back up their entire conversation quickly and efficiently.
Another important point is keeping the Android and Windows clients in sync and bringing useful Windows features over to Android. Me and many other users believe that this is one of those features that should be available on both platforms.
The good news is that amirmahdifard has already implemented this feature in the Android client and submitted a pull request to you. Unfortunately, it is still waiting to be reviewed.
If you merge this pull request, it will help make the Android client more feature-rich, more practical, and more consistent with the Windows client. All it would take is merging it a little sooner.
By doing so, you would be taking a positive step toward making the Android client easier and more convenient to use for blind users, while also helping us benefit from more of the useful features that are already available in the Windows client.

### Application

- [ ] qtTeamTalk
- [x] TeamTalkAndroid
- [ ] iTeamTalk
- [ ] TeamTalkServer

Antonielcriactive:
Closes #3413.\n\nAdds a Copy All action to the Android private message screen so users can copy the visible conversation history to the clipboard in one step. This is useful for TalkBack/screen-reader users who need to back up or reuse a full private message history without copying each message individually.\n\nChanges:\n- adds a Copy All action to the private message menu;\n- copies the adapter's current visible message history to the clipboard;\n- formats each copied line with timestamp, nickname, and message content.\n\nValidation:\n- git diff --check passed;\n- attempted ./gradlew.bat assembleDebug with Android Studio JBR and local SDK configured; build reached dependency resolution but failed because Library/TeamTalkJNI/libs/TeamTalk5.jar is not present in this checkout, which the Android README documents as an SDK-provided dependency.

### DIFF
--- a/Client/TeamTalkAndroid/src/main/java/dk/bearware/data/TextMessageAdapter.java
+++ b/Client/TeamTalkAndroid/src/main/java/dk/bearware/data/TextMessageAdapter.java
@@ -305,6 +305,16 @@ public class TextMessageAdapter extends BaseAdapter {
         }
     }
 
+    public String getAllMessagesText() {
+        StringBuilder text = new StringBuilder();
+        for (MyTextMessage msg : getMessages()) {
+            if (text.length() > 0)
+                text.append(System.lineSeparator());
+            text.append(msg.time).append(" ").append(msg.szNickName).append(": ").append(msg.szMessage);
+        }
+        return text.toString();
+    }
+
     private void copyToClipboard(Context context, String text) {
         ClipboardManager cm = (ClipboardManager) context.getSystemService(Context.CLIPBOARD_SERVICE);
         ClipData clip = ClipData.newPlainText("message", text);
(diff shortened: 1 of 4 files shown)


AbolfazlSaeidifar:
https://github.com/BearWare/TeamTalk5/pull/3379
please left a comment on that post

amirmahdifard:
@Antonielcriactive I have already implemented this feature. Please don't use ai and before opening prs, first look at the open prs.

bear101:
Duplicate

PR #3424, opened 2026-09-02
Fix Android .tt file import and export

Antonielcriactive:
## Summary
- Use Android's document picker for `.tt` import and export instead of direct Downloads/storage access
- Preserve and export join codes plus trusted certificate settings from imported server entries
- Report an import error when a selected `.tt` file is empty or invalid
- Show the selected document display name after import/export
- Align Android `.tt` read/write behavior with the shared desktop client format by writing `nickname` in `<clientsetup>` and reading it from there, while keeping legacy `<auth>` nickname/status message support

## Validation
- `git diff --check`
- `./gradlew.bat assembleDebug --console=plain --no-daemon`

Note: the Android build was validated after placing the SDK-provided `TeamTalk5.jar` and JNI libraries in `Library/TeamTalkJNI/libs`.


amirmahdifard:
@Antonielcriactive tt import and export is working fine in android client.

trypsynth:
@amirmahdifard not here, it crashes on some .tt files, but not with this change.

amirmahdifard:
@trypsynth Can you give some steps to reproduce? because here when I export and import tt files it doesn't crash.

bear101 (changes requested):
Please inform how to the reproduce the crash issue.

amirmahdifard:
@bear101 I can't reproduce a crash here. TT import and export is working fine. I don't know what this pr is trying to fix

PR #3425, opened 2026-09-03
Qt client. Handle user joins before channel tree rebuild

computertech1012:
Hi,

This fixes a Qt client crash that can occur when a user-join event arrives before the corresponding channel item has been added during a channel-tree rebuild.

`ChannelsTree::slotUserJoin()` assumed that `getChannelItem()` always returned a parent and relied on `Q_ASSERT(parent)`. In a Release build the assertion is disabled, so the null parent was passed to `getUserIndex()` and dereferenced by `QTreeWidgetItem::childCount()`.

The failure is deterministic with this event order:

1. Construct `ChannelsTree`.
2. Deliver `slotUserJoin()` for a user in channel 42.
3. Deliver `slotAddChannel()` for channel 42 afterward.

On unmodified master, Debug stops at `Q_ASSERT(parent)`. RelWithDebInfo and Release fail with a `0xC0000005` read at address `0x50` through:

```
QTreeWidgetItem::childCount()
ChannelsTree::getUserIndex()
ChannelsTree::slotUserJoin()
```

The fix defers a join when its channel item is not ready and replays it when that channel is added. Deferred entries retain later user updates and are removed after insertion, user leave, logout, or tree reset. The map is keyed by user ID, so repeated events remain bounded and cannot create duplicate rows.

Tests:

- Focused harness calling the real `ChannelsTree` methods.
- 19 checks covering replay, updates, leave/logout/reset cancellation, duplicate joins, multiple channels, alphabetical ordering, rename, removal, and repeated rebuilds.
- Debug, RelWithDebInfo, and Release configurations all passed.
- Long Release run: 300 rebuild cycles, 300 users per cycle, 157,800 event operations, and zero Qt warnings.
- Working set was 44,232 KB before, 55,384 KB after, and 57,784 KB peak; tripling the stress cycles did not produce proportional growth.

The standalone harness and complete logs are available if they would be useful for review.

Thanks for reading!

### DIFF
--- a/Client/qtTeamTalk/channelstree.cpp
+++ b/Client/qtTeamTalk/channelstree.cpp
@@ -275,6 +275,7 @@ void ChannelsTree::resetChannels()
     delete topLevelItem(0);
     m_channels.clear();
     m_users.clear();   
+    m_pendingusers.clear();
     m_videousers.clear();
     m_desktopaccess_users.clear();
     m_blinkhand_users.clear();
@@ -1007,7 +1008,9 @@ QTreeWidgetItem* ChannelsTree::getUserItem(int userid) const
 
 int ChannelsTree::getUserIndex(const QTreeWidgetItem* parent, const QString& name) const
 {
-    Q_ASSERT(parent);
+    if(!parent)
+        return -1;
+
     int count = parent->childCount();
     int i;
     for(i=0;i<count;i++)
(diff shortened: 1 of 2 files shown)


bear101:
What issue are you trying to solve?

computertech1012:
Hi,

I was trying to prevent a Qt client crash if `CLIENTEVENT_CMD_USER_JOINED` reached `ChannelsTree::slotUserJoin()` before the corresponding channel item existed.

After your question, I checked the SDK event contract again. Login posts the channel events before the user events, and channel creation sends `CLIENTEVENT_CMD_CHANNEL_NEW` followed by `CLIENTEVENT_CMD_USER_JOINED`. My focused harness reproduced the null dereference by forcing an event order that the SDK does not generate.

I do not have a real client reproduction that violates that ordering, so the harness alone does not justify this change. I am closing the PR. Thanks for catching that, @bear101.

Thanks for reading!

Issue #48, opened 2015-03-17
Bandwidth quota system on user account


bear101:
Limit amount of data traffic can be used on a single user account (to prevent abuse)

amirmahdifard:
this is really could bee good if you done it at some point

amirmahdifard:
@bear101 in these old tikits, i really agree with this

PR #3442, opened 2026-09-09
Add per-account login-delay overrides and Qt settings

computertech1012:
Hi,

I've updated this PR to follow @bear101's suggestion: add an account-specific login-delay override to AbusePrevention instead of automatically exempting every administrator.

The server-wide login delay stays as the default. In the Qt user account editor, the account can use the server default, disable its login delay, or specify a custom interval in milliseconds. Administrators follow the same rules unless explicitly exempted.

The account field uses:

- `0`: inherit the server setting.
- `-1`: disable the delay for this account.
- Positive values: a custom delay in milliseconds.

Accounts inheriting the default still share the existing per-IP counter. A positive override uses its own account/IP counter, so a short override does not inherit another account's longer delay. Exempt accounts do not refresh anybody else's counters. Password verification, bans and the other abuse limits are unchanged.

This includes the server enforcement and cleanup, XML persistence and wizard, SDK conversions, C/Java/.NET/Python bindings, Qt controls, documentation and regression tests. Older clients can edit accounts on the standalone server without clearing an existing override. Invalid or overflowing values cannot accidentally create an exemption.

Compatibility note: extending AbusePrevention changes the native SDK structure layout. The updated DLL, headers and bindings must be rebuilt and shipped together; it is not a drop-in DLL for older compiled clients. Older servers do not enforce the new override. SDK-hosted applications still manage their own account persistence; that limitation is documented separately. I have left the release/version assignment to the maintainer.

Updated against current master. Verification on native Windows, without WSL:

- Built the official Qt client, Standard SDK/JNI, and Standard and Professional server binaries.
- Ran all 12 disposable loopback protocol tests three times: 36 passes.
- Ran the three focused Java login-delay tests five times: 15 passes, including the existing default-delay regression and the new account-field round trip.
- Passed all nine focused ServerXML/account serialization cases: 208 assertions.
- Tested 24 off-screen Qt dialog combinations, including administrator/default accounts, create/edit/read-only modes, setting round trips and accessible read-only values.
- Compiled the .NET binding and verified .NET/Python structure sizes against the rebuilt native library.

These are login-delay and account-editor tests, not a full audio qualification. The Windows build used the available native dependencies with WebRTC, FFmpeg and Opus-tools disabled. The full Catch target exposes an existing Opus/Opus-tools feature-guard build issue in that configuration, so the XML cases were built and run in a focused harness against the same server objects. The Professional server was build-tested, not started for licensed runtime testing. No production server was used or changed.

The standalone protocol test harness is kept outside this PR. The Java login-delay tests generate fresh test credentials at runtime rather than embedding fixed passwords. The temporary XML fixture that triggered the Python CodeQL warning has been removed from the proposed changes; it contained no real server credentials. Setup, accounting and compatibility details are in `Documentation/TeamTalkSDK/account-login-delay.md`.

Thanks for reading!

### DIFF
--- a/Library/TeamTalkLib/teamtalk/server/ServerNode.cpp
+++ b/Library/TeamTalkLib/teamtalk/server/ServerNode.cpp
@@ -2672,7 +2672,8 @@ ErrorMsg ServerNode::UserLogin(int userid, const ACE_TString& username,
     {
     case TT_CMDERR_SUCCESS :
     {
-        if (LoginsExceeded(*user))
+        // Administrators are exempt from the per-IP login delay.
+        if ((useraccount.usertype & USERTYPE_ADMIN) == 0 && LoginsExceeded(*user))
             return TT_CMDERR_COMMAND_FLOOD;
         break;
     }
--- a/Library/TeamTalkJNI/test/dk/bearware/TeamTalkTestCase.java
+++ b/Library/TeamTalkJNI/test/dk/bearware/TeamTalkTestCase.java
@@ -3459,6 +3459,44 @@ public abstract class TeamTalkTestCase extends TeamTalkTestCaseBase {
     }
 
     @Test
+    public void testAdminLoginDelayExemption() {
+        String ADMINUSERNAME = "tt_admin_delay";
+        String PASSWORD = "tt_test", NICKNAME = "jUnit - " + getTestMethodName();
+        int USERRIGHTS = UserRight.USERRIGHT_CREATE_TEMPORARY_CHANNEL | UserRight.USERRIGHT_MULTI_LOGIN;
+
+        TeamTalkBase ttadmin = newClientInstance();
+        connect(ttadmin);
+        login(ttadmin, ADMIN_NICKNAME + " - " + getTestMethodName(), ADMIN_USERNAME, ADMIN_PASSWORD);
+
+        ServerProperties srvprop = new ServerProperties();
+        assertTrue(ttadmin.getServerProperties(srvprop));
+        int orgValue = srvprop.nLoginDelayMSec;
+
+        try {
+            srvprop.nLoginDelayMSec = 60000;
+            assertTrue(waitCmdSuccess(ttadmin, ttadmin.doUpdateServer(srvprop), DEF_WAIT));
+
+            UserAccount adminAccount = new UserAccount();
+            adminAccount.szUsername = ADMINUSERNAME;
+            adminAccount.szPassword = PASSWORD;
+            adminAccount.uUserType = UserType.USERTYPE_ADMIN;
+            adminAccount.uUserRights = USERRIGHTS;
+            assertTrue(waitCmdSuccess(ttadmin, ttadmin.doNewUserAccount(adminAccount), DEF_WAIT), "create admin account");
+
+            TeamTalkBase ttadmin1 = newClientInstance();
+            TeamTalkBase ttadmin2 = newClientInstance();
+            connect(ttadmin1);
+            login(ttadmin1, NICKNAME, ADMINUSERNAME, PASSWORD);
+            connect(ttadmin2);
+            login(ttadmin2, NICKNAME, ADMINUSERNAME, PASSWORD);
+        }
+        finally {
+            srvprop.nLoginDelayMSec = orgValue;
+            assertTrue(waitCmdSuccess(ttadmin, ttadmin.doUpdateServer(srvprop), DEF_WAIT));
+        }
+    }
+
+    @Test
     public void testLoginAttempts() {
 
         TeamTalkBase ttadmin = newClientInstance();


amirmahdifard:
@computertech1012 Thanks this is how it should behave

bear101:
If you want this feature then I would suggest adding login delay to AbusePrevension. It's a user account specific setting. Then AbusePrevension could override ServerProperties setting.

amirmahdifard:
@bear101 yes! i love your suggestion! if login delay is removed from server prop and added to per account abuce prevention, then we can set login delay for only the users that we want instead of adding for all users! this would be grate! the qt client should also be changed to reflect configgering this. @computertech1012 do you like the idea?

computertech1012:
Hi @bear101 and @amirmahdifard,

Yes, I think the account-specific setting is a better approach. I've reworked this to add the override to AbusePrevention and the Qt user account editor instead of exempting every administrator automatically.

I'm keeping the server-wide setting as the default, so existing accounts do not lose their protection. An account can inherit that setting, disable its own login delay, or use a custom delay. Administrators follow the same rules unless an exemption is explicitly configured.

I'm finishing the compatibility and regression tests before pushing the changes. This also covers older clients editing accounts without clearing the override.

Thanks for reading!

computertech1012:
Hi @bear101 and @amirmahdifard,

I've pushed the changes in f872549a and brought the branch up to date with master. The Qt account editor now lets you use the server login delay, disable it for that account, or choose a custom interval. The server-wide default is still there, and administrators are no longer automatically exempt.

I also fixed counter cleanup for account overrides, preserved existing overrides when older clients edit accounts, and added validation so malformed values cannot accidentally disable the delay. The SDK bindings, XML configuration, server wizard and documentation are updated too.

I built the official Qt client and server binaries on Windows and ran the focused tests: 36 protocol passes, 15 Java test passes, 208 XML/serialization assertions and 24 off-screen Qt dialog combinations. The updated PR description lists the build configuration and remaining qualification limitations. This does change the SDK structure layout, so the updated library and bindings need to be built together.

Update: I've removed the standalone temporary XML test fixture from the PR and kept that harness local. The affected Java login-delay tests now generate fresh credentials at runtime, and the unnecessary dummy password in the new XML parsing test is removed. No real server credentials were included. No CodeQL suppression or workflow exclusion was added.

Could you please have another look when you have time?

Thanks for reading!

amirmahdifard:
@computertech1012 I think that
@bear101 
Meant that like flud protection, remove this entirely from server prop, And only add it to account abuce prevention. We don't have command flud protection in server prop, Also see this issue that bear101 opened himself.
https://github.com/BearWare/TeamTalk5/issues/48
so may be you can also move these out of server prop and add it per accounts, so that long standing issue can also be fixed while we are at it. Thanks.

computertech1012:
Hi,

@amirmahdifard, I understood @bear101's suggestion to mean keeping the server login delay as the default and allowing each account to override it. That is what the current changes do: use the server setting, disable the delay for an account, or set a custom delay.

@bear101, could you please confirm whether you want both settings with that override behavior, or whether you want the server-wide setting removed entirely? I want to make sure I understood your request before changing how existing accounts behave.

I also looked at #48. That is about bandwidth quotas per account, which seems separate from login delay. I think that would be better handled in a separate change rather than adding it to this PR as well.

Thanks for reading!

Library/TeamTalkLib/bin/ttsrv/ServerXML.cpp
```diff
@@ -1402,11 +1403,27 @@ namespace teamtalk{
...
         }
 
+        user.abuse.login_delay = 0;
         const XMLElement* abuseElement = userElement->FirstChildElement("abuse-prevention");
         if(abuseElement != nullptr)
         {
             GetInteger(abuseElement, "commands-limit", user.abuse.n_cmds);
             GetInteger(abuseElement, "commands-interval-msec", user.abuse.cmd_msec);
+            std::string delay;
+            if (GetString(abuseElement, "login-delay-msec", delay))
```
bear101: Use GetInteger

Library/TeamTalkLib/teamtalk/server/ServerNode.cpp
```diff
@@ -1286,29 +1286,24 @@ void ServerNode::IncLoginAttempt(const ServerUser& user)
     }
 }
 
-bool ServerNode::LoginsExceeded(const ServerUser& user)
+bool ServerNode::LoginsExceeded(const ServerUser& user, const UserAccount& account)
```
bear101: Why is `UserAccount` parameter necessary when you can extract it from `ServerUser` class?

bear101 (changes requested):
Please only use AI for inspiration. Much of the AI generated code in this PR is only understandable by a computer. It is unreadable to humans.

computertech1012:
Hi,

@bear101, I've simplified the login-delay code in 6912463e and pushed it for review. The XML reader now uses GetInteger, the nested conditions are plain if/else branches, and the default and account-specific counters are separate maps. The custom digit-by-digit parser is replaced with standard integer conversion.

There is one ordering detail behind the UserAccount argument: at this point ServerUser has not had SetUserAccount called yet, and calling it early would mark the connection authorized before the remaining login checks. I changed the function to take just the IP address and authenticated account, documented that ordering, and added a regression check that a throttled administrator cannot list accounts.

GetInteger has an optional strict mode for this setting so a value such as -1oops cannot turn into an exemption. Strict parsing returns false on invalid input; existing callers retain their previous behavior. I also fixed counter cleanup on StopServer and added a regression test that failed before that fix and passes afterward.

The native Windows server and Qt builds passed. The final revision passed all 15 loopback protocol tests, three Java tests repeated five times, and 251 assertions across 11 XML/counter tests. The Qt editor also passed 24 off-screen account-dialog combinations. These are focused checks, not a claim that every platform or audio feature has been qualified.

Thanks for reading!

Library/TeamTalkLib/settings/Settings.cpp
```diff
@@ -317,15 +317,29 @@ namespace teamtalk {
...
+    bool XMLDocument::GetInteger(const tinyxml2::XMLElement* parent, const string& szName, int& nValue, bool strict) const
     {
         const tinyxml2::XMLElement* item = parent->FirstChildElement(szName.c_str());
 
         if(item != nullptr)
         {
             string s;
             GetElementText(item, s);
-            nValue = std::stoi(s);
+            size_t end = 0;
```
bear101, line 328: Why is this change necessary?

Library/TeamTalkLib/teamtalk/server/ServerNode.cpp
```diff
@@ -1286,29 +1294,32 @@ void ServerNode::IncLoginAttempt(const ServerUser& user)
     }
 }
 
-bool ServerNode::LoginsExceeded(const ServerUser& user)
+bool ServerNode::LoginsExceeded(const ACE_TString& ipaddr, const UserAccount& account)
 {
     ASSERT_SERVERNODE_LOCKED(this);
 
-    if (m_properties.logindelay == 0)
+    if (account.abuse.login_delay == -1)
```
bear101: make a `static constexpr int LOGIN_DELAY_DISABLED = -1` in `teamtalk::Abuse` and check against this instead.

Library/TeamTalkLib/teamtalk/server/ServerNode.cpp
```diff
@@ -3502,14 +3515,27 @@ ErrorMsg ServerNode::UserListUserAccounts(int userid, int index, int count)
...
-ErrorMsg ServerNode::UserNewUserAccount(int userid, const UserAccount& regusr)
+ErrorMsg ServerNode::UserNewUserAccount(int userid, UserAccount regusr, bool preserveLoginDelay)
 {
     GUARD_OBJ(this, Lock());
 
     serveruser_t const user = GetUser(userid, nullptr);
     if (!user)
         return ErrorMsg(TT_CMDERR_USER_NOT_FOUND);
 
+    if (regusr.abuse.login_delay < -1)
```
bear101: This check is unnecessary.

Library/TeamTalkLib/teamtalk/server/ServerUser.cpp
```diff
@@ -869,7 +870,33 @@ ErrorMsg ServerUser::HandleNewUserAccount(const mstrings_t& properties)
         account.userrights |= USERRIGHT_TEXTMESSAGE_CHANNEL;
     }
 
-    return m_servernode.UserNewUserAccount(GetUserID(), account);
+    // Legacy clients omit the third cmdflood value. Preserve their stored override.
```
bear101: This code doesn't make any sense.

Library/TeamTalkLib/teamtalk/server/ServerUser.cpp
bear101: GetProperties(UserAccount) should be ended tohandle login-delay. Use existing command parameter `TT_LOGINDELAY` and default it to -1.

Library/TeamTalkLib/teamtalk/Common.h
```diff
@@ -225,13 +225,16 @@ namespace teamtalk {
     {
         int n_cmds = 0;
         int cmd_msec = 0;
+        // 0 inherits the server delay, -1 disables it, positive values override it.
+        int login_delay = 0;
```
bear101, line 230: Default to -1

Library/TeamTalkLib/teamtalk/Common.h
```diff
@@ -241,6 +244,7 @@ namespace teamtalk {
                 n_cmds = flood[0];
                 cmd_msec = flood[1];
             }
+            login_delay = flood.size() >= 3 ? flood[2] : 0;
```
bear101: -1 instead of 0

bear101 (changes requested):

computertech1012:
Hi,

@bear101, I have pushed 20cb6f61 with the changes I could make without changing the default protection. Accounts now use the existing `TT_LOGINDELAY` property, parsed in `GetProperties(UserAccount&)`, instead of adding a third `cmdflood` value. I added `Abuse::LOGIN_DELAY_DISABLED` and removed the redundant check from `UserNewUserAccount`. Older account editors that omit the property still preserve an existing override.

Could you please clarify the requested default of `-1`? At present, `0` inherits the server delay and `-1` disables the delay for that account. Making `-1` the default would exempt new accounts and accounts loaded without this XML setting. Did you intend that, or did you want to change what `-1` means? I have kept inheritance as the default until that is clear, including for zero-initialized SDK account structures.

The strict `GetInteger` option is there because `std::stoi("-1oops")` otherwise returns `-1`, which would turn malformed XML into an exemption, and out-of-range values throw. Existing callers keep their previous behavior. The wire parser also checks the integer range before assignment so an oversized value cannot narrow to `-1`. Invalid values are covered by regression tests.

I reviewed the revised paths again and reran the native Windows server, DLL and Qt builds, 15 loopback protocol tests, 267 assertions across 11 XML/counter tests, three Java tests five times, and 24 off-screen Qt account-dialog combinations. These passed. This is focused Windows verification, not a claim that every platform's CI is green.

Thanks for reading!

PR #3444, opened 2026-09-15
Qt client: Switch back to the selected sound device when it is plugged in again

trypsynth:
Unplug a headset on Windows and plug it back in, and the client keeps playing through whatever device it fell back to. It only returns to the selected device after pressing Refresh Sound Devices in Preferences. Other applications follow the hardware on their own.

Two things caused this.

**The plug-in event was ignored.** `CLIENTEVENT_SOUNDDEVICE_ADDED` only printed a status message, and `CLIENTEVENT_SOUNDDEVICE_UNPLUGGED` only logged to `qDebug`. Neither reinitialized sound, so a device coming back was never picked up. Only `CLIENTEVENT_SOUNDDEVICE_REMOVED` and the new-default events did anything.

**The fallback could land on the wrong device.** When the stored device UID no longer resolves, `getSoundInputFromUID` / `getSoundOutputFromUID` returned the stored device *index*. Indices shift as endpoints appear and disappear, so that index often refers to a different device by then, and initialization succeeds on it. That is why playback moved to the speakers instead of the default device.

## Changes

The seven `CLIENTEVENT_SOUNDDEVICE_*` cases now go through one handler, `clienteventSoundDeviceChange`, which decides whether sound needs reinitializing:

- added: the device matches the one selected in settings, so it just came back
- removed or unplugged: the device is the one currently in use, so it needs replacing
- new default input/output, including the communication-device variants: settings are on the default device

Devices are matched on `getSoundDeviceUID`, falling back to the device name so sound systems that report no device ID still match.

A single plug or unplug produces a burst of notifications from Windows, often one per role plus a state change. The restart is therefore deferred by 500 ms and re-armed on each event, so a burst collapses into one reinitialization. The timer also holds off while a modal dialog is open, so the sound system is not restarted underneath the loopback test in Preferences.

`getSoundInputFromUID` and `getSoundOutputFromUID` now fall back to the default device when the stored UID cannot be resolved, instead of reusing a stale index.

## Notes

No new translatable strings. The existing "New sound device available" message is now shown only when the device is not one the client is about to switch to on its own, since telling the user to refresh is misleading in that case.

## Testing

Built against Qt 6.10.3 with MSVC 2022 x64 and tested on Windows 11. With a specific output device selected in Preferences, unplugging moves playback to the default device and plugging back in returns it to the selected device within about half a second, with no manual refresh. The status bar reports the device on each switch.


Client/qtTeamTalk/mainwindow.cpp
```diff
@@ -2630,6 +2647,15 @@ void MainWindow::timerEvent(QTimerEvent *event)
         setMediaFilePosition();
         killLocalTimer(TIMER_CHANGE_MEDIAFILE_POSITION);
         break;
+    case TIMER_SOUNDDEVICE_CHANGE :
+        // leave the timer running so sound devices are not restarted while
+        // the user is testing them in the preferences dialog
+        if (!QApplication::activeModalWidget())
```
bear101, line 2650: Why is this necessary? Testing sound devices in Preferences dialog doesn't unplug devices or change default.

Client/qtTeamTalk/mainwindow.cpp
```diff
@@ -1669,19 +1669,65 @@ void MainWindow::clienteventUserAudioBlock(int source, StreamTypes streamtypes)
...
+    auto devid = getSoundDeviceUID(snddev);
+    if (devid.isEmpty())
+        return;
 
-void MainWindow::clienteventSoundDeviceRemoved(const SoundDevice& snddev)
-{
-    addStatusMsg(STATUSBAR_SOUND_DEVICE_DETECTED, tr("Sound device removed: %1.").arg(_Q(snddev.szDeviceName)));
+    // sound systems without a device ID use the device name as UID, so accept both
+    auto devname = _Q(snddev.szDeviceName);
+    auto matches = [&](const QString& uid) { return uid.size() && (uid == devid || uid == devname); };
```
bear101: Why does it match `uid` and `devname`?

Client/qtTeamTalk/mainwindow.cpp
trypsynth: You're right that the Preferences dialog itself never changes devices, and my comment was misleading about the reason. I've reworded it.

The case I had in mind is an external change landing while the dialog happens to be open, e.g. unplugging a headset while in the sound settings. What goes wrong then isn't corruption, it's quieter than that.

`RestartSoundSystem()` returns early if any stream is open:

```cpp
// SoundSystemBase.h:1017
if (m_input_streamers.size() || m_output_streamers.size() || m_duplex_streamers.size())
    return false;
```

A running loopback test holds streams in that same singleton, via `OpenDuplexStream()` / `OpenInputStream()` / `OpenOutputStream()` in `SoundLoopback.cpp`. So `TT_RestartSoundSystem()` inside `initSoundDevices()` returns false, the device list is never refreshed, and the reinit resolves IDs from the stale list. The timer has already been killed at that point, so there's no retry and the device change is dropped, which is the thing this PR is meant to fix.

Keeping the timer armed just defers it until the test is finished, and then it succeeds normally.

Happy to drop it if you'd rather keep the diff minimal. It's a narrow window and losing the event there is not fatal.

bear101 (commented):
This PR is extremely complicated. What issue are you trying to fix?

Client/qtTeamTalk/mainwindow.cpp
trypsynth: It shouldn't. I've removed the name comparison, it was dead code.

My reasoning was that `getSoundDeviceUID()` falls back to `szDeviceName` when `szDeviceID` is empty, so a stored UID could be either an ID or a name. But that fallback exists for backends that report no device ID, and it doesn't apply here. These events are Windows only, and the patched PortAudio fills in `uniqueID` for every Windows host API (`0001-Add-uniqueID-to-PaDeviceInfo-that-identifies-sound-d.patch` covers dsound, wasapi and wmme), so on Windows the UID is never a name and the extra comparison could never match.

Worth flagging while we're here, since it came out of checking this: the match only works for WASAPI selections. The notification carries an MMDevice endpoint ID, and that's also what `uniqueID` holds for WASAPI. For DirectSound and WMME the patch stores a host API GUID instead, which will never equal an endpoint ID. So a user who picked a DirectSound or WMME device won't get the automatic switch on replug.

That seems acceptable to me given WASAPI is what `TT_GetDefaultSoundDevices()` prefers, and mapping an endpoint ID onto a DirectSound GUID is a much bigger job. Happy to note it in a comment if you'd like, or to look at it separately.

trypsynth:
Sorry, I should have led with this rather than the mechanism.

**The bug:** on Windows, unplug a headset and plug it back in, and the client keeps playing through whatever device it fell back to. It only goes back to the device selected in Preferences once you press Refresh Sound Devices. Other applications follow the hardware on their own.

There's no issue filed for it. I ran into it myself and went looking. Happy to open one if you'd rather have the ticket first.

**Why it happens:** `CLIENTEVENT_SOUNDDEVICE_ADDED` only added a status message and `CLIENTEVENT_SOUNDDEVICE_UNPLUGGED` only wrote to `qDebug`, so a device coming back was never acted on.

**On the size.** Fair, and that's my fault rather than the change's. I folded the seven `CLIENTEVENT_SOUNDDEVICE_*` cases into one method, so the diff shows a block deleted and a block added and the actual delta is buried inside the move. It's four separable things:

1. act on the added and unplugged events, which is the fix
2. the seven cases moved into `clienteventSoundDeviceChange()`, a refactor with no behaviour change
3. the 500 ms debounce, because Windows sends a burst of notifications per plug, one per role plus a state change, and without it the sound system restarts several times in a row
4. `getSoundInputFromUID()` / `getSoundOutputFromUID()` falling back to the default device instead of a stale index

Number 4 is the one that isn't really about hot-plug, and it's shared with `preferencesdlg.cpp`. When the stored UID doesn't resolve, the old code returned the stored device *index*. Indices shift as endpoints appear and disappear, so by then that number often refers to a different device and initialization succeeds on it. That's why playback ended up on the speakers rather than just going quiet.

I'm happy to cut this back to 1 on its own, leave the cases where they are, and send 3 and 4 as separate PRs. Just say which way you'd prefer and I'll redo it. If you'd rather keep it in one piece, I can at least drop the refactor so the behaviour change is visible in the diff.

bear101:
I'm changing this PR to draft because I don't understand what the AI code is doing.

trypsynth:
You asked what issue I was fixing. I answered that on the 15th, in detail, with the mechanism and the affected code paths. In the same comment I offered three ways to reshape it and asked you to pick one: cut it to the minimal fix, drop the refactor so the delta is visible, or leave it as is.

Five days later it's moved to draft with "I don't understand what the AI code is doing."

That isn't review feedback. It names no function, no line, and no behaviour you disagree with, so there's nothing in it I can act on. If you'd answered the question I asked, all three changes would be merged by now.

On authorship, since you raised it: I wrote this, I read every line of it, and I tested it on Windows against a real headset, unplugging and replugging to confirm the device comes back. I also reproduced the stale-index bug that sends audio to the wrong device. That testing is in the PR body. If a specific line is unclear, name it and I'll explain it.

I'm splitting it regardless, because that part of your feedback was fair and I offered it myself:

1. Act on `SOUNDDEVICE_ADDED` and `SOUNDDEVICE_UNPLUGGED`, the actual fix, around fifteen lines with the case labels left where they are
2. The 500 ms debounce, since Windows sends a burst of notifications per plug
3. `getSoundInputFromUID()` / `getSoundOutputFromUID()` falling back to the default device instead of a stale device index, which is a separate bug in shared code

Going forward, it would save us both time if a PR you don't want in its current shape got one line saying what shape you do want. I'll take that over silence followed by a draft flag.

trypsynth:
Split as promised. Closing this in favour of:

- #3460, act on `SOUNDDEVICE_ADDED` and `SOUNDDEVICE_UNPLUGGED`, which is the fix itself. 21 lines added, 1 removed, case labels left where they are.
- #3461, fall back to the default device instead of a stale device index. Separate bug in code shared with `preferencesdlg.cpp`, and the reason audio ended up on the speakers rather than just stopping.

The third piece, debouncing the burst of notifications Windows sends per plug, is held back until #3460 lands, since it wraps the calls that PR adds and would otherwise conflict with its own sibling.

Both are tested on Windows 11. #3460 by unplugging and replugging headphones while listening to my own audio, and #3461 with a USB sound card and headphones both connected, headphones selected, confirming that unplugging moves playback to the Windows default rather than to the other device.

PR #3446, opened 2026-09-15
Android client. Submit join code dialog from the keyboard

trypsynth:
## Problem

"Enter join code" on the server list opens a dialog and brings the keyboard up automatically, but pressing Enter or Done does nothing. The only way to submit is to dismiss the keyboard and reach the OK button, which with a screen reader means several extra gestures for a dialog holding a single field.

This is the same problem as #3427 and #3428, in the one remaining single-line dialog that still had it.

## Change

Sets `IME_ACTION_DONE` on the input and clicks the positive button when the action fires, matching what #3427 did for the channel password dialog.

`setImeOptions` is called before the dialog is created, so the keyboard picks up the action when it first opens. The editor action listener is attached after `show()`, since it needs the dialog to reach `getButton()`.

`IME_NULL` is accepted alongside `IME_ACTION_DONE` so a hardware Enter key works too.

## Notes

The broadcast message dialog in `MainActivity` also has no IME action, but it is deliberately left alone: it is multi-line, so Enter should insert a newline rather than submit.

## Testing

Built and installed on a Pixel 9a. Typing a join code and pressing Enter or Done now submits and closes the dialog. Checked with TalkBack that the dialog reads the same as before, since the listener is attached after the dialog is shown.


bear101 (approved):

bear101 (changes requested):
Should MR use the same approach as #3447

trypsynth:
Closing in favour of #3447, which now covers this dialog through the same `DialogHelper` as the operator password prompts, per @bear101's review here and on #3447.

PR #3447, opened 2026-09-15
Android client. Submit join code and operator password dialogs from the keyboard

trypsynth:
## Problem

Three dialogs hold a single text field but ignore Enter and Done, so submitting means leaving the field and reaching the OK button. With a screen reader that is several extra gestures for a dialog with one input.

- "Enter join code" on the server list
- the operator password prompt on the main screen
- the same prompt on the online users screen

Same problem as #3427 and #3428.

## Change

Adds `DialogHelper.showWithImeSubmit()`, which sets `IME_ACTION_DONE` on the field and clicks the positive button when the action fires. All three call sites replace `alert.show()` with a call to it, so none of them carries its own listener.

The helper shows the dialog rather than taking a shown one, because the ordering matters: the keyboard reads the IME options when it opens, so they have to be set before the dialog is created, while the listener needs a created dialog to reach `getButton()`.

An overload takes a `showKeyboard` flag for dialogs that want the keyboard up immediately. Only the join code dialog needs it today. It previously did that itself between `create()` and `show()`, which is the seam the helper now owns.

`IME_NULL` is accepted alongside `IME_ACTION_DONE` so a hardware Enter key works too.

This supersedes #3446, which did the join code dialog on its own. Folded in here after @bear101 asked for both to use the same approach.

## Notes

The channel password dialog in `MainActivity` still has its own copy of this logic from #3427. It could move onto the helper now that the `showKeyboard` overload exists, but that is working code and not part of what this fixes, so it is left alone. Say the word if you want it converted.

The broadcast message dialog is deliberately untouched. It is multi-line, so Enter should insert a newline rather than submit.

## Testing

Built and installed on a Pixel 9a. All three dialogs submit and close on Enter or Done, and the join code dialog still raises the keyboard on its own.


bear101:
Should #3446 also be changed to use this dialog helper?

trypsynth:
Folded #3446 in here as you suggested, so the join code dialog now goes through the same helper. Added a `showKeyboard` overload for it, since that dialog raises the keyboard itself between `create()` and `show()` and the helper owns that seam now.

Pushed on top of your merge rather than rebasing, so the new commit is the only thing to look at since your last pass. Closing #3446.

Tested on a Pixel 9a: all three dialogs submit on Enter or Done, and the join code dialog still brings the keyboard up on its own.

bear101 (commented):
when I select "Grant Operator" in main activity then there's no keyboard showing up.

bear101:
Hm, it's the same issue on master branch. Is it an Android feature to not pop up the keyboard automatically when there's a dialog popup?

When I press "Join" on a channel the keyboard shows up automatically but when I press "Grant Operator" it does not.

trypsynth:
Correct, and it's pre-existing rather than something this PR changed. The join code dialog calls `setSoftInputMode(SOFT_INPUT_STATE_ALWAYS_VISIBLE)` between `create()` and `show()`, and the operator password dialogs never did. Android only raises the keyboard automatically when a dialog asks it to.

The helper already takes a `showKeyboard` flag for exactly this, so making the operator prompts behave like the join code one is a one-word change at each of the two call sites. Want me to include that here, or keep this PR to the Enter key fix and do it separately?

PR #3448, opened 2026-09-17
Enhance TeamTalkKit with core features and documentation

GruiaChiscop:
Evolves the `TeamTalkKit` Swift package from #3233 into the fuller version
I've been running on my fork since. The `TeamTalkClient` singleton becomes an
instantiable `TeamTalkSession`, and the thin wrapper around the raw C API
grows into: value-type snapshot models, a typed event stream, async command
wrappers with timeouts, and documentation/tests to match.

Scale note: of the ~20.7k added lines, **8191 are the vendored `TeamTalk.h`**
(the full C API header, needed so Swift can see declarations that weren't
previously exposed through `TeamTalkC.h`) — not authored code. The actual
Swift surface is the remaining ~12.5k lines across the 6 commits below.

## Changes (one commit per concern)

1. **`build(teamtalkkit): SPM manifest and vendored C wrapper`** —
   `Package.swift`, the `TeamTalkC` target (`TeamTalkC.c`/`.h` +
   vendored `TeamTalk.h`), swift-tools bumped to 6.0 for `.iOS(.v18)`
   (staying on the v5 language mode).
2. **`feat(teamtalkkit): TeamTalkSession, core value types, payload
   marshaling`** — replaces `TeamTalkClient.shared` with an instantiable
   `TeamTalkSession` that owns its native instance and drains its message
   queue; adds Swift option sets/enums for access rights, subscriptions,
   status and media state; an `AudioCodec` enum instead of the raw C union;
   `Payloads` helpers for the fixed C-array string fields.
3. **`feat(teamtalkkit): snapshot models and raw-struct extensions`** —
   Equatable/Hashable/Sendable snapshots for user, channel, server, admin,
   messaging, media and audio-codec data, built from the raw C structs.
4. **`feat(teamtalkkit): typed event dispatch`** — a `TeamTalkEvent` enum
   with associated values in place of raw `ClientEvent` + `TTMessage`
   polling, with resolvers that fill in what the raw message doesn't carry
   directly (e.g. looking up the affected user/channel).
5. **`feat(teamtalkkit): session commands and async wrappers`** — the
   session's command surface grouped by area (commands, audio, media,
   desktop/video, server state), plus `async` wrappers around the
   fire-command/await-reply pattern, each bounded by a timeout.
6. **`docs(teamtalkkit): guides, API audit, runnable example`** — README,
   GettingStarted/Advanced guides, an audit of the C surface against the
   Swift wrapper, and a small executable example that connects a session
   and prints events.


bear101 (commented):
If a Mac developer wants an AI generated TeamTalk library then I would suggest they generate the library themselves. I don't have any plans to maintain this AI generated code.

GruiaChiscop:
I'm sorry, but the package is not AI-generated. Only the documentation and the TeamTalkC files are, but the Swift Implementation is written by me, with the help of AI in some situations, but I am the one who wrote these lines of code. and the existing iTeamTalk already uses a version of TeamTalkKit. I simply extended the package to cover the whole SDK, like how java and python wrappers do.

bear101:
Java and Python cannot interact directly with C, but Swift can, so why do you want to write Swift wrappers for all TeamTalk structs and functions? It just adds more maintenance.

GruiaChiscop:
It’s simply a personal preference of mine and makes the API easier to use. It’s quite another thing to see and write Swift code everywhere. The C modules remain separate in a package that you don’t interact with. Allmost every  Swift application follow the same pattern: they separate the C libraries from the application wherever possible and use Swift modules.
Furthermore, as I’m pretty much the sole maintainer of iTeamTalk – because apart from me, no one else has opened any pull requests – I’ll take it upon myself to keep this package up to date as well. The changes I’ve made to iTeamTalk, which I’ll submit for review if TeamTalkKit is included in the repository, use TeamTalkKit. There are no longer any C functions used in the application, nor any C structures; everything already uses the Swift API. However, given how frequently the SDK is updated, I don’t think it’s worth worrying about whether TeamTalkKit is up to date. If you read carefully, most of TeamTalkKit’s functions and classes simply call the SDK’s C functions – apart, of course, from the commands and their processing, where I have indeed written a little extra code.

bear101:
Ok, I'll squash the commits when I merge them in. Is everything ready? As far as I can tell there's no changes to the iTeamTalk client's UI in this PR?

GruiaChiscop:
Not in this pr, I'll open another pull request once this one is merged and closed.

bear101 (approved):

PR #3449, opened 2026-09-18
Redesigned the profiles dialog

amirmahdifard:
The new profiles dialog is now modern and much useful, like the new user accounts dialog and server list dialogs, It is also possible to rename a profile raather than deleting it and recreating a new one. And much better.


Client/qtTeamTalk/profiles.ui
```diff
@@ -0,0 +1,111 @@
...
+          <widget class="QPushButton" name="addButton">
+           <property name="text">
+            <string>&amp;New Profile</string>
+           </property>
+          </widget>
+         </item>
+         <item>
+          <widget class="QPushButton" name="currentProfileButton">
+           <property name="text">
+            <string>&amp;Current Profile</string>
```
bear101: Change text to `Use &amp;Current Profile`

bear101 (changes requested):

Client/qtTeamTalk/profiles.ui
amirmahdifard: lovely done!

amirmahdifard:
@bear101 done! thanks!

bear101 (approved):

amirmahdifard:
Thanks!

bear101:
Good to get some code that moves logic out of mainwindow.cpp. It's way to big.

PR #3450, opened 2026-09-19
Qt client: Speak the same tx/rx as the status bar for Ctrl+T

### LINKED ISSUE #3157: Ctrl+T stats consistently lower than shown on status bar
### Description

On any server I've tested via Windows TeamTalk 5.22.0.5193 beta, assuming someone is transmitting, the RX value said by Ctrl+T is lower than the value shown on the status bar, often significantly so. This is also true for TX.

Examples:

* On a server with one transmitter, I see an RX status bar value ranging between 13 and 14 kB but Ctrl+T values ranging from 566 bytes to 11 kB during the same period.

* On a server with six transmitters, I see RX status-bar values from 34 to 36 kB but Ctrl+T values from 2 to 20 kB most of the time, with one jump to 35 kB seen.

### Application

- [x] qtTeamTalk
- [ ] TeamTalkAndroid
- [ ] iTeamTalk
- [ ] TeamTalkClassic
- [ ] TeamTalkServer

### Platform

- [x] Windows
- [ ] macOS
- [ ] Android
- [ ] iOS
- [ ] Linux

### Expected behavior

I expected the values to be very similar across the same span of time.

### Actual behavior

Values for Ctrl+T can occasionally approach values shown on the status bar but seem far more often to be much lower.

1. Join a server with one or more people sending audio and while a screen reader is running or TTS is set up and available (so Ctrl+T will speak).

2. Intermittantly type Ctrl+T to hear RX and TX values, and also examine the status line, visually or with a screen reader command (e.g., Insert+PgDn in JAWS if my TeamTalk scripts are running, object nav to find it if using NVDA).

trypsynth:
Fixes #3157.

## Problem

Ctrl+T speaks RX and TX figures consistently lower than the ones shown in the status bar.

## Cause

Both places compute the same thing, bytes since `m_clientstats`, but only one of them owns that snapshot.

The one second timer takes the delta and then assigns `m_clientstats = stats`, so it always reports a full second of traffic.

`speakClientStats()` computes the delta against that same snapshot but runs whenever the user presses the shortcut, which is some fraction of a second after the last tick. So it reports only the traffic since that tick. Press it right after one and it reads close to nothing.

## Change

The timer stores the figures it already computes, and `speakClientStats()` reads them instead of recomputing. Ctrl+T now speaks exactly what the status bar is showing.

Ping is still read fresh, since it is an instantaneous value rather than a delta and does not have this problem.

## Notes

One behaviour change worth naming: after disconnecting, Ctrl+T now speaks the last figures from while you were connected, instead of a delta taken between two post-disconnect readings. The status bar already stops updating on disconnect and keeps its last value, so this makes the two consistent.

## Testing

Built against Qt 6.10.3 with MSVC 2022 x64 and tested on Windows 11 while connected to a server. Pressing Ctrl+T repeatedly now reads the same RX and TX as the status bar, rather than varying with how soon after the tick it is pressed.


amirmahdifard:
@trypsynth Thanks for fixing this. Last time, @BlindMaster24 broke this. He made ctrl+t refresh everytime when pressed, but still it didn't announce the ping, and it even became worce for some people. If you reverted that and made it to announce the ping also everytime, and not refresh everytime we pressed ctrl+t, thanks. Because that didn't fix anything it just made a delay for us.

Client/qtTeamTalk/mainwindow.cpp
```diff
@@ -7983,11 +7983,8 @@ void MainWindow::speakClientStats()
...
     ClientStatistics stats = {};
     TT_GetClientStatistics(ttInst, &stats);
-    float rx = float(stats.nUdpBytesRecv - m_clientstats.nUdpBytesRecv);
-    float tx = float(stats.nUdpBytesSent - m_clientstats.nUdpBytesSent);
-    int ping = stats.nUdpPingTimeMs;
-    QString strstats = QString("RX: %1, TX: %2").arg(getFormattedSize(rx)).arg(getFormattedSize(tx));
-    if (ping >= 0)
-        strstats += QString(", PING: %3").arg(ping);
+    QString strstats = QString("RX: %1, TX: %2").arg(getFormattedSize(m_rxbytes)).arg(getFormattedSize(m_txbytes));
+    if (stats.nUdpPingTimeMs >= 0)
```
bear101: Use `m_clientstats.nUdpPingTimeMs` instead and delete two first lines in `MainWindow::speakClientStats()`

bear101 (changes requested):

Client/qtTeamTalk/mainwindow.cpp
trypsynth: Done, and it's better than what I had. The ping label in the status bar is set from that same snapshot right after `m_clientstats = stats`, so reading ping from `m_clientstats` makes Ctrl+T report exactly what's on screen rather than pairing a fresh ping with cached throughput. It also means the function no longer needs to call into the SDK at all.

`speakClientStats()` is now:

```cpp
void MainWindow::speakClientStats()
{
    QString strstats = QString("RX: %1, TX: %2").arg(getFormattedSize(m_rxbytes)).arg(getFormattedSize(m_txbytes));
    if (m_clientstats.nUdpPingTimeMs >= 0)
        strstats += QString(", PING: %3").arg(m_clientstats.nUdpPingTimeMs);
    addTextToSpeechMessage(strstats);
}
```

bear101 (approved):

PR #3451, opened 2026-09-20
Add native Windows ARM64 build support

hexxbyte:
## Summary

Adds native Windows ARM64 build support. Every vendored Windows dependency (ACE, OpenSSL, libvpx, speex, speexdsp, ffmpeg, webrtc, Prism) branched on pointer size to pick between x86/x64, which silently misdetects ARM64 as x64 since both are 64-bit. This introduces `TEAMTALK_WIN_ARCH` (derived from the real `-A` platform) and threads it through each toolchain's build.

Along the way, fixes several real ARM64-specific compile/link bugs:
- ACE's PDB race under parallel builds (`C1041`) — fixed with `/FS`
- libvpx's hardcoded VS2026 toolset version — made overridable (`LIBVPX_VS_VER`)
- An impossible pre-Windows-10 `WINVER` for the Media Foundation feature (ARM64 Windows never existed before Windows 10)
- A missing `_ARM64_` define that made ACE's `config-win32-common.h` incorrectly self-define `_AMD64_`, breaking MSVC's ARM64 interlocked intrinsics (`ReadAcquire8`, `InterlockedExchangeAdd64`, etc.) with `C3861`

Prism's `ZDSR`/`byctrl` screen-reader backends are skipped on ARM64 since those vendor SDKs only ship x64 binaries (no source, no ARM64 build exists).

Adds an `arm64` job to the Windows CI matrix (`windows-11-arm` runner, native ARM64 Qt via `windows_arm64` host, `vcvarsarm64.bat`) and wires up the ARM64 portable archive and Inno Setup installer (`/DArch=arm64`). `TeamTalkClassic` (legacy MFC client) and the .NET-based clients/tests are skipped on this job: the former has x86-only inline assembly (`gui/XPTabCtrl.cpp`, `__USE_MASM__`) that MSVC has never supported on any 64-bit target, and classic .NET Framework has no ARM64 JIT.

## Verification

Verified end-to-end with a real local build on ARM64 Windows hardware (not emulated) — produced genuine native ARM64 binaries (`TeamTalk5.exe`, `TeamTalk5.dll`, `tt5srv.exe`/`tt5svc.exe`/pro variants), confirmed via `dumpbin /headers` (`AA64` machine type), plus a working Inno Setup installer and portable-archive package built through the existing packaging scripts.

Requires two additional Visual Studio components not part of a default Build Tools install: `Microsoft.VisualStudio.Component.VC.ATL.ARM64` and `Microsoft.VisualStudio.Component.VC.MFC` / `.MFC.ARM64`.

## Test plan

- [ ] CI: `arm64` job in `windows.yml` builds and runs on `windows-11-arm`
- [x] Local: full build (all toolchain deps + `TeamTalk5DLL`/Pro + Qt client + servers) succeeds with zero errors on ARM64 Windows hardware
- [x] Local: binaries verified as native ARM64 (`dumpbin /headers`)
- [x] Local: portable archive and Inno Setup installer both build successfully
- [ ] Manual smoke test of the built client/server on ARM64 Windows


bear101 (commented):
Doesn't appear to be working: https://github.com/BearWare/TeamTalk5/actions/runs/35485351988/job/106035969303

Did you test any of this or did you just ask AI to do something?

hexxbyte:
Hello,
I tested this on my ARM machine. I'll fix the CI errors.

hexxbyte:
@bear101 I've made an attempt to fix the CI errors. If there are still any errors, I'll work on fixing those as well.

hexxbyte:
I’ve done CI testing on a testing branch, and there are no errors. Should I mark this as ready for review?

hexxbyte:
Either rerunning the workflow or waiting for this to get merged should fix this. It looks like the Windows workflow failed to download Qt.

trypsynth:
Accusing contributors of submitting unreviewed AI code whenever a PR touches unfamiliar platform plumbing has become a recurring pattern here, and it's actively discouraging people who actually take the time to test on real hardware and fix long-standing upstream issues. Plus, the issues commonly aren't even due to the platform code, like in this case. Dismissing concrete hardware-tested fixes because a CI runner had a transient network issue fetching Qt doesn't help move the project forward and makes people just not want to contribute. Talent can and will go elsewhere.

bear101:
@hwangsihu do you have an idea how we're able to compile for ARM64 on Intel hardware? When I pass location of Qt for ARM64 then CMake invokes moc.exe for arm64 and that of couse cannot run on Intel hardware. (just one of several issues as far as I can tell).

hexxbyte:
You can’t compile the ARM64 version of Qt on Intel hardware, they're different architectures. That’s why we use the Windows 11 ARM runner to compile the ARM builds.
What other issues are you seeing? I’d be happy to fix them.

hwangsihu:
We can use ARM64EC compiler, maybe.

hwangsihu:
Ah, Qt does not support that...

PR #3452, opened 2026-09-20
Fix macOS CI by pinning Doxygen to 1.17.0

trypsynth:
## Problem

The macOS workflow has been failing on every push to master and on every pull request. It is not a compile failure, it is the Doxygen step:

```
Library/TeamTalk_DLL/TeamTalk.h:3191: error: explicit link request to
'TTMessage.clienterrormsg' could not be resolved (warning treated as error, aborting now)
```

## Cause

Doxygen 1.18.0 no longer resolves `Struct.member` references to members of an anonymous union. `TTMessage` holds all of its event payload members in one, so every reference of that form fails. `TeamTalk.dox` sets `WARN_AS_ERROR = YES`, so the first one aborts the run.

It is 47 references in total, not just the one in the message. Neutralising the first simply moves the error to the next:

```
TeamTalk.h:3229: error: explicit link request to 'TTMessage.nPayloadSize' could not be resolved
```

The direct members resolve fine, so `#TTMessage.nSource` and `#TTMessage.ttType` are unaffected. Only the union members break.

macOS picks up Doxygen from `brew install doxygen` in `Build/Makefile`, which is now 1.18.0. Windows does not hit this because `windows.yml` already pins 1.17.0 through `ssciwr/doxygen-install`, added in 2befe77. Ubuntu does not build the documentation at all.

## Change

Gives macOS the same pinned Doxygen step Windows already has, so the two platforms build the documentation with the same version. The step runs after `make depend-mac` so it takes precedence over the Homebrew copy.

## Notes

This restores CI rather than fixing the underlying incompatibility. The real options for 1.18 are to rewrite all 47 references into a form it accepts, or to wait for upstream, since this looks like a regression rather than an intended change. Both seemed like the wrong thing to bundle into a CI fix, so this is deliberately the small one.

Worth knowing for whoever picks that up: `TTMessage::clienterrormsg` and `#ClientErrorMsg` both get past 1.18, while `#TTMessage::clienterrormsg` does not.

## Testing

Reproduced locally with Doxygen 1.18.0 against this exact source, which fails at `TeamTalk.h:3191`. Doxygen 1.17.0 on the same source completes with no errors, which is what this pin restores.


trypsynth:
This didn't work on the first attempt, so here's what happened and what changed.

Adding the pinned Doxygen step wasn't enough. `ssciwr/doxygen-install` installs to `/usr/local/bin` on macOS, but the runners are Apple Silicon and Homebrew lives in `/opt/homebrew/bin`, which comes earlier in `PATH`. The run shows CMake picking the wrong one:

```
[207/609] cd .../Documentation/TeamTalkSDK && /opt/homebrew/bin/doxygen TeamTalk.dox
...
Doxygen version used: 1.18.0
```

`make depend-mac` installs Doxygen through Homebrew, so both were present and the wrong one won.

Rather than fight `PATH`, this now passes `-DDOXYGEN_EXECUTABLE=/usr/local/bin/doxygen` in `CMAKE_EXTRA`. `Documentation/TeamTalkSDK/CMakeLists.txt` invokes `${DOXYGEN_EXECUTABLE}` directly and `find_package(Doxygen)` honours a pre-set value, so the version is chosen explicitly rather than by lookup order.

While debugging this I also checked whether the documentation could simply be written in a way 1.18 accepts, which would avoid pinning altogether. It can't, cleanly:

- `@ref TTMessage::clienterrormsg` fails on 1.18 with "unable to resolve reference", so an explicit link request cannot resolve an anonymous union member at all
- bare `TTMessage::clienterrormsg` stops erroring only because autolinking is best effort, so the links would silently disappear from the docs
- `#ClientErrorMsg` links the type rather than the member, which loses the information about which member to read

1.17 resolves all 48 of these references and 1.18 resolves none, so this looks like a regression upstream rather than an intentional change. Pinning seems right until that's confirmed, and it's worth reporting to Doxygen separately.

bear101 (approved):

PR #3454, opened 2026-09-20
Qt client: Check the errors Prism returns instead of discarding them

trypsynth:
## Problem

`PrismWorker` discards every error Prism returns. `prism_backend_initialize`, `prism_backend_speak`, `prism_backend_braille` and `prism_backend_output` all return `PrismError`, and all four are declared `PRISM_NODISCARD`, which expands to `warn_unused_result` on GCC and Clang and `_Check_return_` on MSVC.

The practical effect is a silent failure mode. If `prism_backend_initialize` fails, nothing notices, `m_backend` stays set, and every later call returns `PRISM_ERROR_NOT_INITIALIZED`, which is also discarded. The user gets no speech and no braille, and there is nothing in the log to say why.

## Change

Checks the result of all four calls.

A failed initialize now frees the backend and clears `m_backend`, so the object does not sit there pretending to work, and the `if (m_backend)` guards in the other three do what they look like they do.

`PRISM_ERROR_ALREADY_INITIALIZED` is accepted as success. `prism_registry_create_best` initializes the backend it selects, so on that path `prism_backend_initialize` is the second call and would otherwise look like a failure. Prism's documentation calls this out directly: applications "SHALL call `prism_backend_initialize` and handle `PRISM_ERROR_ALREADY_INITIALIZED` gracefully".

Failures are reported with `qWarning()` and `prism_error_string()`, matching how the rest of the client logs.

## Notes

This is diagnostics only. It changes no behaviour on a working setup, since every one of these calls returns `PRISM_OK` there. What it changes is a broken setup, which currently looks identical to a silent one.

## Testing

`ENABLE_PRISM` is off in an ordinary local build of the client, so these lines are not compiled by it. The Prism calls were instead compiled against real Prism headers and a real `prism.lib` in a standalone program, which confirms the signatures, the enum names and that `prism_error_string` and `prism_backend_name` take what they are given here.


Client/qtTeamTalk/prismworker.cpp
bear101, line 52: Shouldn't you also warn here?

trypsynth: Good catch, that was the one path I left silent while adding warnings to everything else. Fixed:

```cpp
m_context = prism_init(&cfg);
if (!m_context)
{
    qWarning() << "Failed to create a Prism context";
    return;
}
```

Every failure exit from `initialize()` now says something, so "no speech and no braille" always has a matching line in the log.

bear101 (approved):

Issue #3366, opened 2026-07-14
Following the teamtalk's newly remade dialogs, One old way dialog is still remaning


tohidbareza:
### Description

Teamtalk classic and old normal teamtalk client versions had very messy dialogs e.g server list and account management screens
But in the new versions of teamtalk, these screens have became very clean.
But one confusing problem stil remanes even in the new version.
The text to speach events list, sound events list, chat templates list, and status bar events list
These dialogs are stil in the old confusing stile. When I select a item in these lists, I need to press tab to see the edit box for this entri and edit it
My suggestion is that please also make these screens look same stile as the other new screens.
when I select an item in these lists, instead of I press tab, I should be able to press application and select "edit" in the context menu. When edit is pressed, A new dialog box should open with the edit box for that entri, with the variables button or the Browse button if it's sound event, with an ok button and a cancel button.
The globel options like enable all button or reset all buttons should be stayed as they are currently. Only options related to the currently selected event in the list should be in the context menu. I think the context menu options would be edit, reset to default value of that entri and the other options are globel options
Please note that shortcuts tab in the Preferences dialog is already on the new stile and I don't think that screen will need context menu. Currently, I only think about the following screens I have mentioned
This new stile allows fast work, and we no longer confuse event options with globel options
I hope my suggestion would be useful for teamtalk developers, Thank you for making such a nice app! And I think with my suggestion, everywhere of teamtalk will look nice and new. I think may be you didn't think of this idea that these tabs and screens can also become better by doing this stile that I explaned
Thanks everyone, I'm excitedly looking forward for teamtalk new version!

### Application

- [x] qtTeamTalk
- [ ] TeamTalkAndroid
- [ ] iTeamTalk
- [ ] TeamTalkServer

trancer1994:
Were these the new dialogs Bjorn made?

amirmahdifard:
no. All these new dialogs have been made thanks to @CoBC

trancer1994:
Could a developer or maintainer get to making The text to speach events list, sound events list, chat templates list, and status bar events list in the new style all the other dialogs are in?

amirmahdifard:
@trancer1994 That's exactly what is this issue is requesting. This issue is requesting context menu and subdialogs for these mentiond places, to become like the host manager and user accounts dialog. Lets see, The only person which will be moste capable of doing this task is corentin @CoBC So we'll need to see what he does or says.

PR #3455, opened 2026-09-20
Redesigned statusbar events and chat templates

amirmahdifard:
parshal fix of https://github.com/BearWare/TeamTalk5/issues/3366
I'll open one more another pull request to redesign the remaning ones, which is sound events and text to speach tabs.
Status Bar Events
• Double-click-to-check/uncheck on the table is byte-for-byte unchanged — still wired directly to the model, no detour through anything new.
• Added a right-click context menu with "Edit Message" — but it's only enabled when the row is both customizable and currently checked, exactly matching the original code's rule for when the inline edit box used to become visible (customizable && (e & events)). If you right-click a disabled or non-customizable row, Edit Message is grayed out rather than silently doing nothing.
• SBDefAllValButton ("Reset All to Default Value") moved out to sit with Enable all / Clear all / Revert, since it's a global action that was just visually stuck in the wrong groupbox before.
• New StatusBarEventEditDlg holds the message edit box, Variables button, and per-item Reset-to-Default — opened fresh each time via the context menu, no lingering "currently selected" state left over in the main dialog anymore.
Chat Templates
• Since double-click did nothing before, it's now repurposed to open the edit dialog directly — plus Enter as the keyboard equivalent, plus the context menu's "Edit Template."
• Same treatment for CTDefAllValButton (global, moved out) vs. the per-item edit box/Variables/Reset — the per-item stuff now lives in ChatTemplateEditDlg.
• I tested it, it works completely fine, I have tested it on all cases.


CoBC:
Why added a context menu to edit message, it's just add one more action to edit something when the previous UI offered this action quickly. I'm not convince of these changes.

amirmahdifard:
You can just double click to edit, context menu is just available as normal. This redesign is follow up of server list dialog and user accounts dialog redesign from 5.17

amirmahdifard:
@bear101 can you also review and merge this, I have one more pending pr that i'm waiting to open after this. Thanks!

PR #3458, opened 2026-09-20
Build TeamTalkKit XCFrameworks

bear101:
Hi @GruiaChiscop. It seems I merged the TeamTalkKit branch too early because I cannot build iTeamTalk anymore.

I've tried to build TeamTalkKit but the documentation missing some information. How is macOS's `Library/TeamTalk_DLL/libTeamTalk5.a` built? Using `make -C Build mac` builds `Library/TeamTalk_DLL/libTeamTalk5.dylib`.

Using `make -C build lipo` will build `Library/TeamTalk_DLL/libTeamTalk5.a` for iOS. Not macOS.


GruiaChiscop:
Hello!
I Found the cause: `make -C Build mac` can never produce `libTeamTalk5.a` —
`BUILD_TEAMTALK_LIBRARY_LIB` defaults to `OFF` and `BUILD_TEAMTALK_LIBRARY_DLL`
defaults to `ON` (`Library/TeamTalkLib/CMakeLists.txt`), so `mac` only ever
builds the `.dylib`. `ios-build` already avoids this by setting both flags
explicitly; I added `mac-lib` alongside `mac` with the same approach for
macOS.

Also corrected `README-iOS-XCFramework.md` to point at the new command and
explain why the old one never worked.

Left `Package.swift` alone — whether to bring `TeamTalkNativemacOS` back now
that there's a working build path, or keep it out for now, is your call.

On the iOS 16 floor: no objection from me, although I should have explained why I chose v18 in the first place. To be clearer about what's actually
on `master` right now vs. what's coming — iTeamTalk's own deployment target
there is iOS 16 too, unchanged. As I said in my previous comments, I have a follow-up PR (not open yet) that
bumps the app itself to iOS 18 for the current SwiftUI APIs and the
`@Observable` macro; that's a separate decision from TeamTalkKit's package
floor. A package's minimum doesn't force a consuming app down to it, so
TeamTalkKit staying at iOS 16 and iTeamTalk later moving to 18 don't
conflict — lowering the package's floor to widen who else can use it is fine
by me.
If it's okay, I can push the commit right on the  branch, or I can open another pull request that fixes these issues.

bear101:
@GruiaChiscop please rebase your branches with this branch. It has the TeamTalkKit reapplied.

I'm unable to build iTeamTalk with TeamTalkKit, so I had to revert TeamTalkKit from master branch in this PR #3465

PR #3459, opened 2026-09-20
Ability to customize all shortcuts

CoBC:
@bear101 should we separate global shortcuts from the rest of action shortcuts? DO you know if we can make a group in this table?


bear101:
Hm, I searched a bit but didn't find much. How about creating a new column with the name of the category and then initially sort by that column.

CoBC:
I tried something with headers, what do you think?

bear101:
> I tried something with headers, what do you think?

Looks complicated. Isn't it simpler to have a `COLUMN_SHORTCUT_TYPE` that just shows "Global Shortcut" and "Interface Shortcut"?

CoBC:
Yes, you're right. Done!

bear101 (approved):
Looks good. Please squash commits

PR #3460, opened 2026-09-20
Qt client: Switch back to the selected sound device when it is plugged in again

trypsynth:
First of three, splitting #3444 as requested. This one is the fix on its own, with nothing else in it.

## Problem

On Windows, unplug a headset and plug it back in and the client keeps playing through whatever device it fell back to. It only returns to the device selected in Preferences after pressing Refresh Sound Devices. Other applications follow the hardware by themselves.

## Cause

`CLIENTEVENT_SOUNDDEVICE_ADDED` only added a status message, and `CLIENTEVENT_SOUNDDEVICE_UNPLUGGED` only wrote to `qDebug`. Neither reinitialized sound, so a device coming back was never acted on. Only `CLIENTEVENT_SOUNDDEVICE_REMOVED` did anything.

## Change

`clienteventSoundDeviceAdded()` reinitializes when the device that appeared is the one selected in settings, which is the replug case. Otherwise it keeps the existing status message unchanged.

`CLIENTEVENT_SOUNDDEVICE_UNPLUGGED` now reinitializes when the unplugged device is the one currently in use, matching what `CLIENTEVENT_SOUNDDEVICE_REMOVED` already did. Removal covers devices that disappear entirely, such as USB; unplugged covers devices with jack detection that stay enumerated.

Both reuse `getSoundDeviceUID()` and `initSound()`, which already exist and are already used for exactly this by the removal path.

The case labels are left where they are and nothing is moved, so the diff is only the behaviour change.

## Notes

Two further changes are deliberately not here and will follow separately:

- Windows sends a burst of notifications for a single plug, one per role plus a state change, so the reinitialization is worth debouncing. Without it the sound system restarts several times in a row. It works, it is just wasteful.
- `getSoundInputFromUID()` and `getSoundOutputFromUID()` fall back to a stale device index when the stored UID no longer resolves, which can open a different device entirely. That is a separate bug in code shared with `preferencesdlg.cpp`.

## Testing

Built against Qt 6.10.3 with MSVC 2022 x64 and tested on Windows 11. With a specific output device selected in Preferences, unplugging moves playback off it and plugging back in returns it to that device with no manual refresh.


Client/qtTeamTalk/mainwindow.cpp
```diff
@@ -1672,9 +1672,27 @@ void MainWindow::clienteventUserAudioBlock(int source, StreamTypes streamtypes)
 
 void MainWindow::clienteventSoundDeviceAdded(const SoundDevice& snddev)
 {
+    auto devid = getSoundDeviceUID(snddev);
+    // the device the user selected is back, e.g. a headset plugged in again
+    if (devid.size() && (devid == ttSettings->value(SETTINGS_SOUND_INPUTDEVICE_UID, "").toString() ||
+                         devid == ttSettings->value(SETTINGS_SOUND_OUTPUTDEVICE_UID, "").toString()))
+    {
+        initSound();
+        return;
```
bear101, line 1681: Why do you skip the status message? It's still relevant for the user to know that a new sound device has popped up.

Client/qtTeamTalk/mainwindow.cpp
```diff
@@ -1672,9 +1672,27 @@ void MainWindow::clienteventUserAudioBlock(int source, StreamTypes streamtypes)
...
 }
 
+void MainWindow::clienteventSoundDeviceUnplugged(const SoundDevice& snddev)
+{
+    qDebug() << "Unplugged sound device: " << _Q(snddev.szDeviceName);
+
+    auto devid = getSoundDeviceUID(snddev);
+    if (devid.size() && (devid == getSoundDeviceUID(m_devin) || devid == getSoundDeviceUID(m_devout)))
+        initSound();
+}
```
bear101, line 1694: Should we add a status message that the sound device has disappeared?

PR #3461, opened 2026-09-20
Qt client: Fall back to the default sound device instead of a stale device index

trypsynth:
Third of three, splitting #3444. Independent of the other two and can land in any order.

## Problem

When the sound device stored in settings can no longer be resolved, the client can open a completely different device rather than falling back to the default.

## Cause

`getSoundInputFromUID()` and `getSoundOutputFromUID()` look the device up by its stored UID, and when that fails they return the device *index* that was passed in:

```cpp
SoundDevice dev;
if (getSoundDevice(uid, true, inputdev, dev))
    inputid = dev.nDeviceID;

return inputid;
```

That index came from settings and was valid when it was saved. Device indices are positional though, so they shift whenever an endpoint appears or disappears. Once a device is unplugged, the saved number often refers to a different device entirely, initialization succeeds on it, and audio goes somewhere the user never chose.

The failure is quiet, because opening the wrong device looks exactly like opening the right one.

## Change

Both functions now fall back to the default device when the UID cannot be resolved, instead of trusting an index that is known to be stale.

Also changes the successful path to return directly, which makes the two outcomes explicit rather than assigning through a variable that is returned either way.

## Notes

`preferencesdlg.cpp` uses the same two functions through `getSelectedSndInputDevice()` and `getSelectedSndOutputDevice()`, so the refresh button in Preferences benefits from this too.

This is why, in the original report behind #3460, unplugging a headset sent audio to the speakers rather than simply stopping. The device had gone, and its old index pointed at something else.

## Testing

Built against Qt 6.10.3 with MSVC 2022 x64 and tested on Windows 11 with a specific output device selected in Preferences. Unplugging it moves playback to the Windows default device.

Worth noting that the old behaviour was nondeterministic, since which device the stale index lands on depends on where the removed endpoint sat in the enumeration order. The change removes the guesswork rather than fixing one particular wrong outcome.


Client/qtTeamTalk/utilsound.cpp
bear101, line 202: The method is called `getSoundInputFromUID()`. What does this have to do with default sound input device?

bear101 (changes requested):

PR #3463, opened 2026-09-21
Qt client: Make media sliders respond to the keyboard

### LINKED ISSUE #494: Cannot set stream start position using keyboard on
I use TeamTalk5 on Debian Buster so the QT client with Orca.
When I set the starting position for a stream using the keyboard, the stream starts at the beginning instead of the position set.
Moreover, we have not a persentage but a list of values from 0 to 10000 to set the start position for Orca and the time is not changed on the label. So it is not possible to precizely know what starting time we set even if it worked.

### LINKED ISSUE #3224: Qt client. Sliders on media tab are not useable with keyboard, only mouse can moove them and they work
### Description

I've connected to github after my cuntry disabled the internet for 50 days, so I can report this bug finally.
@CoBC Recently, I'm talking to my sited friends in teamtalk. Some times we happen to watch moovys together in teamtalk by streaming, when they stream a file, They could easily seek the file backward and forward to whereever they wish by mooving the position slider in media tab, and turn the volume down and up on real time while playing, with volume slider. I tryed to do the same while streaming, but, when I moved both of those sliders, anything didn't happen, and they would instantly go back to their original position. For example if the volume slider is at 1000, and I press my arrow keys to move it to 1001, it will announce 1001, and instantly it'll go back to 1000 again. But this problem doesn't exist on mouse at all, and my friends could easily use these sliders and enjoy streaming and controling it in real time. 5.22

### Application

- [x] qtTeamTalk
- [ ] TeamTalkAndroid
- [ ] iTeamTalk
- [ ] TeamTalkClassic
- [ ] TeamTalkServer

### Platform

- [x] Windows
- [ ] macOS
- [ ] Android
- [ ] iOS
- [ ] Linux

### Expected behavior

like other normal sliders, the sliders on media tab should be useable and moveable by the keyboard and screen readers

### Actual behavior

the sliders are only useable and moovable by mouse for sited users on media tab

1: stream a file.
2: go to media tab on main window
3: try to move the position and volume sliders both with the keyboard, and mouse.
with mouse there is no problem, but with keyboard they are not useable at all

trypsynth:
Fixes #3224. Fixes #494.

## Problem

Three sliders do nothing when moved from the keyboard:

- the playback position slider on the Media tab
- the volume slider on the Media tab
- the start position slider in the stream media file dialog

On the Media tab the value visibly changes and then snaps back within a second, which is #3224. In the stream dialog the label does not update and the stream starts from the beginning regardless of where the slider was set, which is #494.

## Cause

All three connect to `QSlider::sliderMoved`. Qt only emits that while the handle is being dragged with the mouse. Arrow keys, Page Up and Down, Home and End go through `setValue()` and never emit it, so from the keyboard none of the three handlers run.

The snap back on the Media tab follows from that. `changeMediaFileOffset()` is what arms `TIMER_CHANGE_MEDIAFILE_POSITION`, and `setMediaFileTabProgress()` only leaves the slider alone while that timer exists. With the handler never running, the next progress update puts the slider back to the real playback position.

For #494, `m_mfp.uOffsetMSec` is only ever set by `slotChangePlayOffset()`, and that value is what `streammediafiledlg.cpp:124` saves as the stream's start offset. From the keyboard it stays at zero.

These are the only three sliders in the client connected this way. Every other one uses `valueChanged`.

## Change

All three now connect to `QAbstractSlider::actionTriggered` and read `sliderPosition()`.

## Why not `valueChanged` like the others

That was the first thing I tried, and it would introduce two regressions, because unlike the other sliders these three are also written programmatically as playback progresses:

- `setValue(0)` runs whenever a stream ends. Through `valueChanged` it would reach `changeMediaFileOffset(0)`, arm the seek timer, and half a second later try to seek a stream that has already closed, showing "Failed to change playback position" every time a stream finishes.
- The volume block calls `setRange()` then `setValue()` on every status update. `setRange()` can clamp and emit, reaching `changeMediaFileVolume()`, which on failure shows "Failed to change volume of the stream" and stops the stream.

Both could be worked around with `QSignalBlocker` at each programmatic write, but that is fragile: any future `setValue()` added without a blocker brings the bugs back.

`actionTriggered` fires for every user action, keyboard, mouse wheel and drag alike, and never for a programmatic `setValue()`. So it fixes the keyboard without making the progress updates act like user input, and needs nothing added at the call sites that move these sliders.

When it fires, `sliderPosition()` already holds the new position, which is why the handlers read that rather than `value()`.

## Testing

Built against Qt 6.10.3 with MSVC 2022 x64 and tested on Windows 11 with NVDA, streaming a media file several minutes long:

- position slider on the Media tab, with arrows, Page Up and Down, Home and End: playback moves to the new position and stays there
- volume slider on the Media tab: the stream volume changes live
- start position in the stream media file dialog: the time label follows the slider, and the stream begins at the chosen position
- letting a stream run to the end: no error dialog, which was the regression that ruled out `valueChanged`

**Not tested: dragging with the mouse.** I'm blind and can't test that path myself. It should behave as before, since a drag emits `actionTriggered` with `SliderMove` exactly as it previously emitted `sliderMoved`, but it's the path sighted users rely on and the one that currently works, so I'd rather it was checked than assumed. @bear101, or anyone sighted, could you drag both Media tab sliders and the stream dialog slider once before this is merged?


amirmahdifard:
@trypsynth Thanks for fixing this! this issue was really annoying and this fix was really needed! I was going to have a look, but thank you for doing the hard work and fixing the correct way.

bear101 (commented):
Shouldn't the singleStep and pageStep values be changed. It's very slow to change volume or playback offset.

